PR #2, opened 2016-02-24
Update NSMutableArray setObject atIndexedSubscript to make it more compatible with the call in OS X 10.8

plaurent:
- Grow array if needed, as per OS X 10.8 API documentation
- Add comment to explain what setObject... is doing.

### DIFF
--- a/Source/NSArray.m
+++ b/Source/NSArray.m
@@ -2055,9 +2055,18 @@ compare(id elem1, id elem2, void* context)
   [self subclassResponsibility: _cmd];
 }
 
+/**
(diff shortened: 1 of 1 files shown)


ivucica:
Hello,

we do not use Github for development. We accept smaller patches on our mailing list(s). (We currently have the policy that larger ones require copyright assignment to FSF.)

Can you please send your patch to our developer mailing list?

Thanks.

ivucica:
Your change has been applied upstream in a51249a65d23e7280e51ed534da29df1c82e15cb (Subversion r39411).

Closing.

PR #8, opened 2017-06-26
General fixes for Base

theiostream:
This pull request provides a series of general fixes to Base for increased compatibility with newer OSX APIs used by WebKit.

The commit logs add more insight into the individual changes done to the project.

### DIFF
--- a/Headers/Foundation/NSCoder.h
+++ b/Headers/Foundation/NSCoder.h
@@ -28,6 +28,7 @@
 
 #import	<Foundation/NSObject.h>
 #import	<Foundation/NSGeometry.h>
+#import	<Foundation/NSSet.h>
 #import	<Foundation/NSZone.h>
 
 #if	defined(__cplusplus)
@@ -303,6 +304,8 @@ extern "C" {
  */
 - (id) decodeObjectForKey: (NSString*)aKey;
 
+
+
 /** <override-subclass />
  * Encodes aBool and associates the encoded value with aKey.
  */
@@ -363,6 +366,20 @@ extern "C" {
  * Decodes an NSInteger associated with the key.
  */
 - (NSInteger) decodeIntegerForKey: (NSString *)key;
+#endif
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_8, GS_API_LATEST)
+
+#if GS_HAS_DECLARED_PROPERTIES
+@property (nonatomic, assign) BOOL requiresSecureCoding;
+#else
+- (BOOL) requiresSecureCoding;
+- (void) setRequiresSecureCoding: (BOOL)requires;
+#endif
+
+- (id) decodeObjectOfClass: (Class)cls forKey: (NSString *)key;
+- (id) decodeObjectOfClasses: (NSSet *)classes forKey: (NSString *)key;
+
 #endif
 @end
 
--- a/Headers/Foundation/NSDate.h
+++ b/Headers/Foundation/NSDate.h
@@ -29,6 +29,8 @@
 extern "C" {
 #endif
 
+GS_EXPORT NSString * const NSSystemClockDidChangeNotification;
+
 /**
  * Time interval difference between two dates, in seconds.
  */
--- a/Headers/Foundation/NSFileManager.h
+++ b/Headers/Foundation/NSFileManager.h
@@ -345,6 +345,9 @@ typedef NSUInteger NSDirectoryEnumerationOptions;
  * it can't be read for some reason).
  */
 - (NSArray*) contentsOfDirectoryAtPath: (NSString*)path error: (NSError**)error;
+
+- (NSDictionary*) attributesOfFileSystemForPath: (NSString*)path
+                                          error: (NSError**)error;
 #endif
 
 - (BOOL) copyPath: (NSString*)source
--- a/Headers/Foundation/NSFileManager.h
+++ b/Headers/Foundation/NSFileManager.h
@@ -191,6 +191,8 @@ extern "C" {
 @class NSError;
 @class NSURL;
 
+@protocol NSFileManagerDelegate;
+
 /* MacOS-X defines OSType as a 32bit unsigned integer.
  */
 #ifndef OSTYPE_DECLARED
@@ -210,6 +212,7 @@ typedef NSUInteger NSDirectoryEnumerationOptions;
 {
 #if	GS_EXPOSE(NSFileManager)
 @private
+  id<NSFileManagerDelegate> _delegate;
   NSString	*_lastError;
 #endif
 #if     GS_NONFRAGILE
@@ -228,6 +231,15 @@ typedef NSUInteger NSDirectoryEnumerationOptions;
  * application.
  */
 + (NSFileManager*) defaultManager;
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
+#if GS_HAS_DECLARED_PROPERTIES
+@property (assign) id<NSFileManagerDelegate> delegate;
+#else
+- (id<NSFileManagerDelegate>) delegate;
+- (void) setDelegate: (id<NSFileManagerDelegate>)delegate;
+#endif
+#endif
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
 - (NSDictionary *) attributesOfItemAtPath: (NSString*)path
@@ -444,6 +456,7 @@ typedef NSUInteger NSDirectoryEnumerationOptions;
  * if they wish to deal with copy and move operations performed
  * by NSFileManager.
  */
+
 @interface NSObject (NSFileManagerHandler)
 /**
  * <p>When an error occurs during a copy or move operation, the file manager
@@ -494,7 +507,6 @@ typedef NSUInteger NSDirectoryEnumerationOptions;
 - (void) fileManager: (NSFileManager*)fileManager
      willProcessPath: (NSString*)path;
 @end
-
 
 /**
  *  <p>This is a subclass of <code>NSEnumerator</code> which provides a full
@@ -658,6 +670,77 @@ GS_EXPORT NSString* const NSFileSystemFreeNodes;
 - (NSUInteger) fileSystemNumber;
 - (NSUInteger) fileSystemFileNumber;
 @end
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_5,GS_API_LATEST)
+
+@protocol NSFileManagerDelegate <NSObject>
+#if GS_PROTOCOLS_HAVE_OPTIONAL
+@optional
+#else
+@end
+@interface NSObject (NSFileManagerDelegate)
+#endif
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldCopyItemAtPath: (NSString *)srcPath
+                toPath: (NSString *)dstPath;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldCopyItemAtURL: (NSURL *)srcURL
+                toURL: (NSURL *)dstURL;
+
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+        copyingItemAtPath: (NSString *)srcPath
+                   toPath: (NSString *)dstPath;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+         copyingItemAtURL: (NSURL *)srcURL
+                    toURL: (NSURL *)dstURL;
+
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldMoveItemAtPath: (NSString *)srcPath
+                toPath: (NSString *)dstPath;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldMoveItemAtURL: (NSURL *)srcURL
+                toURL: (NSURL *)dstURL;
+
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+         movingItemAtPath: (NSString *)srcPath
+                   toPath: (NSString *)dstPath;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+          movingItemAtURL: (NSURL *)srcURL
+                    toURL: (NSURL *)dstURL;
+
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldLinkItemAtPath: (NSString *)srcPath
+                toPath: (NSString *)dstPath;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldLinkItemAtURL: (NSURL *)srcURL
+                toURL: (NSURL *)dstURL;
+
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+        linkingItemAtPath: (NSString *)srcPath
+                   toPath: (NSString *)dstPath;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+         linkingItemAtURL: (NSURL *)srcURL
+                    toURL: (NSURL *)dstURL;
+
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldRemoveItemAtPath: (NSString *)path;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldRemoveItemAtURL: (NSURL *)URL;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+       removingItemAtPath: (NSString *)path;
+- (BOOL)fileManager: (NSFileManager *)fileManager
+  shouldProceedAfterError: (NSError *)error
+        removingItemAtURL: (NSURL *)URL;
+@end
+
+#endif
 
 #if	defined(__cplusplus)
 }
--- a/Headers/Foundation/NSHTTPCookie.h
+++ b/Headers/Foundation/NSHTTPCookie.h
@@ -220,6 +220,14 @@ extern NSString * const NSHTTPCookieVersion; /** Obtain cookie version */
 - (BOOL) isSessionOnly;
 #endif
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6,GS_API_LATEST)
+#if GS_HAS_DECLARED_PROPERTIES
+@property (readonly, getter=isHTTPOnly) BOOL HTTPOnly;
+#else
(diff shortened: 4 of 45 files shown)


ivucica:
It might be interesting to have smaller pull requests in the future, consisting of just a few, very related commits?

I'm sorry for not noticing this PR earlier. I think @rfm would also be a better person to do the review.

However, for starters, I guess that since we /do/ have Travis set up, maybe the review should wait until you make the build go green?

Headers/Foundation/NSCoder.h
```diff
@@ -363,6 +363,9 @@ extern "C" {
  * Decodes an NSInteger associated with the key.
  */
 - (NSInteger) decodeIntegerForKey: (NSString *)key;
+
+
+@property (nonatomic, assign) BOOL requiresSecureCoding;
```
fredkiefer: Please replace this with a getter and a setter.

Headers/Foundation/NSDictionary.h
```diff
@@ -67,8 +67,9 @@ extern "C" {
 			     indent: (NSUInteger)level;
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
-DEFINE_BLOCK_TYPE(GSKeysAndObjectsEnumeratorBlock, void,
-  GS_GENERIC_TYPE_F(KeyT,id<NSCopying>), GS_GENERIC_TYPE(ValT), BOOL*);
+/*DEFINE_BLOCK_TYPE(GSKeysAndObjectsEnumeratorBlock, void,
```
fredkiefer: This is wrong and you already know it. But why did you change it anyway? This breaks your pull request completely.

Headers/Foundation/NSFileManager.h
```diff
@@ -230,6 +233,10 @@ typedef NSUInteger NSDirectoryEnumerationOptions;
 + (NSFileManager*) defaultManager;
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
+@property (assign) id<NSFileManagerDelegate> delegate;
```
fredkiefer: Again, please provide a getter and a setter.

Headers/Foundation/NSFileManager.h
```diff
@@ -659,6 +669,68 @@ GS_EXPORT NSString* const NSFileSystemFreeNodes;
 - (NSUInteger) fileSystemFileNumber;
 @end
 
+@protocol NSFileManagerDelegate <NSObject>
```
fredkiefer: Please have a look at NSMetaData.h on how to do this in a way old gcc can cope with.

@protocol NSMetadataQueryDelegate
#if OS_API_VERSION(MAC_OS_X_VERSION_10_6,GS_API_LATEST) && GS_PROTOCOLS_HAVE_OPTIONAL
@optional
#else
@end
@interface NSObject (NSMetadataQueryDelegate)
#endif // GS_PROTOCOLS_HAVE_OPTIONAL

Headers/Foundation/NSHTTPCookie.h
```diff
@@ -204,13 +204,13 @@ extern NSString * const NSHTTPCookieVersion; /** Obtain cookie version */
  * Returns whether the receiver should only be sent over
  * secure connections.
  */
-- (BOOL) isSecure;
+@property (readonly, getter=isSecure) BOOL secure;
```
fredkiefer: Please restore the old state.

Headers/Foundation/NSHTTPCookie.h
```diff
@@ -204,13 +204,13 @@ extern NSString * const NSHTTPCookieVersion; /** Obtain cookie version */
...
 
 /**
  * Returns whether the receiver should be destroyed at the end of the
  * session.
  */
-- (BOOL) isSessionOnly;
+@property (readonly, getter=isSessionOnly) BOOL sessionOnly;
```
fredkiefer: Please restore the old state.

Headers/Foundation/NSHTTPCookie.h
```diff
@@ -247,6 +247,10 @@ extern NSString * const NSHTTPCookieVersion; /** Obtain cookie version */
  */
 - (NSUInteger) version;
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6,GS_API_LATEST)
+@property (readonly, getter=isHTTPOnly) BOOL HTTPOnly;
```
fredkiefer: Again a getter will be needed.

Headers/Foundation/NSProcessInfo.h
```diff
@@ -83,6 +101,10 @@ enum {
  */
 + (NSProcessInfo*) processInfo;
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6,GS_API_LATEST) 
+@property (readonly) NSUInteger systemUptime;
```
fredkiefer: Please add a getter method instead.

Headers/Foundation/NSURL.h
```diff
@@ -149,6 +171,8 @@ GS_EXPORT NSString* const NSURLFileScheme;
 - (id) initWithString: (NSString*)aUrlString
 	relativeToURL: (NSURL*)aBaseUrl;
 
+@property (readonly, getter=isFileURL) BOOL fileURL;
```
fredkiefer: Please add a getter method.

Headers/Foundation/NSURLConnection.h
```diff
@@ -86,6 +86,24 @@ extern "C" {
  */
 - (id) initWithRequest: (NSURLRequest *)request delegate: (id)delegate;
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_8,GS_API_LATEST)
+@property (readonly, copy) NSURLRequest *currentRequest;
```
fredkiefer: getter and setter...

Headers/Foundation/NSURLProtectionSpace.h
```diff
@@ -129,6 +138,10 @@ authenticationMethod: (NSString *)authenticationMethod;
  */
 - (BOOL) receivesCredentialSecurely;
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6,GS_API_LATEST) && GS_API_VERSION( 11300,GS_API_LATEST)
+@property (readonly, copy) NSArray *distinguishedNames;
```
fredkiefer: Getter...

Source/NSCoder.m
```diff
@@ -397,6 +408,12 @@ - (id) decodeObjectForKey: (NSString*)aKey
   return nil;
 }
 
+- (id) decodeObjectOfClass: (Class)cls forKey: (NSString *)aKey
```
fredkiefer: Please also add the method decodeObjectOfClasses:forKey: as that is the more basic one, which should actually be implemented.

Source/NSFileManager.m
```diff
@@ -2032,6 +2026,7 @@ - (NSDictionary*) fileSystemAttributesAtPath: (NSString*)path
 
   if (!GetVolumePathNameW(lpath, volumePathName, 128))
     {
+      *error = [NSError _last];
```
fredkiefer: Shouldn't you check whether error is not NULL?

Source/NSHTTPCookie.m
```diff
@@ -59,6 +59,7 @@ after this (See GSRangeOfCookie).
 NSString * const NSHTTPCookieSecure = @"Secure";
 NSString * const NSHTTPCookieValue = @"Value";
 NSString * const NSHTTPCookieVersion = @"Version";
+static NSString * const NSHTTPCookieHTTPOnly = @"HTTPOnly";
```
fredkiefer: I would not use a variable name starting with NS for a static variable. Apple uses this name space and we might get a clash later on.

Source/NSHTTPCookie.m
```diff
@@ -736,7 +745,18 @@ static inline id parseUnquotedString(pldata *pld, char endChar)
 	}
       else
 	{
+	  unsigned int oldpos = pld->pos, keyvalpos;
```
fredkiefer: Please use a separate line for each variable.

Source/NSURLConnection.m
```diff
@@ -226,12 +257,36 @@ - (id) initWithRequest: (NSURLRequest *)request delegate: (id)delegate
 	initWithRequest: this->_request
 	cachedResponse: nil
 	client: (id<NSURLProtocolClient>)self];
-      [this->_protocol startLoading];
+      if (immediately)
+        {
+          this->_started = YES;
```
fredkiefer: Why don't you call the method start here?

Source/NSURLConnection.m
```diff
@@ -452,6 +507,11 @@ - (void) URLProtocol: (NSURLProtocol *)protocol
 	initWithRequest: this->_request
 	cachedResponse: nil
 	client: (id<NSURLProtocolClient>)self];
+
```
fredkiefer: Again, wouldn't calling the method start be more consistent here? E.g. that would set the new _started instance variable.

Source/NSURLConnection.m
```diff
@@ -476,5 +536,80 @@ - (void) URLProtocol: (NSURLProtocol *)protocol
...
+    }
+  
+  if (this->_customOperationQueue)
+    {
+      NSLog(@"you cannot use both run loops and an operation queue");
+      return;
+    }
+  
+  NSMutableArray *modes = [this->_customRunLoops objectForKey: loop];
+  if ([this->_customRunLoops objectForKey: loop])
```
fredkiefer: Why don't you use your local variable modes here?

Source/NSURLCredential.m
```diff
@@ -110,6 +110,9 @@ - (id) initWithUser: (NSString *)user
     }
   if ((self = [super init]) != nil)
     {
+      if (persistence == NSURLCredentialPersistenceSynchronizable)
+	persistence = NSURLCredentialPersistencePermanent;
```
fredkiefer: Please use {} even for single line if statements.

Source/NSURLProtocol.m
```diff
@@ -228,10 +228,14 @@ - (void) close
 {
   [ip setDelegate: nil];
   [op setDelegate: nil];
+
+  // FIXME: should we add support for other run loops
+  // or to the operation queue here?
```
fredkiefer: Yes! And maybe call this close method from dealloc?

Source/NSURLProtocol.m
```diff
@@ -1802,10 +1979,32 @@ - (void) stopLoading
     {
       [this->input setDelegate: nil];
       [this->output setDelegate: nil];
-      [this->input removeFromRunLoop: [NSRunLoop currentRunLoop]
-			     forMode: NSDefaultRunLoopMode];
-      [this->output removeFromRunLoop: [NSRunLoop currentRunLoop]
-			      forMode: NSDefaultRunLoopMode];
+      if (this->customOperationQueue)
```
fredkiefer: Could you please try to reduce the duplication of code in this class? To me it seems like you had to copy similar bits over and over again.

Headers/Foundation/NSDate.h
```diff
@@ -29,6 +29,8 @@
 extern "C" {
 #endif
 
+GS_EXPORT NSString * const NSSystemClockDidChangeNotification;
```
fredkiefer: You need to add a definition of this value somewhere.

fredkiefer:
I added a few commets inline.

Headers/Foundation/NSHTTPCookie.h
ivucica: To clarify: if the `@property` syntax is required for compatibility, please provide it as an alternative and guarded by appropriate feature guards.

Alternatively, provide `- (BOOL) isSessionOnly` plus `- (BOOL) sessionOnly` as a wrapper around it.

`@property` is really breaking the 3.4.5 requirement which we outlined in the mailing list.

Headers/Foundation/NSDictionary.h
theiostream: Was a temporary (and very quick) fix to avoid a casting hell error when building WebKit, and I kept it as a commit as a reminder that I'd have to look at it later. Should not have included it in this pull request.

Source/NSFileManager.m
theiostream: I think there are other places in the code that do not check for this, but I'll investigate further.

In any case, having `*error` set to `NULL` is not wrong (although it would be quite misleading). In this case, what should we do? Generate a new `NSError` right here? What should it contain?

rfm:
> On 28 Jun 2017, at 13:09, Daniel Ferreira <notifications@github.com> wrote:
> 
> @theiostream commented on this pull request.
> 
> In Source/NSFileManager.m:
> 
> > @@ -2032,6 +2026,7 @@ - (NSDictionary*) fileSystemAttributesAtPath: (NSString*)path
>  
>    if (!GetVolumePathNameW(lpath, volumePathName, 128))
>      {
> +      *error = [NSError _last];
> 
> I think there are other places in the code that do not check for this, but I'll investigate further.
> 
> In any case, having *error set to NULL is not wrong (although it would be quite misleading). In this case, what should we do? Generate a new NSError right here? What should it contain?

I think Fred was saying that you have a potential null pointer dereference.
So the code ought to be:

 if (!GetVolumePathNameW(lpath, volumePathName, 128))
     {
       if (NULL != error)
	{
	  *error = [NSError _last];
        }

Generally if a method has pointer arguments used to return valuers, we ought to check to see if those pointers are null and do somethjing sensible (either quietly refrain from assigning to the address or raise an exception ... whatever OSX does).

theiostream:
Hi @fredkiefer, thanks for the comments. I replied in those I had something to say about; otherwise, I agree completely.

I think I'd mentioned in a prior e-mail that I knew `@property`s were still an unsolved problem (I hope I'd spare you the effort of pointing out every misuse of them). @rfm's comment did not show up in the issue tracker, so I'll copy it for visibility:

```
While we definitely want to provide explicit getter/setter methods for any properties declared in new Apple headers, that's not really sufficient.  We also need to conditionally use the @property declaration.

[...]

I'm wondering what the best way to proceed is:
1. change the configure script for the base library to check to see if the current compiler supports @property and define GS_HAS_DECLARED_PROPERTIES in GSConfig.h so all our headers can use it.
2. change library-combo.make to define GS_HAS_DECLARED_PROPERTIES if the next generation (ng) librarary is selected (for new ObjC features with clang) and don't use properties otherwise.
```

How do gccs with new Objective-C 2.0 runtimes enter this picture?

Source/NSFileManager.m
theiostream: Oh, nevermind. I just got where the problem is (thanks to @rfm's insight). Thanks for the heads-up! :)

theiostream:
A separate comment about `NSURLConnection`/`NSURLProtocol` stuff: I agree with all of the issues you mentioned (in fact, I realized them myself as I went on coding it). I'm determined to fix them, but before I continued any work on that I wanted your two cents on that compatibility matter with `NSURLQueue` I mentioned in some prior emails (copying here for visibility):

```
So it seems like -setDelegateQueue: not only gets delegate calls
redirected to that queue. It *also* guarantees that even if the thread
where -[NSURLConnection start] was called on dies, the delegate calls
will still be placed on the queue. Unfortunately, this seems hard to
conciliate with our reliance on NSStream's -handleEvent: model, which
requires -[NSURLConnection start] to be run on a thread with a run
loop running.

While this does seem complicated, the only workaround I can think of
not to break compatibility with this Apple guarantee is do everything
-[NSURLProtocol startLoading] does in a new thread managed by us that
will always have the run loop running. In case we're using
NSOperationQueue, we could easily add operations to it from there; in
case we're not, we could keep a reference to the thread that called
-[NSURLConnection init] or -[NSURLConnection start] originally and
dispatch the delegate selectors to them. This way we guarantee that if
the thread that called -[NSURLConnection start] dies, we'll still get
the NSStream events. Of course, I may be delirious here, so please
correct me if that's a stupid way to go about this.
```

Again, thanks for the thorough review.

theiostream:
Hm, @rfm, when Fred pointed me to the `GS_PROTOCOLS_HAVE_OPTIONAL` I couldn't help but think our definition of `GS_HAS_DECLARED_PROPERTIES` could have the same style: https://github.com/gnustep/libs-base/pull/8/commits/8ffc352abed8ea156bff7fcf9f7e253f2dad6656

What do you think?

@fredkiefer the branch has been updated fixing everything you pointed out (hopefully). I removed the two WIP commits.

rfm:
> On 28 Jun 2017, at 15:31, Daniel Ferreira <notifications@github.com> wrote:
> 
> Hm, @rfm, when Fred pointed me to the GS_PROTOCOLS_HAVE_OPTIONAL I couldn't help but think our definition of GS_HAS_DECLARED_PROPERTIES could have the same style: 8ffc352
> 
> What do you think?

Yes, I actually think that looks better than either of my suggestions.

Headers/Foundation/NSCoder.h
```diff
@@ -364,6 +367,19 @@ extern "C" {
  */
 - (NSInteger) decodeIntegerForKey: (NSString *)key;
 #endif
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_8, GS_API_LATEST)
+
+#if GS_HAS_DECLARED_PROPERTIES
+@property (nonatomic, assign) BOOL requiresSecureCoding;
+#else
+- (BOOL) requiresSecureCoding;
```
fredkiefer: I think you will also need a setter, not just a getter method.

Headers/Foundation/NSFileManager.h
```diff
@@ -230,6 +233,14 @@ typedef NSUInteger NSDirectoryEnumerationOptions;
 + (NSFileManager*) defaultManager;
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
+#if GS_HAS_DECLARED_PROPERTIES
+@property (assign) id<NSFileManagerDelegate> delegate;
+#else
+- (id<NSFileManagerDelegate>) delegate;
```
fredkiefer: Again the setter seems to be missing.

theiostream:
Gah, I've just noticed that (apparently starting some days ago) GitHub decided to sort commits not by their `git log` order but by their date, which is totally braindead. It makes it seem that the `GS_HAS_DECLARED_PROPERTIES` definition, on top of which the other commits are built, comes after because I created it today, even though it was created through a rebase.

This probably makes reviewing individual commits (as usually makes sense -- see the git project itself, for instance) hell through this web UI. Knowing that I will continue to rebase my changes, let me know if there is any better way for you to read my commits.

ivucica:
Looking at Github review UI for anything but the smallest PRs, I wish I could justify setting up Gerrit and getting people to use its approach to submitting changes. This in particular is a disaster.

ivucica:
Or if Github had a more decent diff between pushes, which breaks especially if you have rebases.

I think you could have merged this as well, but I'll do it now.

theiostream:
@fredkiefer will make sure to do that next time. Thanks for the heads-up.

PR #9, opened 2017-08-04
NSObjCRuntime: add availability macro definitions

theiostream:
Add empty definitions for OSX Availability macros for compatibility
purposes.

### DIFF
--- a/Headers/Foundation/NSObjCRuntime.h
+++ b/Headers/Foundation/NSObjCRuntime.h
@@ -286,13 +286,20 @@ DEFINE_BLOCK_TYPE(NSComparator, NSComparisonResult, id, id);
 /**
  * Declare Apple availability macros for compatibility purposes as no-ops.
  */
-#ifndef NS_CLASS_AVAILABLE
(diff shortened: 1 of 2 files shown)


Headers/Foundation/NSObjCRuntime.h
```diff
@@ -286,13 +286,20 @@ DEFINE_BLOCK_TYPE(NSComparator, NSComparisonResult, id, id);
 /**
  * Declare Apple availability macros for compatibility purposes as no-ops.
  */
-#ifndef NS_CLASS_AVAILABLE
-#define NS_CLASS_AVAILABLE(x, y)
-#endif
```
ivucica: I'm only concerned that we lost `#ifndef` guards. Is this intentional and justified? If so I'll merge.

theiostream: I added that macro myself some time ago and the `#ifndef` was mostly due to initial paranoia; I don't think this is ever defined anywhere else though and they seemed to clutter the code.

PR #13, opened 2017-11-18
Get compatible with Apple LLVM API

zenny-chen:
Make the Blocks API compatible with Apple LLVM.


danilaml:
Fixes https://github.com/gnustep/libobjc2/issues/46

Issue #12, opened 2017-09-27
NSJSONSerialization serializes all integral NSNumbers as signed numbers


zneak:
Per https://github.com/gnustep/libs-base/blob/master/Source/NSJSONSerialization.m#L944:

```objc
  else if ([obj isKindOfClass: NSNumberClass])
    {
      const char        *t = [obj objCType];

      if (strchr("cCsSiIlLqQ", *t) != 0)
        {
          long long     i = [(NSNumber*)obj longLongValue];

          [output appendFormat: @"%lld", i];
        }
      else
        {
          [output appendFormat: @"%.17g", [(NSNumber*)obj doubleValue]];
        }
    }
```

Regardless of whether the `NSNumber` has been created a signed or unsigned variant of `numberWith(sign)(size)`, it'll be JSON-serialized as an unsigned number. This is a problem, because negative numbers can be equal to positive numbers mostly only in C-family languages, and JSON interops with much higher-level languages.

iamleeg:
I have a fix for this but to test it would like to use `ULLONG_MAX` from `limits.h`, which is a C99 header. Does gnustep still support compiler versions where that would be a problem?

PR #15, opened 2017-11-25
fix overflow of integer types when writing large unsigned long longs …

iamleeg:
…to JSON. This is a fix for #12.


zneak:
I did mess up the original problem description: numbers were always treated as signed, not unsigned.

I don't know what your test was like, but it could be as simple as:

* JSON-serializing `[NSNumber numberWithUnsignedLongLong:0xffffffffffffffff]` and checking that it's equal to `@"18446744073709551615"`
* JSON-serializing `[NSNumber numberWithLongLong:0xffffffffffffffff]` and checking that it's equal to  `@"-1"`

iamleeg:
my test was doing a round-trip (so writing the number to JSON, reading the JSON back and checking the number), which doesn't work because numbers are scanned as `double`. Your test suggestion should be easy to build :).

davidchisnall:
Looks good to me, but the ChangeLog entry needs a merge from master to avoid the conflict.

Please also add this to the commit message:

Fixes #12

davidchisnall:
Looks fine.  I'll merge once Travis finishes testing.

PR #16, opened 2017-11-26
fix bug #51592: do not assert on nonexistent files when copying

iamleeg:
I agree with the reporter that these should not be assertions: the filesystem is not under the developer's control so it is unreasonable for the developer to make any expectations about its state. Additionally, there was a time-of-check to time-of-use issue - the method fails early if the file doesn't exist, but it might exist when the check is made and then be unlinked before opening it is attempted.

I have therefore completely removed the test for the source file's existence: if it didn't exist then, it still won't exist when we try to read its attributes. I have changed the assertion on failure to read the file's attributes into a report to the handler, so that the application can deal with the error.

Additionally I noticed when building that the method signature was wrong for `-setDelegate:` and changed that.

### DIFF
--- a/Source/NSFileManager.m
+++ b/Source/NSFileManager.m
@@ -361,7 +361,7 @@ static NSStringEncoding	defaultEncoding;
   return _delegate;
 }
 
-- (void) setDelegate: (NSFileManager *)delegate {
+- (void) setDelegate: (id <NSFileManagerDelegate>)delegate {
   _delegate = delegate;
 }
 
@@ -2821,13 +2821,15 @@ static inline void gsedRelease(GSEnumeratedDirectory X)
   int		wbytes;
   char		buffer[bufsize];
 
-  /* Assumes source is a file and exists! */
-  NSAssert1 ([self fileExistsAtPath: source],
-    @"source file '%@' does not exist!", source);
-
   attributes = [self fileAttributesAtPath: source traverseLink: NO];
-  NSAssert1 (attributes, @"could not get the attributes for file '%@'",
-    source);
+  if (attributes == nil)
+    {
+      return [self _proceedAccordingToHandler: handler
+				     forError: @"could not get attributes for source file"
+				       inPath: source
+				     fromPath: source
+				       toPath: destination];
+    }
 
   fileSize = [attributes fileSize];
   fileMode = [attributes filePosixPermissions];
--- a/Tests/base/NSFileManager/general.m
+++ b/Tests/base/NSFileManager/general.m
@@ -128,6 +128,11 @@ int main()
     PASS([str1 isEqualToString: str2],"NSFileManager moved file contents match");
   }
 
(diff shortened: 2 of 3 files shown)


Source/NSFileManager.m
```diff
@@ -2821,13 +2821,15 @@ - (BOOL) _copyFile: (NSString*)source
...
-  /* Assumes source is a file and exists! */
-  NSAssert1 ([self fileExistsAtPath: source],
-    @"source file '%@' does not exist!", source);
-
   attributes = [self fileAttributesAtPath: source traverseLink: NO];
-  NSAssert1 (attributes, @"could not get the attributes for file '%@'",
-    source);
+  if (attributes == nil)
+    {
+      return [self _proceedAccordingToHandler: handler
```
davidchisnall: This should probably call the delegate method `fileManager:shouldProceedAfterError:copyingItemAtURL:toURL:` if it exists, otherwise `fileManager:shouldProceedAfterError:copyingItemAtPath:toPath:
`, neither of which appear to be called by this method.  The accompanying `NSError` probably wants a code along the lines of `NSFileReadNoSuchFileError`.

Ideally, we would pass the `NSError` returned by `attributesOfItemAtPath:error:`.

Tests/base/NSFileManager/general.m
```diff
@@ -128,6 +128,11 @@ int main()
     PASS([str1 isEqualToString: str2],"NSFileManager moved file contents match");
   }
 
+  PASS(![mgr copyPath: @"NSFMFile"
+	       toPath: @"NSFMDestination"
+	      handler: nil],
+       "NSFileManager does not copy nonexistent file");
+
```
davidchisnall: The test isn't checking the error.  It would be good to check that we give the same error (and call the same delegate methods) as Cocoa.

Source/NSFileManager.m
iamleeg: Is that in addition to or instead of the handler method? I understand that Apple have deprecated the handler methods in Cocoa but don't know what the state is in GNUstep.

davidchisnall: The implementation of the deprecated `-copyPath:toPath:handler:` method should call back to the handler (though it's not clear what happens if the hander returns `YES` to `fileManager:shouldProceedAfterError:` - your code doesn't appear to handle it?).  The implementation of `-copyItemAtURL:toURL:error:` should set the error correctly.  I think fixing this properly needs some refactoring, though if @rfm is happy then this solution is less bad than what was there already.

Source/NSFileManager.m
```diff
@@ -361,7 +361,7 @@ - (void) dealloc
   return _delegate;
 }
 
-- (void) setDelegate: (NSFileManager *)delegate {
```
fredkiefer: I think this error has already been corrected in master. Please merge and verify.

iamleeg:
Closing as @rfm merged the patch anyway.

PR #18, opened 2017-12-11
Remove use of NSDecrementRefCountWasZero.

davidchisnall:
This function is unreliable and is impossible to implement safely in the
persence of ARC (even in other modules).  ARC requires that the runtime delete
all weak references before object deallocation and doesn't permit resurrection.

This change is initially motivated by the desire to steal a bit from the
refcount to indicate that an object has had weak references taken.

### DIFF
--- a/Tests/base/NSFileHandle/singleton.m
+++ b/Tests/base/NSFileHandle/singleton.m
@@ -21,6 +21,10 @@ int main(void)
   PASS_EXCEPTION([a release], NSGenericException, "Cannot dealloc stdin");
   PASS_EXCEPTION([b release], NSGenericException, "Cannot dealloc stdout");
   PASS_EXCEPTION([c release], NSGenericException, "Cannot dealloc stderr");
+  // The following are expected to fail with an ARC-supporting runtime.  ARC
+  // doesn't allow resurrection and the objc_retain() function will check for
+  // this so the [self retain] in GSFileHandle's -dealloc will not actually
+  // resurrect the object.
   PASS([a retainCount]> 0, "stdin persists");
   PASS([b retainCount]> 0, "stdout persists");
   PASS([c retainCount]> 0, "strerr persists");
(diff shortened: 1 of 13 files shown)


davidchisnall:
This is created for review.  With a new libobjc2 that has a patch applied to improve performance for deallocation, this now no longer crashes.  There is probably some cleanup to do on the -base / libobjc2 interface, because we do now infinite loop if you do `[super retain]` from an `NSObject` subclass's `-retain` method, so I'm not suggesting that this be committed yet.

Tests/base/NSFileHandle/singleton.m
```diff
@@ -21,6 +21,10 @@ int main(void)
   PASS_EXCEPTION([a release], NSGenericException, "Cannot dealloc stdin");
   PASS_EXCEPTION([b release], NSGenericException, "Cannot dealloc stdout");
   PASS_EXCEPTION([c release], NSGenericException, "Cannot dealloc stderr");
+  // The following are expected to fail with an ARC-supporting runtime.  ARC
+  // doesn't allow resurrection and the objc_retain() function will check for
+  // this so the [self retain] in GSFileHandle's -dealloc will not actually
+  // resurrect the object.
```
ivucica: Ah, this is why TravisCI is erroring out.

If this is true, should we not omit them with an ARC-supporting runtime? Or, should we not replace them with the opposite test?

I don't think we should intentionally break tests, especially if we agree that use of CI is useful. What do you think?

davidchisnall: Yes, I need to fix the tests before committing.  Simply not running these ones on an ARC runtime is probably fine.

fredkiefer:
David, is it OK to close this pull request?

fredkiefer:
Ping!

davidchisnall:
Yup, the new runtime APIs provide a way to implement this function.  Note; however, that the -base code is currently broken on Windows.  It relies on being able to do weak imports of functions from DLLs.  We should probably make a decision based on the runtime available at compile time and require upgrading the runtime and -base in sync on Windows to avoid this.

PR #19, opened 2017-12-13
Refactor refcount usage.

davidchisnall:
This makes it easier for the runtime to change how reference counts are
stored by removing any refcount manipulation from -base when the runtime
provides accessors.  This should have no functionality change with
existing runtimes, but will let newer runtimes drop in alternative
representations easily.

### DIFF
--- a/Tests/base/NSProxy/basic.m
+++ b/Tests/base/NSProxy/basic.m
@@ -1,6 +1,9 @@
 #import "ObjectTesting.h"
 #import <Foundation/NSAutoreleasePool.h>
 #import <Foundation/NSProxy.h>
+#if __has_include(<objc/capabilities.h>
+#include <objc/capabilities.h>
+#endif
 
 int main()
 {
@@ -24,8 +27,10 @@ int main()
   
   obj1 = [NSProxy allocWithZone:testZone];
   PASS(obj1 != nil, "%s has working allocWithZone:",prefix);
+#ifndef OBJC_CAP_ARC
   PASS(NSZoneFromPointer(obj1) == testZone, "%s uses zone for alloc",prefix);
   PASS([obj1 zone] == testZone, "%s -zone works",prefix);
+#endif
   
   PASS([obj1 hash] != 0, "%s has working -hash",prefix);
   PASS([obj1 isEqual:obj1] == YES, "%s has working -isEqual:",prefix);
--- a/Tests/base/NSProxy/basic.m
+++ b/Tests/base/NSProxy/basic.m
@@ -1,7 +1,7 @@
 #import "ObjectTesting.h"
 #import <Foundation/NSAutoreleasePool.h>
 #import <Foundation/NSProxy.h>
-#if __has_include(<objc/capabilities.h>
+#if __has_include(<objc/capabilities.h>)
 #include <objc/capabilities.h>
 #endif
 
--- a/Source/NSObject.m
+++ b/Source/NSObject.m
@@ -444,15 +444,8 @@ struct obj_layout {
 };
 typedef	struct obj_layout *obj;
 
-/**
- * Examines the extra reference count for the object and, if non-zero
- * decrements it, otherwise leaves it unchanged.<br />
- * Returns a flag to say whether the count was zero
- * (and hence whether the extra reference count was decremented).<br />
- * This function is used by the [NSObject-release] method.
- */
-inline BOOL
-NSDecrementExtraRefCountWasZero(id anObject)
+__attribute__((weak))
+BOOL objc_release_fast_no_destroy_np(id anObject)
 {
   if (double_release_check_enabled)
     {
@@ -483,6 +476,9 @@ NSDecrementExtraRefCountWasZero(id anObject)
          * have been greater than zero)
          */
         (((obj)anObject)[-1].retained) = 0;
+#  ifdef OBJC_CAP_ARC
+        objc_delete_weak_refs(anObject);
+#  endif
         return YES;
       }
 #else	/* GSATOMICREAD */
@@ -491,6 +487,9 @@ NSDecrementExtraRefCountWasZero(id anObject)
     [theLock lock];
     if (((obj)anObject)[-1].retained == 0)
       {
+#  ifdef OBJC_CAP_ARC
+        objc_delete_weak_refs(anObject);
+#  endif
         [theLock unlock];
         return YES;
       }
@@ -505,6 +504,33 @@ NSDecrementExtraRefCountWasZero(id anObject)
   return NO;
 }
 
+__attribute__((weak))
+void objc_release_fast_np(id anObject)
+{
+  if (objc_release_fast_no_destroy_np(anObject))
+    {
+      [anObject dealloc];
+    }
+}
+
+/**
+ * Examines the extra reference count for the object and, if non-zero
+ * decrements it, otherwise leaves it unchanged.<br />
+ * Returns a flag to say whether the count was zero
+ * (and hence whether the extra reference count was decremented).<br />
+ */
+inline BOOL
+NSDecrementExtraRefCountWasZero(id anObject)
+{
+  return objc_release_fast_no_destroy_np(anObject);
+}
+
+__attribute__((weak))
+size_t object_getRetainCount_np(id anObject)
+{
+  return ((obj)anObject)[-1].retained + 1;
+}
+
 /**
  * Return the extra reference count of anObject (a value in the range
  * from 0 to the maximum unsigned integer value minus one).<br />
@@ -513,7 +539,7 @@ NSDecrementExtraRefCountWasZero(id anObject)
 inline NSUInteger
 NSExtraRefCount(id anObject)
 {
-  return ((obj)anObject)[-1].retained;
+  return object_getRetainCount_np(anObject) - 1;
 }
 
 /**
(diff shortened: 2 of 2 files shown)


davidchisnall:
This is a simpler version of #18, and is enough to allow the runtime to change how reference counts are handled.  In the short term, I want to take one bit from the refcount to indicate that weak references exist to an object, which lets us avoid acquiring a lock on all object deallocations.  In the longer term, I'd like to move small (15-bit) reference counts into the `isa` pointer on 64-bit systems (where you really only have a 48-bit address space) and avoid having an extra 8-byte header for them.

davidchisnall:
There's a follow-on part that moves to using `class_createInstance` and `object_dispose` for creating and destroying objects.  In an ARC-supporting runtime, this will move all refcount manipulation into the runtime, rather than having it spread awkwardly between Foundation and libobjc.

davidchisnall:
Second commit added.  After this, we now have cleaner layering between -base and libobjc: the runtime is entirely responsible for managing object creation, reference counting, and destruction.

Tests/base/NSProxy/basic.m
```diff
@@ -1,6 +1,9 @@
 #import "ObjectTesting.h"
 #import <Foundation/NSAutoreleasePool.h>
 #import <Foundation/NSProxy.h>
+#if __has_include(<objc/capabilities.h>
```
fredkiefer: Here a closing bracket ")" is missing. This doesn't look like you ever did run this patch. I rather withdraw my approval until I have time to actually test this change myself.

davidchisnall: Ooops.  I ran the tests and then the ones that failed test for functionality that's gone away from OS X since 10.7, so I tried to disable those and made a typo.  I've now pushed a fix and all of the tests are passing for me (except the `NSNumberFormatter` one that was failing before).

rfm:
I like the approach here:
> There's a follow-on part that moves to using class_createInstance and object_dispose
> for creating and destroying objects.
For the old runtime we would implement the functions from the new runtime (based on the existing code from NSObject.m) - a cleaner separation of functionality.

davidchisnall:
Those changes are all in this set.  We now have clean layering, where -base provides the services if required, but libobjc2 can replace them.

I don't anticipate any problems, but it would be great if someone with a Linux system could test this with the weakref branch of libobjc2 as well (works fine on FreeBSD).

davidchisnall:
Any objections to merging this?

davidchisnall:
Unfortunately, the -base test suite doesn't have any tests that are compiled with `-fobjc-arc`, though the runtime has a mixture of ARC and non-ARC tests.  I don't know if `gnustep-test` supports ARC easily - the CMake stuff in libobjc2 that drives CTest will compile any file that ends with _arc with `-fobjc-arc` (it can also compile and run all of the tests in parallel, which would be nice features for `gnustep-test`) .

As an aside, while working on this, I discovered that you can do some evil things with ARC on macOS:

```objc

@interface Evil : NSObject
@end

static Evil *s;
	__weak Evil *we;
@implementation Evil
- (void)dealloc
{
	s = self;
	//[[NSException exceptionWithName: @"Exception" reason: @"Broke" userInfo: nil] raise];
}
+ (void)doEvil
{
	@autoreleasepool
	{
		Evil *e = [Evil new];
		we = e;
	}
}
@end



int main()
{
	@try {
		[Evil doEvil];
	} @catch (...) {
		NSLog(@"Caught exception!");
	}
	NSLog(@"we: %@", we);
	NSLog(@"s: %p", s);
	NSLog(@"s: %@", s);
}
```

This will increment the reference count on the assignment to `s`, but then still deallocate the object, so `s` becomes a dangling pointer.  If you uncomment the line throwing the exception, then the `s` is a strong reference, the object is not deallocated (though ivars from subclasses might be cleared), but the weak reference (`we`) is zeroed.

So I was only half-right that ARC doesn't allow resurrection: it kind-of does, but it's really dangerous.

#18 removes the cases of resurrection that I saw, so something like it is probably worth merging after the next release..

rfm:
> On 15 Dec 2017, at 10:15, davidchisnall <notifications@github.com> wrote:
> 
> Unfortunately, the -base test suite doesn't have any tests that are compiled with -fobjc-arc, though the runtime has a mixture of ARC and non-ARC tests. I don't know if gnustep-test supports ARC easily - the CMake stuff in libobjc2 that drives CTest will compile any file that ends with _arc with -fobjc-arc (it can also compile and run all of the tests in parallel, which would be nice features for gnustep-test) .

According to the documentation you can turn ARC on for individual files in a GNUmakefile.preamble
eg.

ifeq ($(OBJC_RUNTIME_LIB), ng)
file1.m_FILE_FLAGS+=-fobjc-arc
file2.m_FILE_FLAGS+=-fobjc-arc
file9.m_FILE_FLAGS+=-fobjc-arc
endif

Or you can provide a GNUmakefile to build all tests in a directory with ARC (defining GS_WITH_ARC at the start).

I suppose a nice addition might be to have some setting to automatically run all tests twice, once in ARC mode and once without ARC.  On the other hand that mauy have very little practical point (I don't think much code is expected to work in both situations), and it's a bad idea to add features that aren't genuinely useful

gnustep-tests uses gnustep-make's ability to build in parallel by default (though you can turn it off).
Running the tests is sequential to minimise confusion and keep things simple though.

ivucica:
Maybe a custom makefile rule would be good, with a TODO to rewrite it as a `gnustep-test` case?

Or `TestInfo` might be usable for this purpose:

```
In addition to being a marker, the TestInfo file is a shell script which 
is sourced before execution of each test program in its directory, typically
it is used to set up environment variables (eg. LD_LIBRARY_PATH to tell the
program where to find dynamic libraries the tests use).
```

Looks like you can override `BUILD_CMD` (taken from `gnustep-tests.in`)

```
  # Compile it if necessary.
  # Redirect errors to stdout so it shows up in the log,
  # but not in the summary.
  if test "$NEEDBUILD" = "yes"
  then
    echo "Building $dir/$TESTFILE"
    echo "$BUILD_CMD"
    if test -r ./make-check.env
    then
      ( . ./make-check.env; . ./TestInfo; $BUILD_CMD) 2>&1
    else
      ( . ./TestInfo; $BUILD_CMD) 2>&1
    fi
    BUILDSTATUS=$?
  else
    BUILDSTATUS=0
  fi
```

`BUILD_CMD` will be exposed to `TestInfo`, and because `TestInfo` is sourced in the same subshell as where `$BUILD_CMD` is used, it will also be overridable.

Similar situation if a makefile is present; in that case, `MAKE_CMD` can be overridden.

I certainly won't be blocking the merge for missing tests, I'm just wondering if it would make sense to cover this. 😀

davidchisnall:
> I suppose a nice addition might be to have some setting to automatically run all tests twice, once in ARC mode and once without ARC.  On the other hand that mauy have very little practical point (I don't think much code is expected to work in both situations), and it's a bad idea to add features that aren't genuinely useful

It might be interesting to have an option to run all of the tests through the ARC migration tool, and then run the ones where the tool doesn't report any errors.  

> gnustep-tests uses gnustep-make's ability to build in parallel by default (though you can turn it off).

I've seen the parallel build, but it's not ideal for a few reasons:

- It seems to only support parallelism within a directory.
- The layering between it and the main GNUmakefiles mean that I don't get the tests as normal targets, so rebuilding a single test is very hard (contrast with CTest used by libobjc, I can do `{make,ninja} Tests/{test name}` and get a single test (plus the main library as a dependency) rebuilt, which makes testing round trips a lot faster.

> Running the tests is sequential to minimise confusion and keep things simple though.

Reporting the results sequentially minimises confusion.  Most other test frameworks (CTest, llvm-lit, whatever) support running each test in parallel with redirected `std{in,out,err}` and only reporting the test output when the test fails.  For example, for libobjc2, I can run `$ ctest -j32 --output-on-failure` and it will run 32 tests in parallel and show me the output for any failing ones (without any interleaving of output).  

Most tests small and are I/O-bound from the process startup, so you get quite a significant speedup from running them in parallel.

rfm:
> On 14 Dec 2017, at 10:43, davidchisnall <notifications@github.com> wrote:
> 
> Those changes are all in this set. We now have clean layering, where -base provides the services if required, but libobjc2 can replace them.

I meant that the wrapper round the gnu runtime (in Source/ObjectiveC2) which generaly lets us use the modern runtime API even when we are using the gnu runtime, could be extended to provide these new functions, so the conditional compilation in NSObject.m could be simplified.
NB. I don't suggest we should do that work now, just that it would be a nice cleanup in future.

davidchisnall:
@rfm The problem with putting it in ObjectiveC2 is that we still need these code paths for older (actually, all released so far) versions of libobjc2.

@fredkiefer Can you confirm that you tested this on Linux with the the weakref branch from libobjc2?  Andreas Fink is reporting that the Linux runtime linker isn't picking the strong symbols from libobjc2 to override the weak ones in -base.

fredkiefer:
@davidchisnall I only tested with the old gcc runtime and would have to recompile everything with the unreleased version of libobjc2. Maybe I could find time for that tomorrow night, but even that is unclear.
How would I know which version of the functions gets actually used?

PS: Could you please add an entry in the ChangeLog file?

davidchisnall:
It looks as if the Linux run-time linker gives up when it finds the first symbol def, even if it's weak.  I can work around this by doing some run-time checks in the code.  

I forgot we were doing ChangeLog entries - I'll add one once I've made this work properly with stupid linkers.

davidchisnall:
25e67bd1074e9be09bccc90359cbf10a1c304ddd should fix the issue with Linux.  It should now be possible to build -base with the ng library option and switch between old and new versions of the runtime (I have run the test suite with both the version from the FreeBSD package and built from trunk), and get much better performance from deallocation on multicore systems with the new one.

PR #22, opened 2018-04-30
add +[NSCalendar calendarWithIdentifier:]

iamleeg:


davidchisnall:
This crash is expected in released versions of clang.  It's there to catch a regression for a bug that's fixed in the release version of clang.  I don't know why the GNUstep CI builds are building the runtime with tests enabled and especially why it's aborting if the runtime tests don't build.

PR #32, opened 2018-12-31
Issue 31: Use pkg-config to detect icu

hughmcmaster:
This patchset uses `pkg-config` to detect icu, as `icu-config` is deprecated by upstream ICU and will not be installed by default in icu 64.


ivucica:
I've merged this.

PR #33, opened 2019-02-12
GSTLS.h: Change guard to check for /value/ of GS_USE_GNUTLS.

ivucica:
Previously, code checked for whether its defined or not. This
preprocessor  constant might be defined with a zero value
indicating desire not to use GnuTLS, which would still pass the
old check.

Hey, @rfm, can you take a look if this change makes sense? It helps when crosscompiling for a platform that has no GnuTLS, and I /think/ (but didn't check) it won't break our usual platforms.


rfm:
Looks right to me, thanks.

PR #34, opened 2019-03-11
Minor fix

cwchentw:
Fix some function declaration that prevent libs-base from compilation on Windows.

### DIFF
--- a/Source/GSSocketStream.m
+++ b/Source/GSSocketStream.m
@@ -56,7 +56,7 @@
 #endif
 
 #ifdef _WIN32
-extern const char *inet_ntop(int, const void *, char *, size_t);
+extern const char *inet_ntop(int, void *, char *, size_t);
 extern int inet_pton(int , const char *, void *);
 #define	OPTLEN	int
 #else
--- a/Source/NSHost.m
+++ b/Source/NSHost.m
@@ -41,7 +41,7 @@
 #if defined(_WIN32)
 #include <winsock2.h>
 #include <ws2tcpip.h>
-extern const char *inet_ntop(int, const void *, char *, size_t);
+extern const char *inet_ntop(int, void *, char *, size_t);
 extern int inet_pton(int , const char *, void *);
 #else
 #include <netdb.h>
--- a/Source/inet_ntop.m
+++ b/Source/inet_ntop.m
@@ -58,7 +58,7 @@ static  __attribute__((unused)) const char *inet_ntop6(const u_char *src, char *
 const char *
 inet_ntop(af, src, dst, size)
         int af;
-        const void *src;
+        void *src;
         char *dst;
         size_t size;
 {


gcasa:
Please take a look at this pull request….

ivucica:
Windows docs say it should be a `const void*`.

```
PCSTR WSAAPI inet_ntop(
  INT        Family,
  const VOID *pAddr,
  PSTR       pStringBuf,
  size_t     StringBufSize
);
```

(Source https://docs.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-inet_ntop )

@cwchentw What kind of a compile time error are you running into? Which compiler and which environment (e.g.: clang/gcc, and mingw32/mingw64/msvc/cygwin)? Include versions. I suspect the headers you're using are buggy.

ivucica:
It also seems odd that we're reimplementing `inet_ntop` and using that on Windows, when the platform offers a native implementation of the function.

cwchentw:
I compiled GNUstep on MinGW 64 bit shell of MSYS2. `pacman -Qo` failed to show the version of the library (libc?). MSYS2 seems holding a rolling release policy as in Arch Linux.

fsb4000:
@ivucica 
>It also seems odd that we're reimplementing inet_ntop and using that on Windows, when the platform offers a native implementation of the function.

That is not odd. 
inet_ntop exists since Windows Vista. But GNUStep shoud works from Windows 2000(or XP?). So you're reimplementing the function for support Windows XP and 2000

https://docs.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-inet_ntop

fsb4000:
By the way, 
classic MinGW 8.2.0 (without w64 and MSYS2) 
```
didn't have inet_ntop prototype.
```
Msys2 and mingw-w64 have the prototype: 
```
WINSOCK_API_LINKAGE LPCSTR WSAAPI inet_ntop(INT Family, PVOID pAddr, LPSTR pStringBuf, size_t StringBufSize);
```
Visual Studio 2017 have the prototype:
```
PCSTR
WSAAPI
inet_ntop(
    _In_                                INT             Family,
    _In_                                const VOID *    pAddr,
    _Out_writes_(StringBufSize)         PSTR            pStringBuf,
    _In_                                size_t          StringBufSize
    );
```

fsb4000:
Maybe the patch will be better?
```
...
#ifdef _WIN32
#ifdef __MINGW64__
extern const char *inet_ntop(int, void *, char *, size_t);
#else
extern const char *inet_ntop(int, const void *, char *, size_t);
#endif
 extern int inet_pton(int , const char *, void *);
...
```
What do you think?

rfm:
Almost everywhere uses the 'const' version (posix/ansi, gnu/linux, windows), so I think this should be viewed as a workaround for a bug in the msys2 headers.
The proposed '#ifdef __MINGW64__' approach seems the right one to me (though I don't know if __MINGW64__ covers all the cases where there's a faulty header to work around).

gcasa:
@rfm rfm Do you approve of merging this?

rfm:
Not as it stands.  As I understand it the patch works around a bug in a particular mingw version, but might introduce a bug when using other versions.  It sounds to me like making the changes conditional on #ifdef __MINGW64__ would be good, but I was hoping someone else could confirm.

fsb4000:
I did some tests with many mingw installations:
 Command: gcc -dM -E 1.c >1.txt

Results:
mingw64 32bit without any includes: https://pastebin.com/tRcRzcfw
mingw64 64bit without any includes: https://pastebin.com/CU3G3qAR
MinGW 32bit without any includes:  https://pastebin.com/WK2CVcNx

mingw64 32bit with stdlib.h: https://pastebin.com/U3GikjB1
mingw64 64bit with stdlib.h: https://pastebin.com/Bkf5hQ2W
MinGW 32bit with stdlib.h:  https://pastebin.com/BcNbLgiP

**rfm**, you're right **MINGW64** didn't exist for mingw64 32 bit. But I think you could use:

**__MINGW64_VERSION_MAJOR**

instead.  It exist in both 
mingw64 32bit with stdlib.h 
and
mingw64 64bit with stdlib.h
but doesn't exist in
MinGW 32bit with stdlib.h

gcasa:
Can anyone give me a reason to keep this open?  Otherwise I'm going to close it.

gcasa:
Is there any reason to leave this open?  I've asked this twice now.... if not we should close it.

rfm:
I don't have a windows system to try things out on.  Presumably there *is* an issue for windows users, so closing this doesn't seem right, but if nobody is actually trying it out and providing an updated patch to work with the expected versions of windows, I don't know what else makes sense.

gcasa:
> I don't have a windows system to try things out on. Presumably there _is_ an issue for windows users, so closing this doesn't seem right, but if nobody is actually trying it out and providing an updated patch to work with the expected versions of windows, I don't know what else makes sense.

I will fire up one of my windows virtual machines and try it.

rmottola:
> Not as it stands. As I understand it the patch works around a bug in a particular mingw version, but might introduce a bug when using other versions. It sounds to me like making the changes conditional on #ifdef **MINGW64** would be good, but I was hoping someone else could confirm.

In a parallel fix on windows (which is awaiting PR) we proved that MINGW64 is not the way to go to support different MSYS versions and 32bit vs 64bit.
I could try to "extend" the hack to this too, but perhaps in this case it is best to use a configure to detect if inet_ntop is present or not and avoid reimplementing it?

fredkiefer:
> In a parallel fix on windows (which is awaiting PR) we proved that MINGW64 is not the way to go to support different MSYS versions and 32bit vs 64bit.
> I could try to "extend" the hack to this too, but perhaps in this case it is best to use a configure to detect if inet_ntop is present or not and avoid reimplementing it?

@fsb4000 already noted that and suggested to use __MINGW64_VERSION_MAJOR instead. But we should rethink more of this patch. First for MSYS2 we really don't need to declare our own version of these functions as they are already present there. What seems to go wrong is detecting these functions, at least inet_pton, this could be because this is a #define to a different function, InetptonA, the same may be true for inet_ntop. We could remove this detection for MSYS2. Here it really would help if somebody with a Windows system would report the results of the configuration for different setups.

rmottola:
__MINGW64_VERSION_MAJOR is *not* the way to go in my tests, if you want to support Mingw2-32bit! Generally MINGW64 should be used as little as possible.
I agree with you, we should test better.

Furthermore, looking in the patch, I see that other headers include winsock2.h, we should not and use GSConfig.h which is now "fixed" where possible, I already fixed one place in my last patch. That way we are consistent with the order of includes.

fredkiefer:
Here are the configure results from another pull request on MINGW64:

```
checking for library containing inet_ntop... no
checking for gethostbyaddr_r... no
checking for inet_aton... no
checking for inet_pton... no
checking for inet_ntop... no
```
We really need to fix these checks.

triplef:
> What seems to go wrong is detecting these functions, at least inet_pton, this could be because this is a #define to a different function, InetptonA, the same may be true for inet_ntop. We could remove this detection for MSYS2. Here it really would help if somebody with a Windows system would report the results of the configuration for different setups.

Looking at the configure output from a recent CI build [confirms this](https://travis-ci.org/github/gnustep/libs-base/jobs/695350143#L753):

> checking for inet_pton... no

`inet_pton` is directly [provided by Windows Sockets](https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-inetptonw) in `ws2tcpip.h`:

> When UNICODE or _UNICODE is not defined, InetPton is defined to InetPtonA, the ANSI version of this function. The ANSI version of this function is always defined as inet_pton.

Maybe we can do something similar as these MSYS2 packages and tweak the configure check to be able to find the function:

- https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python/2010-configure-have-inet-pton.patch
- https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-vlc/0021-fix-inet_pton-search.patch

rmottola:
I have completely rewritten the concept of this patch to be configure based. The tests were wrong. It works on MSYS1-32bit and MSYS2-64bit for me, but still cause a redefinition clash on MSYS2-64bit. Better discuss this on the specific msys-fix branch.

fredkiefer:
I am closing this PR as the new one by Riccardo #146 addresses this issue better.

PR #37, opened 2019-03-29
Decrease ICU version requirements to 49

fsb4000:
to support Windows 2000.


fredkiefer:
This change looks fine to me. We don't loose to much with it and would support Windows 2000. If you could add a ChangeLog entry for the changes and if Richard does not oppose, I am willing to merge this patch.

fsb4000:
>If you could add a ChangeLog entry for the changes

Thank you. I added.

PR #38, opened 2019-04-12
NSString and NSCharacterSet changes

gcasa:
Incorporates changes from 10.9 for URL character sets and percent encoding and decoding.   I have written tests for both of these.  Please give any feedback you feel pertinent.  Thanks.  GC


Headers/Foundation/NSString.h
```diff
@@ -802,6 +802,11 @@ typedef NSUInteger NSStringEncodingConversionOptions;
 - (const char *)UTF8String;
 #endif
 
+// #if OS_API_VERSION(MAC_OS_X_VERSION_10_9,GS_API_LATEST)
```
fredkiefer: Why are these guards commented out?

Source/NSCharacterSet.m
```diff
@@ -537,7 +543,7 @@ - (void) removeCharactersInString: (NSString*)aString
 
 
 /* A simple array for caching standard bitmap sets */
-#define MAX_STANDARD_SETS 15
+#define MAX_STANDARD_SETS 20
```
fredkiefer: Shouldn't this be 21?

Source/NSString.m
```diff
@@ -1886,6 +1922,70 @@ - (void) getCharacters: (unichar*)buffer
...
+{
+  NSData *data = [self dataUsingEncoding: NSUTF8StringEncoding];
+  NSString *result = nil;
+  char *s = (char *)[data bytes];
+  NSUInteger slen = 0; 
+  char *o = NULL;
+
+  // Allocate memory...
+  slen = strlen(s);
+  o = (char *)NSZoneMalloc(NSDefaultMallocZone(), slen * 3);
```
fredkiefer: Why three times the size of the original? The decoded length should always be at most the encoded one.

Tests/base/NSCharacterSet/general.m
```diff
@@ -88,8 +88,49 @@ int main()
...
-  
+  theSet = [NSCharacterSet URLFragmentAllowedCharacterSet];
+  NSString *setString = @"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ?/:@-._~!$&'()*+,=";
+  NSUInteger i = 0;
+  for(i = 0; i < [setString length]; i++)
+    {
+      char c = [setString characterAtIndex: i];
+      if([theSet characterIsMember: c] == NO) // , [msg cStringUsingEncoding: NSUTF8Encoding]);
+	{
+	  NSLog(@"%c is not in set", c);
```
fredkiefer: This doesn't look like a real automated test. It should fail on wrong results not just print a message.

Tests/base/NSString/test09.m
```diff
@@ -0,0 +1,59 @@
...
+  PASS(testUrlCharacterSetEncoding(urlDecodedString, urlEncodedString, allowedCharacterSet), "fragmentCharacterSet");  
+  
+  NSLog(@"Test3");
+  urlDecodedString = @"All alphabetic characters should be encoded. Symbols should not be: !@#$%^&*()_+-=";
+  urlEncodedString = @"%41%6C%6C %61%6C%70%68%61%62%65%74%69%63 %63%68%61%72%61%63%74%65%72%73 %73%68%6F%75%6C%64 %62%65 "
+    @"%65%6E%63%6F%64%65%64. %53%79%6D%62%6F%6C%73 %73%68%6F%75%6C%64 %6E%6F%74 %62%65: !@#$%^&*()_+-=";
+  allowedCharacterSet = [[NSCharacterSet alphanumericCharacterSet] invertedSet];
+  PASS(testUrlCharacterSetEncoding(urlDecodedString, urlEncodedString, allowedCharacterSet), "inverted");  
+
+  // NSLog(@"Test4");
```
fredkiefer: Why ist this commented out?

gcasa: Because multibyte Unicode isn’t supported by gcc

triplef: I wonder if there is a feature check pragma that this could be wrapped in so the test would be enabled on compilers that to support it.

Headers/Foundation/NSString.h
gcasa: Because with them it wasn't seeing the selectors.   I'm not sure what is not defined.   Will come back to this in a bit.

Source/NSString.m
gcasa: Fixed

Tests/base/NSCharacterSet/general.m
gcasa: Fixed

Source/NSCharacterSet.m
gcasa: fixed

Tests/base/NSString/test09.m
gcasa: Added this back in.  It simply gives a warning.

Headers/Foundation/NSString.h
gcasa: Re-added these into the file.  Re-tested and working.

gcasa:
I'm merging this since there is no more feedback to respond to.   All tests are passing (locally).

PR #40, opened 2019-05-07
Fixed libdispatch configuration switch.

triplef:
Passing --enable-libdispatch was incorrectly disabling libdispatch.


fredkiefer:
Looks fine to me, the only issue I see is that the content of "action if given" is just the default behaviour. We could as well leave this blank.

PR #41, opened 2019-05-07
Add compatibility for Swift corelibs libdispatch release

triplef:
This updates the libdispatch runloop integration to be compatible with Apple’s [Swift corelibs libdispatch](https://github.com/apple/swift-corelibs-libdispatch) release, which to my knowledge is the most up-to-date and actively maintained release of libdispatch.

In that release, the main queue handle and drain functions have been renamed with a "_4CF" (for CoreFoundation) suffix and have moved to private.h, so we now check for the existance of this header and function names.

Note that libdispatch must be compiled with `INSTALL_PRIVATE_HEADERS=YES`. Please let me know if there is a good place to document this.

Also fixes the checks for the `HAVE_LIBDISPATCH_RUNLOOP` define (was inverted) and ensures that both the handle and drain functions are available (as expected by NSRunLoop.m).

### DIFF
--- a/Headers/GNUstepBase/config.h.in
+++ b/Headers/GNUstepBase/config.h.in
@@ -216,6 +216,9 @@
 /* Define to 1 if you have the `dispatch_main_queue_drain_np' function. */
 #undef HAVE_DISPATCH_MAIN_QUEUE_DRAIN_NP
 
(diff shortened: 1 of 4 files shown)


triplef:

PR #42, opened 2019-05-07
Added basic spinlock implementation using builtins.

triplef:
This will be used on platforms without pthread_spin_lock(), e.g. Android targeting API level < 24.

The implementation lifted from [libobjc2’s spinlock.h](https://github.com/gnustep/libobjc2/blob/master/spinlock.h). Please let me know if this requires attribution and where it should go.


triplef:

PR #43, opened 2019-05-08
Always use syslog for NSLog on Android

triplef:
As there is no way to access stdout/stderr on Android, and syslog is available and outputs to the system log (accesible using "adb logcat"), this improves the developer experience when developing for Android by enabling NSLog output without having to set the "GSLogSyslog" flag.

### DIFF
--- a/Source/NSLog.m
+++ b/Source/NSLog.m
@@ -181,8 +181,10 @@ _NSLog_standard_printf_handler(NSString* message)
 #else
 
 #if	defined(HAVE_SYSLOG)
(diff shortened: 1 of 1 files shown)


triplef:
To further improve the experience, all calls to `fprintf(stderr, ...)` to output error logs could be replaced by `NSLog()`. Let me know if this is something you’d be open to merge.

triplef:

rfm:
Looks good ... merging syslog commit

PR #45, opened 2019-05-23
Added support for asset loading on Android.

triplef:
This requires passing the Android activity's AssetManager object from Java to GNUstep by calling `+[NSBundle setJavaAssetManager:withJNIEnv:]`, which then enables the following features:

- NSBundle main bundle resource paths support for Android assets, e.g. for `pathForResource:ofType:`, `URLForResource:ofType:` and related methods.
- NSBundle main bundle info dictionary support if Info.plist exists in Android assets.
- `-initWithContentsOfFile:` and related methods support for reading Android assets from main bundle in various classes (e.g. NSData, NSDictionary, NSArray, etc.) by patching NSData `readContentsOfFile()` backbone method.
- NSFileManager `fileExistsAtPath:(isDirectory:)` and `isReadableFileAtPath:` return YES for main bundle asset / asset directory paths.
- NSFileHandle support for reading Android assets from main bundle.
- NSDirectoryEnumerator support for enumerating Android assets from main bundle. Note that recursion into subdirectories is currently not supported by the native Android asset manager API (see https://issuetracker.google.com/issues/37002833).

Also adds support for automatic NSProcessInfo initialization on Android with a fake executable path `/data/data/<app identifier>/exe` (as Android apps don't have a real executable path), and tweaks main bundle initialization to allow that path. Main bundle resource paths are prefixed by `/data/data/<app identifier>/Resources`.

I’d appreciate anyone’s feedback on this, and especially from people who previously worked with GNUstep on Android (@ivucica, @jordo). Thanks!

### DIFF
--- a/Source/NSBundle.m
+++ b/Source/NSBundle.m
@@ -3287,6 +3287,11 @@ IF_NO_GC(
 
 + (AAsset *)assetForPath:(NSString *)path
 {
+  return [self assetForPath:path withMode:AASSET_MODE_UNKNOWN];
+}
+
++ (AAsset *)assetForPath:(NSString *)path withMode:(int)mode
+{
   AAsset *asset = NULL;
   
   if (_assetManager && _mainBundle)
@@ -3298,7 +3303,7 @@ IF_NO_GC(
       NSString *assetPath = [path substringFromIndex:[resourcePath length]+1];
 
       asset = AAssetManager_open(_assetManager,
-        [assetPath fileSystemRepresentation], AASSET_MODE_BUFFER);
+        [assetPath fileSystemRepresentation], mode);
     }
   }
   
--- a/Source/NSBundle.m
+++ b/Source/NSBundle.m
@@ -241,6 +241,11 @@ static NSString	*library_combo =
   nil;
 #endif
 
+#ifdef __ANDROID__
+static jobject _jassetManager = NULL;
+static AAssetManager *_assetManager = NULL;
+#endif
+
 
 /*
  * Try to find the absolute path of an executable.
@@ -1382,6 +1387,7 @@ _bundle_load_callback(Class theClass, struct objc_category *theCategory)
 	  isNonInstalledTool = YES;
 	}
 
+#ifndef __ANDROID__ /* don't check suffix on Android's fake executable path */
       if (isApplication == YES)
 	{
 	  s = [path lastPathComponent];
@@ -1422,6 +1428,7 @@ _bundle_load_callback(Class theClass, struct objc_category *theCategory)
 		}
 	    }
 	}
+#endif /* !__ANDROID__ */
 
       if (isApplication == NO)
 	{
@@ -1805,9 +1812,6 @@ IF_NO_GC(
     {
       NSString		*identifier = [self bundleIdentifier];
       NSUInteger        count;
-      NSUInteger	plen = [_path length];
-      NSEnumerator	*enumerator;
-      NSString		*path;
 
       [load_lock lock];
       if (_bundles != nil)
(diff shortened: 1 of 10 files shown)


triplef:
I realized that currently asset reading will be restricted to threads that are already attached to the JavaVM, e.g. it will crash the app when used from threads created using NSThread.

I would like to request holding off merging this until I have resolved this.

gcasa:
I have made a request for a current and future copyright assignment for algoriddim to the FSF so it should be okay to incorporate their changes into the library going forward.

gcasa:
I have copied all of you on the copyright assignment request.    Please proceed with review and integration.   I have already spoken to Frederick.

Source/NSBundle.m
```diff
@@ -2144,6 +2117,48 @@ + (NSArray *) _bundleResourcePathsWithRootPath: (NSString*)rootPath
...
+      primary = [primary stringByAppendingPathComponent:
+        [localization stringByAppendingPathExtension:@"lproj"]];
+      contents = bundle_directory_readable(primary);
+      addBundlePath(array, contents, primary, nil, nil);
+    } else {
+      NSString *subPathPrimary = primary;
+      enumerate = [languages objectEnumerator];
+      while ((language = [enumerate nextObject])) {
+        primary = [subPathPrimary stringByAppendingPathComponent:
+          [localization stringByAppendingPathExtension:@"lproj"]];
```
fredkiefer: Shouldn't this be "language" instead of "localization" here?

Source/NSBundle.m
```diff
@@ -2144,6 +2117,48 @@ + (NSArray *) _bundleResourcePathsWithRootPath: (NSString*)rootPath
...
+  if (localization) {
+    primary = [originalPrimary stringByAppendingPathComponent:
+      [localization stringByAppendingPathExtension:@"lproj"]];
+    contents = bundle_directory_readable(primary);
+    addBundlePath(array, contents, primary, nil, nil);
+  } else {
+    enumerate = [languages objectEnumerator];
+    while ((language = [enumerate nextObject])) {
+      primary = [originalPrimary stringByAppendingPathComponent:
+        [localization stringByAppendingPathExtension:@"lproj"]];
```
fredkiefer: Here as well...

Source/NSData.m
```diff
@@ -253,6 +253,35 @@
       NSWarnFLog(@"Open (%@) attempt failed - bad path", path);
       return NO;
     }
+  
+#ifdef __ANDROID__
+  // Android: try using asset manager if path is in main bundle resources
+  AAsset *asset = [NSBundle assetForPath:path];
```
fredkiefer: As all the following operations are already present at GSFileHandle wouldn't it be better to reuse these?

Source/NSBundle.m
triplef: Good catch, thanks. Fixed.

Source/NSBundle.m
triplef: Fixed.

Source/NSData.m
triplef: We probably could, but it would have two downsides:

- As NS/GSFileHandle doesn’t have a way to pass in an AAsset object, we’d have to open the AAsset twice: once to check if we should take that code path (as currently done), and then once more in GSFileHandle.
- Also, we now use [`AASSET_MODE_BUFFER`](https://developer.android.com/ndk/reference/group/asset#group___asset_1gga16af7b253440dadd46a80a4b9fddba4da40ec098f4afb7c2869fa449d3059f6bb) in NSData, whereas GSFileHandle needs to use `AASSET_MODE_RANDOM` to support random access, so it might be slightly slower to use GSFileHandle (although I have not profiled this).

Let me know if you still think it’d be better to use GSFileHandle here.

triplef:
Thanks for your feedback Fred.

I have now also updated the [Android toolchain](https://github.com/gnustep/tools-android) project to contain the current version of this patch, and the [Android Studio example](https://github.com/gnustep/android-examples) to utilize it by passing the Asset Manager to GNUstep and reading some assets. This should make it somewhat easier to try this out.

I also started looking into supporting localized strings with this, and found that it works by manually settings the "userLanguages" in NSUserDefaults (I guess because we don’t have langinfo on Android?), as this will enable the NSBundle resource paths backbone to find the corresponding localizable strings (also demoed in example app). Not sure if there’s a better way to support this automatically.

Finally there are two more places I’d like to extend with assets support (will look into this next week):

- Extend NSFileManager _copyFile:toFile:handler: to support copying from Android assets.
- Look into extending NSFileManager file attributes (GSAttrDictionary) to support Android assets in some basic way so that code paths depending on file attributes (e.g. _copyFile:...) do not have to be changed as much.

Please let me know if you can think of any other methods that would be good to support with this.

triplef:
I have added support for Android assets in NSFileManager _copyFile:toFile:handler: as well as basic support for file attributes of assets (they will report the correct file size and a posix permissions of 400 / owner read-only, but otherwise zeroed stat() values).

With this all open issues would be resolved on my end (ready to merge). Let me know if you want me to squash the commits into one, or if there’s any further feedback.

gcasa:
@rfm @fredkiefer Please review this for inclusion.

triplef:
If possible it’d also be great if @ivucica could look over this with his prior experience using GNUstep on Android.

rfm:
Having looked at these changes and made cosmetic modifications to match coding styles, I merged this.

triplef:
@rfm thanks for merging, much appreciated.

PR #57, opened 2019-07-24
Added generics support to NSHashTable

triplef:
This adds support for using generics with NSHashTable as on Apple platforms.

### DIFF
--- a/Headers/Foundation/NSHashTable.h
+++ b/Headers/Foundation/NSHashTable.h
@@ -31,6 +31,7 @@
 
 /**** Included Headers *******************************************************/
 
+#import <Foundation/NSObject.h>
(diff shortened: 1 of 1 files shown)


fredkiefer:
Shouldn't we change the implementation file as well so that a compiler may check the consitency?

triplef:
Yeah I thought the same but didn’t since it hadn’t been done for any of the other generic classes like NSArray, and I wasn’t sure why. I’ll update the PR.

triplef:
So I didn’t know this before, but generics are interface-only and cannot be used in the method signatures of the implementations.

I couldn’t find any official documentation from Apple on lightweight generics, but you can try it yourself quite easily that the compiler won’t let you write a method definition with a generic type in the method signature.

PR #58, opened 2019-08-02
NSBlockOperation changes

gcasa:
Putting this in for review.  Still need to add tests to this branch, but I want Fred and others to look at it to see if there is any obvious miss.

### DIFF
--- a/Headers/Foundation/NSOperation.h
+++ b/Headers/Foundation/NSOperation.h
@@ -37,7 +37,8 @@ extern "C" {
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
 #import <GNUstepBase/GSBlocks.h>
 DEFINE_BLOCK_TYPE_NO_ARGS(GSOperationCompletionBlock, void);
-#endif
+DEFINE_BLOCK_TYPE(GSBlockOperationBlock, void, void);
+#endif  
 
 @class NSMutableArray;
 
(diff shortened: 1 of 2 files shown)


fredkiefer:
All travis builds are failing. Please fix these first.

Headers/Foundation/NSOperation.h
```diff
@@ -37,7 +37,8 @@ extern "C" {
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
 #import <GNUstepBase/GSBlocks.h>
 DEFINE_BLOCK_TYPE_NO_ARGS(GSOperationCompletionBlock, void);
-#endif
+DEFINE_BLOCK_TYPE(GSBlockOperationBlock, void, void);
```
fredkiefer: Why didn't you use DEFINE_BLOCK_TYPE_NO_ARGS here?

gcasa: Didn't know of it's existence. :)

Source/NSOperation.m
```diff
@@ -533,6 +533,56 @@ - (void) _finish
...
+	  
+@implementation NSBlockOperation
+
+// Initialize
+- (id) init
+{
+  self = [super init];
+  if(self != nil)
+    {
+      _executionBlocks = [[NSMutableArray alloc] initWithCapacity: 10];
```
fredkiefer: I think 1 would be a more likely capacity, as in most cases we will have only one block.

Source/NSOperation.m
```diff
@@ -533,6 +533,56 @@ - (void) _finish
...
+- (void) dealloc
+{
+  RELEASE(_executionBlocks);
+  [super dealloc];
+}
+
+// Managing the blocks in the Operation
++ (instancetype)blockOperationWithBlock: (GSBlockOperationBlock)block
+{
+  NSBlockOperation *op = [[NSBlockOperation alloc] init];
```
fredkiefer: Using self here instead of NSBlockOperation would allow subclasses to reuse the code more easily. But then it is highly unlikely that there ever will be a subclass.

gcasa: Yep... I will change that.  Thanks.

gcasa: Done

PR #59, opened 2019-08-02
NSSortDescriptor branch

gcasa:
Added the ability to use a comparator the NSSortDescriptor.   Need to add tests for this class as well, but wanted to get it into review.

### DIFF
--- a/Source/NSSortDescriptor.m
+++ b/Source/NSSortDescriptor.m
@@ -162,6 +162,7 @@ static BOOL     initialized = NO;
 - (void) dealloc
 {
   TEST_RELEASE(_key);
+  TEST_RELEASE(_comparator);
   [super dealloc];
 }
 
@@ -222,7 +223,7 @@ static BOOL     initialized = NO;
 
       ASSIGN(_key, key);
       _ascending = ascending;
-      _comparator = cmptr;
+      ASSIGN(_comparator, cmptr);
 
       return self;
     }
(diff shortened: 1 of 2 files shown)


Source/NSSortDescriptor.m
```diff
@@ -167,11 +186,52 @@ + (id) sortDescriptorWithKey: (NSString *)aKey
...
+      if (cmptr == NULL)
+        {
+          [NSException raise: NSInvalidArgumentException
+                      format: @"%@", _(@"Passed NULL comparator when initializing "
+                                       @"an NSSortDescriptor.")];
+        }
+
+      ASSIGN(_key, key);
+      _ascending = ascending;
+      _comparator = cmptr;
```
fredkiefer: Shouldn't we somehow retain the NSBlock used for comparison? And if so, we will need a matching release in the dealloc method.

gcasa: Done

Source/NSSortDescriptor.m
```diff
@@ -113,8 +113,16 @@ - (NSComparisonResult) compareObject: (id) object1 toObject: (id) object2
   id comparedKey1 = [object1 valueForKeyPath: _key];
   id comparedKey2 = [object2 valueForKeyPath: _key];
 
-  result = (NSComparisonResult) [comparedKey1 performSelector: _selector
-                                                   withObject: comparedKey2];
+  if(_comparator != NULL)
```
fredkiefer: The code would be easier to read if you used the same order in both checks. That is either (_comparator != NULL) or (_comparator == NULL) both here and in line 149. Also a space after the "if" is required.

PR #61, opened 2019-08-05
NSFileManager branch

gcasa:
Add methods to NSFIleManager URLForDirectory:... enumeratorAtURL....

### DIFF
--- a/Source/NSFileManager.m
+++ b/Source/NSFileManager.m
@@ -179,6 +179,17 @@
               followSymlinks: (BOOL)follow
                 justContents: (BOOL)justContents
 			 for: (NSFileManager*)mgr;
+
+- (id) initWithDirectoryPath: (NSString*)path
+   recurseIntoSubdirectories: (BOOL)recurse
+	      followSymlinks: (BOOL)follow
+		justContents: (BOOL)justContents
+                  skipHidden: (BOOL)skipHidden
+                errorHandler: (GSDirEnumErrorHandler) handler
+			 for: (NSFileManager*)mgr;
+
+- (void) _setSkipHidden: (BOOL)flag;
+- (void) _setErrorHandler: (GSDirEnumErrorHandler) handler;
 @end
 
 /*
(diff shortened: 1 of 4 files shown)


Source/NSFileManager.m
```diff
@@ -773,6 +784,65 @@ - (NSArray*) contentsOfDirectoryAtURL: (NSURL*)url
...
+    }
+  else if(directory == NSItemReplacementDirectory)
+    {
+      result = [NSURL URLWithString: NSTemporaryDirectory()];
+    }
+
+  // If we should created it, create it...
+  if (shouldCreate)
+    {
+      [self       createDirectoryAtPath: [result absoluteString]
```
fredkiefer: The formatting here is strange.

Source/NSFileManager.m
```diff
@@ -773,6 +784,65 @@ - (NSArray*) contentsOfDirectoryAtURL: (NSURL*)url
...
+            withIntermediateDirectories: YES
+                             attributes: nil
+                                  error: NULL];
```
fredkiefer: You should at least pass on the error.

Source/NSFileManager.m
```diff
@@ -773,6 +784,65 @@ - (NSArray*) contentsOfDirectoryAtURL: (NSURL*)url
...
+    {
+      result = [NSURL URLWithString: [urlArray objectAtIndex: 0]];
+    }
+  else if(directory == NSItemReplacementDirectory)
+    {
+      result = [NSURL URLWithString: NSTemporaryDirectory()];
+    }
+
+  // If we should created it, create it...
+  if (shouldCreate)
```
fredkiefer: Shouldn't we first test whether the directory exists?

fredkiefer:
I am not sure whether the overall structure of the change is correct. I don't think that any extension of the NSDirectoryEnumerator is required, nor is that desirable. I also think that more of the existing code should be merged with this new code. It would probably be best if Richard could comment on this code as he knows the existing code and how to integrate with that.

gcasa:
I’m not sure why you believe that.   The issue is this.  The enumerator created in enumeratorAtUrl clearly needs to 1. Skip hidden files which the previous implementation was not doing and 2. It needs to report an error to a handler when one is encountered and either continue enumeration or stop.   So we do, in my opinion, need to extend NSDirectoryEnumerator.

PR #63, opened 2019-08-06
NSProgress branch

gcasa:
This branch contains an implementation of the NSProgress class.  This class allows for easy monitoring of progress for a given task and also allows the creation of a tree of dependent progress objects to monitor subtasks.

### DIFF
--- a/Headers/Foundation/NSProgress.h
+++ b/Headers/Foundation/NSProgress.h
@@ -32,16 +32,13 @@
 extern "C" {
 #endif
 
-@class	NSString, NSDictionary, NSArray;
+@class	NSString, NSDictionary, NSArray, NSNumber;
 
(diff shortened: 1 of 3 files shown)


triplef:
NSProgress.h is not included in the Foundation.h umbrella header.

triplef:
We found a couple more issues with this PR:

- KVO support is completely missing. KVO should be supported for most of the properties, as indicated in the [official documentation](https://developer.apple.com/documentation/foundation/nsprogress?language=objc) if you search for "KVO"

- `localizedDescription` and `localizedAdditionalDescription` are missing setters

- the `isCancelled` property should also be accessible via `cancelled` (and `cancelled` should be KVO-compilant), as it’s defined like this on Apple platforms which supports both access via `progress.cancelled` and `progress.isCancelled`:
```objc
@property(readonly, getter=isCancelled) BOOL cancelled;
```

PR #64, opened 2019-08-12
Fixed NSOperation calling removeObserver more than once.

triplef:
NSOperation was removing self as an observer of "isFinished" both in the KVO callback and in dealloc.

While this is a no-op with the current implementation of `-removeObserver:forKeyPath:`, it would cause an error if/when the implementation is changed to throw an error when unregistering non-registered observers (as is the case on Apple platforms).

### DIFF
--- a/Source/NSOperation.m
+++ b/Source/NSOperation.m
@@ -206,7 +206,10 @@ static NSArray	*empty = nil;
     {
       NSOperation	*op;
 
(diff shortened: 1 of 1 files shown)


fredkiefer:
I am going to merge this as the change itself is correct. I was a bit reluctant as the whole mechanism of observing self seems wrong to me

PR #65, opened 2019-08-14
NSProgress fixes branch

gcasa:
Fix issues with implementation of NSProgress


gcasa:
Per Frederick:

KVO support is completely missing. KVO should be supported for most of the properties, as indicated in the official documentation if you search for "KVO"

localizedDescription and localizedAdditionalDescription are missing setters

the isCancelled property should also be accessible via cancelled (and cancelled should be KVO-compilant), as it’s defined like this on Apple platforms which supports both access via progress.cancelled and progress.isCancelled:

gcasa:
Fixing issues mentioned by Frederick in previous merge request.

Source/NSProgress.m
```diff
@@ -284,9 +308,16 @@ - (BOOL) isIndeterminate
...
 }
 
+- (BOOL) indeterminate
+{
+  return internal->_indeterminate;
+}
+
 - (void) setIndeterminate: (BOOL)flag
 {
+  [self willChangeValueForKey: @"indeterminate"];
```
fredkiefer: I think this is unnecessary as the basic KVO mechanism should create this automatically. This is also true for some of the following methods. We should only need to add willChange.. and didChange.. methods to support different names and even for these there was a better way, if I remember correctly.

PR #66, opened 2019-08-14
Use UTF-8 as default encoding on Android.

triplef:
Android uses UTF-8 as default encoding, e.g. for file paths.

This fixes file handling support for paths containing non-ASCII characters on Android, which would previously throw an exception as they couldn't be converted to C strings using ISO Latin 1 encoding.

(Please let me know if there’s a better place than in GSPrivateCStringEncoding() to make this change.)

### DIFF
--- a/Source/Additions/Unicode.m
+++ b/Source/Additions/Unicode.m
@@ -2865,7 +2865,12 @@ GSPrivateCStringEncoding(const char *encoding)
 
   if (enc == GSUndefinedEncoding)
     {
+#ifdef __ANDROID__
+      // Android uses UTF-8 as default encoding (e.g. for file paths)
+      enc = NSUTF8StringEncoding;
+#else
       enc = NSISOLatin1StringEncoding;
+#endif
     }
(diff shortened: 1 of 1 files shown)


triplef:
To add a bit more info: all the existing methods of determining the default encoding fail on Android, including the ICU methods (`ucnv_getStandardName()`) even though ICU is being used.

rfm:
On 14 August 2019 16:50:32 GMT+01:00, Frederik Seiffert <notifications@github.com> wrote:
>To add a bit more info: all the existing methods of determining the
>default encoding fail on Android, including the ICU methods
>(`ucnv_getStandardName()`) even though ICU is being used.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/gnustep/libs-base/pull/66#issuecomment-521304642

:-). In that case your solution sounds great.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #68, opened 2019-09-02
Fixing NSDictionary keysSortedByValueUsingComparator:

plaurent:
NSDictionary keysSortedByValueUsingComparator: appears to have a bug.  It does not actually return the keys sorted by the values.  (Instead it appears to be sorting the keys based on their own value rather than the value each maps to in the dictionary...)

### DIFF
--- a/Source/NSDictionary.m
+++ b/Source/NSDictionary.m
@@ -29,6 +29,8 @@
 #import "common.h"
 #import "Foundation/NSDictionary.h"
 #import "Foundation/NSArray.h"
+#import "Foundation/NSMutableArray.h"
+#import "Foundation/NSOrderedSet.h"
 #import "Foundation/NSData.h"
 #import "Foundation/NSException.h"
 #import "Foundation/NSAutoreleasePool.h"
@@ -958,8 +960,13 @@ compareIt(id o1, id o2, void* context)
 - (NSArray *)keysSortedByValueWithOptions: (NSSortOptions)opts
 			  usingComparator: (NSComparator)cmptr
 {
-  return [[self allKeys] sortedArrayWithOptions: opts
-				usingComparator: cmptr];
+  NSArray* sortedValues = [[self allValues] sortedArrayWithOptions: opts usingComparator: cmptr];
+  NSArray* noDuplicates = [[NSOrderedSet orderedSetWithArray:sortedValues] array];
+  NSMutableArray* result = [[NSMutableArray alloc] initWithCapacity:[sortedValues length]];
+  for (NSObject* value in noDuplicates) {
+      [result addObjectsFromArray:[self allKeysForObject:value]];
+  }
+  return result;
 }
 
 /**
--- a/Source/NSDictionary.m
+++ b/Source/NSDictionary.m
@@ -29,7 +29,6 @@
 #import "common.h"
 #import "Foundation/NSDictionary.h"
 #import "Foundation/NSArray.h"
-#import "Foundation/NSMutableArray.h"
 #import "Foundation/NSOrderedSet.h"
 #import "Foundation/NSData.h"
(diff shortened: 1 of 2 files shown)


plaurent:
base/NSDictionary/sort.m:
2407 Failed test:     sort.m:39 ... Can sort a dictionary's keys by its values
2408

fredkiefer:
This is a great test that proves that the current implementation is wrong. We cannot merge it by itself, as this would lead to a breaking build system. A fix for the implementation is needed first, But that should be trivial as we already have the method keysSortedByValueUsingSelector: close to the failing method and this looks correct to me. (You could add a test for that method as well.)
It should be rather simple to take over the implementation from the correct method.

In the long run we should think about rewriting all the sort operations based on NSSortDescriptor to allow for more complex sorting and have just one implementation that needs to be maintained.

triplef:
@plaurent good catch! @gcasa did the original implementation of this method and should be able to help with the fix if needed.

Tests/base/NSDictionary/sort.m
```diff
@@ -0,0 +1,43 @@
...
+  NSArray* keys = [NSArray arrayWithObjects:
+        @"shouldSortToSecond",
+        @"shouldSortToFirst",
+        @"shouldSortToThird",
+        @"shouldSortToFourth",
+        nil];
+
+  NSDictionary *d = [NSDictionary dictionaryWithObjects:values forKeys:keys];
+  NSArray* keysOrderedByKeyedValue = [d keysSortedByValueUsingComparator:
+                      ^NSComparisonResult(id obj1, id obj2) {
```
fredkiefer: You should protect this test with a HAS_BLOCK (or what ever this is called) guard, as gcc won't be able to compile this.

Source/NSDictionary.m
```diff
@@ -958,8 +959,13 @@ - (NSArray *)keysSortedByValueUsingComparator: (NSComparator)cmptr
 - (NSArray *)keysSortedByValueWithOptions: (NSSortOptions)opts
 			  usingComparator: (NSComparator)cmptr
 {
-  return [[self allKeys] sortedArrayWithOptions: opts
-				usingComparator: cmptr];
+  NSArray* sortedValues = [[self allValues] sortedArrayWithOptions: opts usingComparator: cmptr];
+  NSArray* noDuplicates = [[NSOrderedSet orderedSetWithArray:sortedValues] array];
+  NSMutableArray* result = [[NSMutableArray alloc] initWithCapacity:[sortedValues count]];
+  for (NSObject* value in noDuplicates) {
```
fredkiefer: All current compiler are able to process this modern for loop, still we prefer to use the macros FOR_IN and END_FOR_IN.
Although somebody should check the support of gcc in GSFastEnumeration.h as you did not get a compiler warning for this.

rfm: Fred, I added fixes for the issues you spotted, plus I tidied to match coding style and fixed a memory leak.

fredkiefer:
The only test case that fails to build is this and that is unrelated:

Testing general.m...
6792Building base/NSMutableArray/general.m
6793make debug=yes general
6794Making all for test_tool general...
6795 Compiling file general.m ...
6796general.m: In function ‘main’:
6797general.m:107:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
6798     for (int i = 0; i < 5; i++)
6799     ^
6800general.m:107:5: note: use option -std=c99 or -std=gnu99 to compile your code
6801make[5]: *** [obj/general.obj/general.m.o] Error 1
6802make[4]: *** [internal-tool-all_] Error 2
6803make[3]: *** [general.all.test-tool.variables] Error 2
6804make[2]: *** [general] Error 2
6805Failed build:

PR #70, opened 2019-09-10
Move NSFileWrapper to Foundation

gcasa:
NSFileWrapper was migrated to Foundation from AppKit in macOS 10.7. This change implements that move for GNUstep.

### DIFF
--- a/Source/NSFileWrapper.m
+++ b/Source/NSFileWrapper.m
@@ -1,0 +1,780 @@
+/** <title>NSFileWrapper</title>
+
+   <abstract>Hold a file's contents in dynamic memory.</abstract>
+
+   Copyright (C) 1996 Free Software Foundation, Inc.
+
+   Author: Felipe A. Rodriguez <far@ix.netcom.com>
+   Date: Sept 1998
+   Author: Jonathan Gapen <jagapen@whitewater.chem.wisc.edu>
+   Date: Dec 1999
+   
+   This file is part of the GNUstep GUI Library.
+
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2 of the License, or (at your option) any later version.
+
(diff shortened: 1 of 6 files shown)


gcasa:
I believe I have removed all AppKit/GUI references now.

gcasa:
Oh wait... I see an additional issue.  Fixing unarchiving. :)

Source/NSFileWrapper.m
```diff
@@ -0,0 +1,765 @@
...
+  // FIXME
+  return [self writeToFile: [url path]
+                atomically: options & NSFileWrapperWritingAtomic
+           updateFilenames: options & NSFileWrapperWritingWithNameUpdating];
+}
+
+// Simple foundation implementation to avoid issues when unarchiving.
+- (void) setIcon: (id)icon
+{
+  ASSIGN(_icon, icon);
```
fredkiefer: You renamed this ivar to _iconImage, this also needs to be fixed for the gui pull request. Or even better, use this method there.

gcasa: I did not.  It was like that when I moved it!

gcasa: Changed to be more sensible anyway.  I am still leaving the setIcon method in gui since it may have some specific things which need to happen there.

Source/NSFileWrapper.m
```diff
@@ -0,0 +1,768 @@
...
+{
+  // FIXME
+  return [self writeToFile: [url path]
+                atomically: options & NSFileWrapperWritingAtomic
+           updateFilenames: options & NSFileWrapperWritingWithNameUpdating];
+}
+
+- (void) setIcon: (id)image
+{
+  // this method is here to quell compiler warnings.
```
fredkiefer: Now that you renamed the instance variable again, the old implementation of this method would work again.

fredkiefer:
This current version would work, but I would prefer an implementation that would allow gui to work without knowing the internal instance variable names.

PR #71, opened 2019-09-19
Added type definitions and protocols for NSURLSession.

triplef:
This allows building code that references those types without requiring an implementation of NSURLSession.

### DIFF
--- a/Headers/Foundation/NSURLSession.h
+++ b/Headers/Foundation/NSURLSession.h
@@ -4,6 +4,67 @@
 #import <Foundation/NSObject.h>
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_9,GS_API_LATEST)
+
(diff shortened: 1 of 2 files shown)


gcasa:
@rfm is there any reason not to merge this?  LGTM.

rfm:
Generally we try *not* to implement fake interfaces/stubs because we thought that, where something is not supported, it is better to find out at compile time rather than at run time.
The exception would be where a feature is being actively developed and we expect to have the full implementation in place before we do another release.
While, we have never been completely rigorous about ensuring interfaces are implemented when added, this seems like quite a large addition (a whole class left unimplemented).
I don't have particularly strong feelings on changing the general policy though.

triplef:
Thanks @rfm, that makes sense to me. I’ll close this for now.

PR #72, opened 2019-09-25
Perform non-concurrent block enumeration in place.

triplef:
Was previously using serial queue, which doesn't match Apple platforms.

Also fixes incorrectly hardcoded `enumQueue` and `enumQueueGroup` variables in `GS_DISPATCH_TEARDOWN_QUEUE_AND_GROUP_FOR_ENUMERATION()` macro.


triplef:
Can anyone say if the failed tests (NSURLConnection) could be related to this change, or whether there’s some other cause for them failing?

gcasa:
base/NSURL/test02.m:
Failed test:     test02.m:102 ... NSURLConnection synchronous load returns a response
Failed test:     test02.m:109 ... NSURLConnection capture test OK
--- Running tests in base/NSURLConnection ---
base/NSURLConnection/test03.m:
Failed file:     test03.m aborted without running all tests!
base/NSURLConnection/test05.m:
Failed file:     test05.m aborted without running all tests!
base/NSURLConnection/test07.m:
Failed file:     test07.m aborted without running all tests!

Are you sure your changes didn't effect serial queues as well?

triplef:
https://github.com/gnustep/libs-base/pull/73 has the exact same tests failing, so I’m pretty sure it’s unrelated. Not sure why these tests are failing though.

rfm:
That failure is a different bug, exposed by a recent fix elsewhere.  it should now be fixed.

gcasa:
Any updates on this branch's status?  Why hasn't it been merged?

fredkiefer:
I resolved the conflict on the ChangeLog file and merged master into this branch to see whether there are still open issues with this change.

triplef:
Looks like I’m not quite there yet... I’ll give it another go.

triplef:
Looks like all tests are passing now. The patch turned out to be a bit more extensive as the `GS_DISPATCH_*` macros are also used for enumeration in the various collection classes. All these cases should now be performing the block in place unless the `NSEnumerationConcurrent` option is used.

triplef:
Thanks for the approval @fredkiefer. Let me try to explain the change in a bit more detail.

The change basically updates the behavior of all collection class methods that allow enumerating the collection with a block (i.e. `enumerate*UsingBlock:` methods on NSArray/NSDictionary/NSSet/NSOrderedSet/NSIndexSet and `indexesOfObjectsWithOptions:passingTest:` on NSArray/NSOrderedSet) to call the block _inline_ (i.e. on the calling thread) by default, unless `NSEnumerationConcurrent` is specified as an option (in which case the behavior is unchanged).

Previously the `GS_DISPATCH_*` macros used for the above methods would async-dispatch the given block onto a newly created serial dispatch queue even if `NSEnumerationConcurrent` was _not_ given, which did not match Apple platform behavior and was inherently slower (due to the potential creation of a thread and the required thread dispatch).

PR #74, opened 2019-09-26
NSFile changes branch

gcasa:
NSFileCoordinator, NSFilePresenter, NSPersonNameComponents, NSPersonNameComponentsFormatter, etc


Source/NSByteCountFormatter.m
```diff
@@ -67,6 +67,13 @@ + (NSString *)stringFromByteCount: (long long)byteCount
   return [formatter stringFromByteCount: byteCount];
 }
 
+- (NSString *)stringForObjectValue: (id)obj
+{
+  long long byteCount = [obj longLongValue];
+  return [[self class] stringFromByteCount: byteCount
```
fredkiefer: I don't understand this code here, why don't we just call [self stringFromByteCount: byteCount]?

Source/NSFileCoordinator.m
```diff
@@ -1,7 +1,245 @@
...
+
++ (void) addFilePresenter: (id)presenter
+{
+  [__presenters addObject: presenter];
+  [__presenterMap setObject: presenter forKey: [presenter presentedItemURL]];
+}
+
++ (void) removeFilePresenter: (id)presenter
+{
+  [__presenters removeObject: presenter];
```
fredkiefer: Shouldn't you remove it from __presenterMap as well?

Source/NSFileCoordinator.m
```diff
@@ -1,7 +1,245 @@
...
+- (void) setPurposeIdentifier: (NSString *)ident  // copy
+{
+  ASSIGNCOPY(_purposeIdentifier, ident);
+}
+
+- (void)cancel
+{
+  NSEnumerator *en = [__presenters objectEnumerator];
+  id obj = nil;
+  while((obj = [en nextObject]) != nil)
```
fredkiefer: You could use fast enumeration for all your loops.

Source/NSPersonNameComponentsFormatter.m
```diff
@@ -0,0 +1,236 @@
...
+  return result;
+}
+
+- (NSPersonNameComponents *) personNameComponentsFromString: (NSString *)string
+{
+  NSPersonNameComponents *pnc = AUTORELEASE([[NSPersonNameComponents alloc] init]);
+  NSArray *nameArray = [string componentsSeparatedByString: @" "];
+  NSUInteger count = [nameArray count];
+
+  switch(count)
```
fredkiefer: Shouldn't this method look at the _style variable?
At least it should set the given name in the first case to match the formatting above.

Tests/base/NSPersonNameComponentsFormatter/GNUmakefile
```diff
@@ -0,0 +1,29 @@
+# __GENERATED__ makefile marker
```
fredkiefer: I think this is a generated file, it should not be on git. The same seems to be true about the next file as well.

Source/NSByteCountFormatter.m
gcasa: You're right.  Done.

Source/NSFileCoordinator.m
gcasa: Fixed.

Tests/base/NSPersonNameComponentsFormatter/GNUmakefile
gcasa: Removed

Source/NSPersonNameComponentsFormatter.m
gcasa: I had assumed that it doesn't only because the documentation makes no mention of it.  The logic of the method is fairly simple without it.   Additionally I think style should effect output and not parsing.

Source/NSPersonNameComponentsFormatter.m
```diff
@@ -0,0 +1,236 @@
...
+      _phonetic = NO;
+      _style = NSPersonNameComponentsFormatterStyleDefault;
+      _nameOptions = 0L;
+    }
+  return self;
+}
+
+- (void) _setNameOptions: (NSPersonNameComponentsFormatterOptions)opts
+{
+  _nameOptions = opts;
```
fredkiefer: I think the only option that is currently defined in the one for phonetic output. You could just set that value here and use it further down in stringFromPersonNameComponents:

PR #75, opened 2019-10-02
Fixed mutable dictionary keyed subscript behavior.

triplef:
Assigning nil value via keyed subscript now correctly removes object for key.

This matches the documented reference implementation:
https://developer.apple.com/documentation/foundation/nsmutabledictionary/1574187-setobject?language=objc


rfm:
Thanks ... seems an obviously correct fix.

triplef:
Thanks for the quick merge. 👍

PR #79, opened 2019-11-03
gitignore update according to https://www.gitignore.io/api/xcode

IOOI-SqAR:
libs-base contains base.xcodeproj so this makes sense

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,43 @@ Tests/base/coding/ulong-8.type
 # Editor byproducts
 *.orig
 *.swp
+
+# Created by https://www.gitignore.io/api/xcode
+# Edit at https://www.gitignore.io/?templates=xcode
+
+### Xcode ###
+# Xcode
+#
+# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
+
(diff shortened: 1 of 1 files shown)


gcasa:
This appears to break the build in it's current state, though I can't imaging why... could you update your branch and see if it corrects this?

IOOI-SqAR:
@gcasa are you sure it is this PR that breaks the build? This is just one changed .gitignore file, I wonder how this could possibly break a build.

IOOI-SqAR:
@gcasa have a look here, it is something about NSConstantString: https://travis-ci.org/gnustep/libs-base/jobs/606852523?utm_medium=notification&utm_source=github_status

gcasa:
> @gcasa have a look here, it is something about NSConstantString: https://travis-ci.org/gnustep/libs-base/jobs/606852523?utm_medium=notification&utm_source=github_status

Yes, I see.  It's not the PR at all.  I think I am going to go ahead and merge this if there are no objections.  I simply don't want it hanging out here for no reason and it seems reasonable since it filters out a lot of things we don't want to consider committing.

IOOI-SqAR:
@gcasa thanks for merging this!

PR #83, opened 2019-11-15
Added NSException support for libunwind

triplef:
Adds NSException backtrace support using libunwind for platforms where the other facilities like `backtrace()` are not available (e.g. Android).

While unfortunately the backtrace will not include Objective-C symbols (because `dladdr()` does not support these), it fixes the app crashing in `NSCountFrames()` on Android when an exception is thrown. If anyone has a pointer how to obtain Objective-C symbols please let me know.

Also contains a [small improvement](https://github.com/gnustep/libs-base/pull/83/files#diff-42146764331f4ac76b55f9724cc604f4R1302-R1303) in the symbol frame formatting for platforms without any of the supported backtrace libraries (previously the frame pointer would be printed as `{pointer=0x1234;}` instead of just `0x1234`).

### DIFF
--- a/Source/NSException.m
+++ b/Source/NSException.m
@@ -565,6 +565,35 @@ GSListModules()
 #endif	/* USE_BFD */
 
 
+#if defined(HAVE_UNWIND_H) && !defined(HAVE_BACKTRACE)
+
+#include <unwind.h>
+#include <dlfcn.h>
+
+struct GSBacktraceState
+{
+  void **current;
+  void **end;
+};
+
+static _Unwind_Reason_Code
+GSUnwindCallback(struct _Unwind_Context* context, void* arg)
+{
+    struct GSBacktraceState *state = (struct GSBacktraceState*)arg;
+    uintptr_t pc = _Unwind_GetIP(context);
+    if (pc) {
+      if (state->current == state->end) {
+        return _URC_END_OF_STACK;
+      } else {
+        *state->current++ = (void*)pc;
+      }
+    }
+    return 0; //_URC_OK/_URC_NO_REASON
+}
+
+#endif	/* HAVE_UNWIND_H && !HAVE_BACKTRACE */
+
+
 #if	defined(_WIN32) && !defined(USE_BFD)
 typedef USHORT (WINAPI *CaptureStackBackTraceType)(ULONG,ULONG,PVOID*,PULONG);
 typedef BOOL (WINAPI *SymInitializeType)(HANDLE,char*,BOOL);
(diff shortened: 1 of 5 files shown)


triplef:
Thanks @rfm. I added a small tweak so that the library name is always output even if the symbol name is missing.

Here’s an example backtrace (truncated) to get a better picture:
```
0: 0xd51f87f4 libgnustep-base.so unknown
1: 0xd4ee6ad6 libnative-lib.so unknown
2: 0xd4ee6f1c libnative-lib.so Java_com_example_helloobjectivec_MainActivity_throwObjCException + 124
3: 0xee52ba18 libart.so unknown
4: 0xee525a03 libart.so unknown
5: 0xedfd7f4f libart.so _ZN3art9ArtMethod6InvokeEPNS_6ThreadEPjjPNS_6JValueEPKc + 223
6: 0xee1d0b23 libart.so _ZN3art11interpreter34ArtInterpreterToCompiledCodeBridgeEPNS_6ThreadEPNS_9ArtMethodEPNS_11ShadowFrameEtPNS_6JValueE + 339
...
```

There are two open issues here:

1. Frame 0 is in libgnustep-base, even though the actual exception was thrown in libnative-lib in frame 1 below. Does anyone know if this is the same on other platforms? It seems that `FrameOffset` should ideally be adjusted from 4 to 5 (at least with our configuration), but I was unsure about this change.

2. As you can see C++ symbols are mangled. Nice to have: Does anyone know a way to demangle C++ symbols from C code? Unfortunately cxxabi.h and `__cxxabiv1::__cxa_demangle()` don’t seem to be usable from non-C++ code.

Source/NSException.m
```diff
@@ -846,7 +875,19 @@ NSUInteger NSCountFrames(void)
 GSPrivateReturnAddresses(NSUInteger **returns)
 {
   unsigned      numReturns;
-#if HAVE_BACKTRACE
+#if defined(HAVE_UNWIND_H)
```
fredkiefer: Does this order lead to unwind being used instead of backtrace if both are available? And are there system where we could have both? Which one would be preferable?

triplef: You’re right currently unwind would be used over backtrace if both are available, because unwind seemed like the more modern library but I’m not sure which one is preferable. Does anyone know? Otherwise I can just switch it so we don’t change current behavior.

triplef: Thinking about this some more it does make sense to me to keep backtrace() as the preferred solution as it likely has better results for symbolication. I’ve updated the PR to prefer backtrace over unwind, so the change now only applies to systems without support for backtrace (e.g. Android).

PR #84, opened 2019-11-24
Rework some of the weak handling for NS{Map,Hash}Table.

davidchisnall:
The existing code was assuming that weak object pointers were safe to
read directly, without going via the read barrier, which is incorrect.
It was also attempting to retain the result in some places.  The new
code is hopefully somewhat more correct.

### DIFF
--- a/Source/NSConcreteHashTable.m
+++ b/Source/NSConcreteHashTable.m
@@ -75,7 +75,9 @@ typedef GSIMapNode_t *GSIMapNode;
 #define	GSI_MAP_KTYPES	GSUNION_PTR | GSUNION_OBJ
 #define	GSI_MAP_TABLE_T	NSConcreteHashTable
 #define	GSI_MAP_TABLE_S	instanceSize
-
+  
+#define IS_WEAK(M) \
+      M->cb.pf.options & (NSPointerFunctionsZeroingWeakMemory | NSPointerFunctionsWeakMemory)
 #define GSI_MAP_HASH(M, X)\
  (M->legacy ? M->cb.old.hash(M, X.ptr) \
  : pointerFunctionsHash(&M->cb.pf, X.ptr))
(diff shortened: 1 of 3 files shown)


fredkiefer:
The Travis checks are currently failing because of libobjc2 not building. Could you please have a look into this first? The error message is:
/home/travis/dependency_source/libobjc2/arc.mm:6:10: fatal error: 
'third_party/robin-map/include/tsl/robin_map.h' file not found
#include "third_party/robin-map/include/tsl/robin_map.h"

Looks like there is a new dependency for libobjc2 that needs to be included in the build script.

Source/NSConcreteMapTable.m
```diff
@@ -116,16 +116,20 @@ @interface	NSConcreteMapTable : NSMapTable
 	else\
 	  pointerFunctionsAssign(&M->cb.pf.v, (void**)addr, (x).obj);
 */
+#define IS_WEAK_KEY(M) \
```
fredkiefer: Shouldn't these definitions come before the usages?

Source/NSConcreteHashTable.m
```diff
@@ -84,13 +86,13 @@ @interface	NSConcreteHashTable : NSHashTable
  : pointerFunctionsEqual(&M->cb.pf, X.ptr, Y.ptr))
 #define GSI_MAP_RELEASE_KEY(M, X)\
  (M->legacy ? M->cb.old.release(M, X.ptr) \
- : pointerFunctionsRelinquish(&M->cb.pf, &X.ptr))
+  : IS_WEAK(M) ? nil : pointerFunctionsRelinquish(&M->cb.pf, &X.ptr))
 #define GSI_MAP_RETAIN_KEY(M, X)\
  (M->legacy ? M->cb.old.retain(M, X.ptr) \
- : pointerFunctionsAcquire(&M->cb.pf, &X.ptr, X.ptr))
+  : IS_WEAK(M) ? nil : pointerFunctionsAcquire(&M->cb.pf, &X.ptr, X.ptr))
```
fredkiefer: Could you please provide a short explanation why nil is the correct value here?

davidchisnall: It isn't particularly correct, but the value is unused.  The correct solution is to rewrite GSIMap to use ARC-like semantics.

davidchisnall:
libobjc2 has some git submodules.  These two commands need to be run after the git checkout:
```
    git submodule sync
    git submodule update
```
Azure Pipelines does this automatically, Cirrus doesn't, so the libobjc2 CI script does it manually for the FreeBSD builds.

Source/NSConcreteMapTable.m
davidchisnall: They're macros, so they need only to come before macros that instantiate them are, themselves, instantiated.

fredkiefer: True for the compiler but confusing for me. Our most important goal should be to make the code easy to understand. For this moving the macros would really help.

fredkiefer:
Hi David,
could you please merge the change from master first (maybe rebasing your pull request?). I also would like to have my question from the mailing list answered, that is: Why didn't you implement your changes in NSConcretePointerFunctions?

Source/NSConcreteHashTable.m
davidchisnall: In particular, I copied the use of `nil` here from other macros with the same structure using GSIMap.

fredkiefer:
Richard, could you please comment on this? I don't think the current patch is the best possible solution but it looks more correct than the current code and I would like to merge it. Hoping for something cleaner later on.

triplef:
Maybe it would also make sense to merge #80 into this first to see if the tests I added for this are passing with this change.

fredkiefer:
As Richard is currently travelling on an important mission. I am willing to  merge this one without his consent. This may result in some necessary cleanup in the future.
Any objections?

PR #86, opened 2019-11-24
Set up libobjc2's git submodules for travis CI

plaurent:
Adding setup for libobjc2 submodules, if builds correctly can be merged into #84

### DIFF
--- a/travis-deps.sh
+++ b/travis-deps.sh
@@ -22,6 +22,7 @@ install_ng_runtime() {
     cd $DEP_SRC
     git clone https://github.com/gnustep/libobjc2.git
     cd libobjc2
+    git submodule init
     git submodule sync
     git submodule update
     cd ..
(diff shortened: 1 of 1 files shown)


fredkiefer:
There still is an error when building with clang:

In file included from /usr/lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/initializer_list:36:
/usr/lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/c++0x_warning.h:32:2: error: 
      This file requires compiler and library support for the ISO C++ 2011
      standard. This support is currently experimental, and must be enabled with
      the -std=c++11 or -std=gnu++11 compiler options.

What is missing is just one line in the makefile to  set this flag. It is a bit harder to do it in the Travis build script. Something like
export CPPFLAGS=-std=c++11
should work.

davidchisnall:
I'm very surprised that you need to manually set `-std=c++11`.  Any version of clang new enough to support the modern ABI defaults to at least C++14.  What version is GNUstep's CI using for the ng-gnu-gnu library combo?

I'll modify the CMake to explicitly select C++14 mode, please remove this line from the CI script once it's merged.

PR #89, opened 2019-12-02
Fix CI for runtime ABI 2.0

ngrewe:
Travis isn't currently *really* running Ci for base built with the 2.0 runtime ABI. This PR fixes that. The magic sauce here is to use at least clang 8 and some linker which is not ld.bfd.

I've also fixed one of the blocks related NSArray tests which wasn't accounting for concurrency properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ matrix:
           env: LIBRARY_COMBO=gnu-gnu-gnu
 sudo: required
 before_install:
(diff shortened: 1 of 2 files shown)


fredkiefer:
I would prefer if we could use the clang9 from a Ubuntu repository, as this is what most users will do. But it looks like we will have to wait for the next Ubuntu release for that.

PR #91, opened 2019-12-15
* Headers/GNUstepBase/GSIMap.h: Use GSI_MAP_NODE_IS_EMPTY macros

fredkiefer:
when checking in enumerator.
* Source/NSConcreteHashTable.m,
* Source/NSConcreteMapTable.m: Revert part of the weak pointer change.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2019-12-15  Fred Kiefer <fredkiefer@gmx.de>
+
+	* Headers/GNUstepBase/GSIMap.h: Use GSI_MAP_NODE_IS_EMPTY macros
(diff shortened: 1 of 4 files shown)


fredkiefer:
OK, this didn't work out as expected. Back to the drawing board :-(

fredkiefer:
Anybody got an idea why this is no longer running travis on this pull request? Does it only run at specific intervals?

fredkiefer:
OK, the mixed approach seems to work for weak pointers. Now Sergii needs to test, whether it resolves his issue.
On the other hand this also means that most of the other places in NSConcrete*Table where we directly access the key or value will require changes.

PR #96, opened 2020-01-31
configure.ac: Use PKG_CHECK_MODULES to detect libxml2

hughmcmaster:
Debian has begun removing `xml2-config` from its libxml2-dev package, meaning `configure` fails to find libxml2.

This patch modifies `configure.ac` to use `PKG_CHECK_MODULES` to detect libxml2.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3146,13 +3146,13 @@ AC_ARG_ENABLE(xml,
   [  --disable-xml			Build even if XML-dependencies are not met],,
   enable_xml=yes)
 
-if test $enable_xml = yes; then
-  # Save CFLAGS and LIBS as AM_PATH_XML clobbers these variables regardless
-  # of the success of the macro.
-  saved_LIBS="$LIBS"
-  saved_CFLAGS="$CFLAGS"
+HAVE_LIBXML=0
 
-  AM_PATH_XML(2.3.0, enable_libxml=yes, enable_libxml=no)
+if test $enable_xml = yes; then
+  PKG_CHECK_MODULES([XML], [libxml-2.0 >= 2.3.0], [enable_libxml=yes], [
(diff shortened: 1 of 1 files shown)


rfm:
Unless I'm missing something, this patch replaces the existing mechanism and would therefore break libxml2 detection on all older platforms (and possibly some modern ones).

I'd very much like to see a patch which tries pkgconfig first and then falls back to use the existing detection mechanism if pkgconfig is not available or can't find the library.

hughmcmaster:
> 
> 
> Unless I'm missing something, this patch replaces the existing mechanism and would therefore break libxml2 detection on all older platforms (and possibly some modern ones).

libs-base already attempts to use pkg-config, and ICU is detected solely by pkg-config. Modernising would be ideal. However...

> I'd very much like to see a patch which tries pkgconfig first and then falls back to use the existing detection mechanism if pkgconfig is not available or can't find the library.

Okay, I've updated the patch with this in mind.

Note that I've removed the save and restore CFLAGS/LIBS lines, as I could not find any evidence that AM_PATH_XML clobbers $CFLAGS or $LIBS. Besides, the existing restore code was useless, since configure terminated anyway. And, if $LIBS was clobbered, `LIBS="$XML_LIBS $LIBS` would be broken too.

hughmcmaster:
Hi @rfm, will you have time to look at my patch this week?

rfm:
> On 4 Feb 2020, at 10:22, Hugh McMaster <notifications@github.com> wrote:
> 
> Hi @rfm, will you have time to look at my patch this week?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.

I'm away on a business trip, but should be able to look at it on friday.
Or, Fred might give it a loom and merge it earlier if he has time.

PR #99, opened 2020-02-17
NSPropertyList.h: mention base64 encoding

Artoria2e5:
Base64 parsing is supported since 3f2fa9ddfe507b68fd92843c345c7994b1c1adc8 and it is the default since 9aa5d4cd0403c2d990262c15884181da5d1e32ae. We should definitely document that.

### DIFF
--- a/Headers/Foundation/NSPropertyList.h
+++ b/Headers/Foundation/NSPropertyList.h
@@ -164,13 +164,13 @@ typedef NSUInteger NSPropertyListFormat;
  *       }
  *     </example>
  *     In XML format, a dictionary is an element whose name is
- *     <code>dictionary</code> and whose content consists of pairs of
- *     strings and other <em>property list</em> objects.
+ *     <code>dict</code> and whose content consists of pairs of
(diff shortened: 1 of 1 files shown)


Artoria2e5:
How the heck does a documentation change cause segfaults in CI

Artoria2e5:
And the dict XML documentation needs updating too. Gah.

Artoria2e5:
I am not familiar with the gsautodoc thing, so I am leaving the author and date fields unchanged for now. It might be a good idea to bump the date up by a dozen years to make people feel more confident in it.

Artoria2e5:
About the CI... It's been broken for `LIBRARY_COMBO=ng-gnu-gnu` and `LIBRARY_COMBO=ng-gnu-gnu BASE_ABI=--disable-mixedabi` since c8cfcec652cba1402cfb61aea9668a8805d1a59a ([build 1095](https://travis-ci.org/gnustep/libs-base/builds/649350515)). Looks like too minor a change to break stuff though.

Issue #101, opened 2020-02-18
Add plutil(1)


Artoria2e5:
Since its introduction in Mac OS X 10.2, many users of the said system have started to seen plutil(1) as the only plist command-line editor. plutil(1) is a program that does most of pldes(1) and friend does, but all put in one tool. Around macOS 10.15, plutil(1) seems to have received a lot of editing capabilities that are not found in online manpages.

I am currently working on a version of plutil for GNUstep, sans the programming language and JSON literal features for now. This issue is really just for a future pull request reference.

plutil(1) manpage in macOS 10.14: https://www.unix.com/man-page/mojave/1/plutil
plutil(1) manpage in macOS 10.15: [plutil.1.gz](https://github.com/gnustep/libs-base/files/4218369/plutil.1.gz)

* * *
A related utility found on macOS but not yet in GS is PlistBuddy. Both utilities have a MIT/Expat licensed C++ implementation by Facebook at https://github.com/facebook/xcbuild/tree/master/Libraries/plist/Tools. I am mainly writing plutil(1) from scratch for the new 10.15 features.

rfm:
Contribution now merged.

PR #102, opened 2020-02-18
Add plutil utility

Artoria2e5:
This is supposedly an expansion of the macOS plutil(1). See #101.

- [x] Main utility
- [x] extra formats:
  - [ ] write `objc` (modern literal format)
  - [ ] write `swift`
  - [x] read/write NSJSONSerialization 
- [ ] tests (shell script)
- [ ] manpage
- [ ] gsautodoc


Source/NSPropertyList.m
```diff
@@ -345,7 +345,7 @@ - (void) parser: (NSXMLParser *)parser
     }
   else if ([elementName isEqualToString: @"real"])
     {
-      ASSIGN(plist, [NSNumber numberWithDouble: [value doubleValue]]);
+      ASSIGN(plist, [NSNumber numberWithDouble: strtod([value cString], 0)]);
```
ngrewe: Cool! A minor (teensy, in fact) nitpick, though: The type of the second parameter to `strtod` is also a pointer, so `NULL` would be more appropriate than `0`.

Artoria2e5:
```suggestion
      ASSIGN(plist, [NSNumber numberWithDouble: strtod([value cString], NULL)]);
```

rfm:
I note that build is failing because at plutil.m lines 107 and 109 there is a comparison between a pointer to a pointer and nil.  This comparison should use 0 rather than nil.

probonopd:
Is this still being worked on? I am also running into the same compilation issue.

Artoria2e5:
Yes! Sorry for all the delays…

Artoria2e5:
Okay, it compiles now. I am going to leave the cool new language formats unimplemented, and work on the doc and the tests instead.

probonopd:
Thank you very much. This is a very useful tool. I hope it gets added at some point.

rfm:
Thanks very much ... merged.

PR #105, opened 2020-02-23
fix CI build for gnustep ABI < 2.0

ngrewe:
The CI jobs for the older GNUstep Objective-C ABI were failing because newer gnustep-make versions were implicitly upgrading the build to the gnustep-2.0 ABI. This PR fixes that by stating explicitly which runtime ABI should be used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,16 @@ compiler:
     - gcc
 env:
     - LIBRARY_COMBO=gnu-gnu-gnu
-    - LIBRARY_COMBO=ng-gnu-gnu
-    - LIBRARY_COMBO=ng-gnu-gnu BASE_ABI=--disable-mixedabi
+    - LIBRARY_COMBO=ng-gnu-gnu RUNTIME_VERSION=gnustep-1.9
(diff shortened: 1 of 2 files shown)


ngrewe:
> Looks great to me. I will be copying this over to gui unless you beat me on that :-)

Copy away! I'm still trying to come up with a more DRY way to do share the CI dependencies between the different GNUstep repositories…

Letterus:
Does setting RUNTIME_VERSION=gnustep-1.9 work for you to select the older runtime? This failed to me two days ago. I had to checkout 1.9 branch. Can you confirm this?

ngrewe:
> Does setting RUNTIME_VERSION=gnustep-1.9 work for you to select the older runtime? This failed to me two days ago. I had to checkout 1.9 tag. Can you confirm this?

Not quite. If you want to use version 1.9 of the runtime library, you still  need to check out the v1.9 tag. Setting `RUNTIME_VERSION=gnustep-1.9` tells the compiler to emit code that is ABI  compatible with v1.9 of the runtime (which is still supported  by v2.0).

Letterus:
Thank you for your explanation. I thought 2.0 was not backwards compatible.

PR #107, opened 2020-03-01
Extend NSBundle test to check for framework resource loading.

fredkiefer:
I was trying to write a test that would allow to automatically test the clang libobjc2 issue we are currently fixing on make. As I do not use this environment myself this is just a blind guess. 

For this I reused a test by Niels and rewrote that to load the framework differently, hoping that this will trigger the issue.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2020-03-01  Fred Kiefer <fredkiefer@gmx.de>
+
+	* Tests/base/NSBundle/TestInfo,
+	* Tests/base/NSBundle/GNUmakefile.preamble,
+	* Tests/base/NSBundle/resources2.m: Extend test to check for
+	framework resource loading.
+
 2020-02-24  Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Source/objc-load.m: update GSPrivateSymbolPath() so that, on the
--- a/Tests/base/NSBundle/GNUmakefile.preamble
+++ b/Tests/base/NSBundle/GNUmakefile.preamble
@@ -1,7 +1,7 @@
 
 ADDITIONAL_INCLUDE_DIRS += -I../GenericTests/ -I../../..
 ADDITIONAL_OBJCFLAGS += -Wall
-
+resources2_LDFLAGS += -Wl,-rpath -Wl,$(CURDIR)/Resources/TestFramework.framework/Versions/Current/$(GNUSTEP_TARGET_LDIR)
 resources2_LIB_DIRS += -L./Resources/TestFramework.framework/$(GNUSTEP_TARGET_LDIR)
 resources2_TOOL_LIBS += -lTestFramework
 
--- a/Tests/base/NSBundle/GNUmakefile.preamble
+++ b/Tests/base/NSBundle/GNUmakefile.preamble
@@ -2,6 +2,9 @@
 ADDITIONAL_INCLUDE_DIRS += -I../GenericTests/ -I../../..
 ADDITIONAL_OBJCFLAGS += -Wall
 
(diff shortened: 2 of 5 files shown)


fredkiefer:
OK, this is rather strange, the test fails in more cases than I expected. As it works locally with gcc for me, there must be another difference in the setup.

ngrewe:
> As it works locally with gcc for me, there must be another difference in the setup.

Travis CI is running the tests Ubuntu Xenial, which is a bit dated, so I wonder whether that might be relevant?

I hope to be able to dig into this in detail sometime this week. One thing I believe would definitely be worth adding is a check for whether `+bundleForClass:` produces the correct result.

fredkiefer:
This patch actually uses ` +bundleForClass:` what needs to change is the version number for the framework. Checking zero is always working as this is the default.

ngrewe:
See, that’s why I need to have more than a cursory look at this 😊

rfm:
Using my locally patched version of gnustep-make,  these testcases are working for me whether I build with gcc (1.8 runtime), clang-3 (1.9 runtime) or clang-9 (2.0 runtime).
I guess that's good.

fredkiefer:
@rfm that is great to hear. Still I would like to understand why the gcc Travis build is failing. I know that gcc with the corresponding libobjc works.

fredkiefer:
I changed the framework version to 2. But the result is different from what I expected. The framework version is not visible to the outside. And even the pathsForResourcesOfType: methods don't use the version directories. This is strange as the pathForResource: methods use these directories. Does anybody have a deeper insight on this?

rfm:
I have never used frameworks, and while I know other people (Niels for instance) have done, it seems possible that they are sufficiently rarely using the the GNUstep community that some features of frameworks have never worked.

ngrewe:
I've figured out the framework test failure for CI with the gcc ABI: We're still using a pretty ancient CI runner (Ubuntu 16.04), where the linker seems to discard the linkage against the TestFramework because we're not using any symbols from the library. Once I fixed that, however, I still had to add the framework library directory to the linker search path… which puzzles me a bit because I would have thought that that would crop up anyways.

ngrewe:
> Once I fixed that, however, I still had to add the framework library directory to the linker search path… which puzzles me a bit because I would have thought that that would crop up anyways.

Now looking over the PR -- that seems to imply that the test harness does not interpret the TestInfo stuff as it should?

ngrewe:
Unfortunately I don't have any deeper insights into the `-bundleVersion` thing either. It's well possible that it's never worked. The API is a GNUstep extension, so there is no particular contract that we have to support with this.

fredkiefer:
> > Once I fixed that, however, I still had to add the framework library directory to the linker search path… which puzzles me a bit because I would have thought that that would crop up anyways.
> 
> Now looking over the PR -- that seems to imply that the test harness does not interpret the TestInfo stuff as it should? (at least not in all cases)

Or the path I used in TestInfo was wrong. Either we correct it or we rather should remove it again.

Will the LD path you had to put in work for all the linkers we currently support? I did not follow the clang/libobjc2 linke thread too closely.

ngrewe:
> Or the path I used in TestInfo was wrong. Either we correct it or we rather should remove it again.

The path looks correct to me, but using -rpath has the nice side effect that you can also run the test binary without the `gnustep-tests` script, so I‘d be in favour of removing the TestInfo setting.

> Will the LD path you had to put in work for all the linkers we currently support?

I think we mostly care about bfd-ld and gold. I‘ve tested those explicitly. People keep mentioning having various issue with lld, so I‘m currently still ignoring it. But since lld claims to be a drop-in replacement for GNU linkers, I suppose -rpath should work with it as well.

PR #108, opened 2020-03-04
Updated NSURL to use instancetype.

triplef:
Enables using the dot syntax with NSURL e.g. `[NSURL URLWithString:@"/"].standardizedURL`.

### DIFF
--- a/Headers/Foundation/NSURL.h
+++ b/Headers/Foundation/NSURL.h
@@ -85,16 +85,16 @@ enum
  * Calls -initFileURLWithPath: which escapes characters in the
  * path where necessary.
  */
-+ (id) fileURLWithPath: (NSString*)aPath;
++ (instancetype) fileURLWithPath: (NSString*)aPath;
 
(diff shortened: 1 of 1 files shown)


Headers/Foundation/NSURL.h
```diff
@@ -85,29 +85,29 @@ enum
  * Calls -initFileURLWithPath: which escapes characters in the
  * path where necessary.
  */
-+ (id) fileURLWithPath: (NSString*)aPath;
++ (instancetype) fileURLWithPath: (NSString*)aPath;
```
fredkiefer: Interestingly Apples uses (NSURL *) for all the factory methods.

triplef: I noticed that too, but it seemed more like a case of API not having been updated.

That being said, please let me know if you prefer matching Apple’s headers.

rfm: I guess Apple uses NSURL rather than instancetype because there are no subclasses (and perhaps they don't intend NSURL to be subclassed), so every returned instance must be an NSURL.  I have no strong feelings about which is better (both are better than id).

PR #109, opened 2020-03-07
* Source/NSString.m: Attempt to make calls to GSICUCollatorOpen()

fredkiefer:
more consistent and correct for ICU 65.

The old code resulted in numerical comparisons failing with ICU65, because the code was just ignoring the system locale. I moved the tow different preprocessing instructions into the  function GSICUCollatorOpen and removed the obviously wrong code path.
The code is at least better than the previous version, I am not sure whether it is already correct for all cases.


rfm:
While I'm no expert on the localisation stuff, I couldn't see any problems.

PR #111, opened 2020-03-10
Fixed attributed string range initialization.

triplef:
Effective range was initialized incorrectly.

Caveat: I’m not entirely sure if this could have been an actual issue, or if this changes the behavior of these functions (though I don’t think so).


triplef:
Makes sense, that’s what I thought. However I’d still argue for using a correct NSRange just to ensure the invalid value is not perpetuated somehow. I stumbled across this by searching for "NSMakeRange" in order to use it in my other patch (I know... 🙈), and almost didn’t catch the mistake.

PR #115, opened 2020-03-30
Fix nsurlcomponents

gnustep-admin:
This change is to make sure that NSURLComponents update it's NSURL when the components are modified.

### DIFF
--- a/Source/NSURL.m
+++ b/Source/NSURL.m
@@ -2389,27 +2389,27 @@ GS_PRIVATE_INTERNAL(NSURLComponents)
   ASSIGNCOPY(internal->_user, [url user]);
 
   // Percent encoded portions...
-  [self setPercentEncodedFragment:
-          [[url fragment] stringByAddingPercentEncodingWithAllowedCharacters:
-                            [NSCharacterSet URLFragmentAllowedCharacterSet]]];
-  [self setPercentEncodedHost:
-          [[url host] stringByAddingPercentEncodingWithAllowedCharacters:
-                        [NSCharacterSet URLHostAllowedCharacterSet]]];
-  [self setPercentEncodedPassword:
-          [[url password] stringByAddingPercentEncodingWithAllowedCharacters:
-                            [NSCharacterSet URLPasswordAllowedCharacterSet]]];;
-  [self setPercentEncodedPath:
-          [[url path] stringByAddingPercentEncodingWithAllowedCharacters:
-                        [NSCharacterSet URLPathAllowedCharacterSet]]];
-  [self setPercentEncodedQuery:
-          [[url query] stringByAddingPercentEncodingWithAllowedCharacters:
-                         [NSCharacterSet URLQueryAllowedCharacterSet]]];
-  [self setPercentEncodedScheme:
-          [[url scheme] stringByAddingPercentEncodingWithAllowedCharacters:
-                           [NSCharacterSet URLPathAllowedCharacterSet]]];
-  [self setPercentEncodedUser:
-          [[url user] stringByAddingPercentEncodingWithAllowedCharacters:
-                         [NSCharacterSet URLUserAllowedCharacterSet]]];
+  ASSIGNCOPY(internal->_percentEncodedFragment,
+             [[url fragment] stringByAddingPercentEncodingWithAllowedCharacters:
+                               [NSCharacterSet URLFragmentAllowedCharacterSet]]);
+  ASSIGNCOPY(internal->_percentEncodedHost,
+             [[url host] stringByAddingPercentEncodingWithAllowedCharacters:
+                           [NSCharacterSet URLHostAllowedCharacterSet]]);
+  ASSIGNCOPY(internal->_percentEncodedPassword,
+             [[url password] stringByAddingPercentEncodingWithAllowedCharacters:
+                               [NSCharacterSet URLPasswordAllowedCharacterSet]]);;
+  ASSIGNCOPY(internal->_percentEncodedPath,
+             [[url path] stringByAddingPercentEncodingWithAllowedCharacters:
+                           [NSCharacterSet URLPathAllowedCharacterSet]]);
+  ASSIGNCOPY(internal->_percentEncodedQuery,
+             [[url query] stringByAddingPercentEncodingWithAllowedCharacters:
+                            [NSCharacterSet URLQueryAllowedCharacterSet]]);
+  ASSIGNCOPY(internal->_percentEncodedScheme,
+             [[url scheme] stringByAddingPercentEncodingWithAllowedCharacters:
+                             [NSCharacterSet URLPathAllowedCharacterSet]]);
+  ASSIGNCOPY(internal->_percentEncodedUser,
+             [[url user] stringByAddingPercentEncodingWithAllowedCharacters:
+                           [NSCharacterSet URLUserAllowedCharacterSet]]);
 
   {
     // Find ranges
--- a/Source/NSURL.m
+++ b/Source/NSURL.m
@@ -2629,23 +2629,39 @@ GS_PRIVATE_INTERNAL(NSURLComponents)
 
 - (NSString *) query
 {
-  NSString *query = @"";
-  NSEnumerator *en = [internal->_queryItems objectEnumerator];
-  NSURLQueryItem *item = nil;
+  NSString	*result = nil;
 
-  while ((item = (NSURLQueryItem *)[en nextObject]) != nil)
+  if (internal->_queryItems != nil)
     {
-      NSString *name = [item name];
-      NSString *value = [item value];
-      NSString *itemString = [NSString stringWithFormat: @"%@=%@",name,value];
+      NSMutableString	*query = nil;
+      NSURLQueryItem	*item = nil;
+      NSEnumerator	*en;
 
-      if ([query length] > 0)
-        {
-          query = [query stringByAppendingString: @"&"];
-        }
-      query = [query stringByAppendingString: itemString];
+      en = [internal->_queryItems objectEnumerator];
+      while ((item = (NSURLQueryItem *)[en nextObject]) != nil)
+	{
+	  NSString	*name = [item name];
+	  NSString	*value = [item value];
+
+	  if (nil == query)
+	    {
+	      query = [[NSMutableString alloc] initWithCapacity: 1000];
+	    }
+	  else
+	    {
+	      [query appendString: @"&"];
+	    }
+	  [query appendString: name];
+	  if (value != nil)
+	    {
+	      [query appendString: @"="];
+	      [query appendString: value];
+	    }
+	}
+      result = AUTORELEASE([query copy]);
+      RELEASE(query);
     }
-  return query;
+  return result;
 }
 
 - (void) _setQuery: (NSString *)query fromPercentEncodedString: (BOOL)encoded
@@ -2848,7 +2864,7 @@ GS_PRIVATE_INTERNAL(NSURLComponents)
 	  name = [name _stringByAddingPercentEscapesForQuery];
 	  value = [value _stringByAddingPercentEscapesForQuery];
 	  ni = [NSURLQueryItem queryItemWithName: name
-					   value: name];
+					   value: value];
 	  [items addObject: ni];
 	}
       result = AUTORELEASE([items copy]);
--- a/Source/NSURL.m
+++ b/Source/NSURL.m
@@ -69,8 +69,7 @@ function may be incorrect
   NSRange   _rangeOfQuery; \
   NSRange   _rangeOfQueryItems; \
   NSRange   _rangeOfScheme; \
-  NSRange   _rangeOfUser; \
-  BOOL      _regenerating;
+  NSRange   _rangeOfUser; 
 
 #import "common.h"
 #define	EXPOSE_NSURL_IVARS	1
--- a/Source/NSURL.m
+++ b/Source/NSURL.m
@@ -2447,10 +2447,17 @@ GS_PRIVATE_INTERNAL(NSURLComponents)
       urlString = [urlString stringByAppendingFormat: @"%@://", internal->_scheme];
     }
 
-  if (internal->_user != nil && internal->_password)
+  if (internal->_user != nil) 
     {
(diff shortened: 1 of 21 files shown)


Source/NSURL.m
```diff
@@ -69,7 +72,7 @@
   NSRange   _rangeOfQuery; \
   NSRange   _rangeOfQueryItems; \
   NSRange   _rangeOfScheme; \
-  NSRange   _rangeOfUser;
+  NSRange   _rangeOfUser;
```
fredkiefer: Trailing whitespace

gnustep-admin: Removed.

gcasa:
> I think the concept of the current implementation is wrong. The class NSURLComponents is there to construct an URL, this means most of the time instances of this class will be incomplete. For that reason the method _regenerateURL should only be called when the URL gets explicitly requested.
> Also storing all components twice, once encode and once decoded seems wrong, we will get inconsistencies that way. We should only store the decoded version and produce and consume the encoded one on the fly.
> And the whole rangeOf handling is very optimistic. This expects that each string occurs only once in the URL. We should rather construct these values while building up the URL string.
> Finally the current code isn't where consistent where encoded strings get concatenated and where the decode version gets used.

I believe I have addressed most of these concerns.   The rangeOf issue you mention is a difficult one since it seems as though the components can be added in any order.

fredkiefer:
I think the pattern that Richard suggested would be a lot better. Keep only the unencoded values (for query the query items) and when setting a variable flag the object as invalid. When requesting an encoded value, produce it on the fly- When the url or a range gets requested and the object is invalid regenerate all these values and flag as valid.
The current code still has a lot of inconsistencies that would mostly vanish if you adopt that pattern. Fixing these one by one is much more hassle.

gcasa:
> I think the pattern that Richard suggested would be a lot better. Keep only the unencoded values (for query the query items) and when setting a variable flag the object as invalid. When requesting an encoded value, produce it on the fly- When the url or a range gets requested and the object is invalid regenerate all these values and flag as valid.
> The current code still has a lot of inconsistencies that would mostly vanish if you adopt that pattern. Fixing these one by one is much more hassle.

I have done this and committed.

Source/NSURL.m
```diff
@@ -92,15 +85,12 @@
...
 
-#define	GSInternal NSURLComponentsInternal
-#include "GSInternal.h"
-GS_PRIVATE_INTERNAL(NSURLComponents)
-
 NSString * const NSURLErrorDomain = @"NSURLErrorDomain";
 NSString * const NSErrorFailingURLStringKey = @"NSErrorFailingURLStringKey";
 
 @interface	NSString (NSURLPrivate)
 - (NSString*) _stringByAddingPercentEscapes;
```
fredkiefer, line 92: We should replace these two methods by proper NSCharactersets for the encoding and call stringByAddingPercentEncodingWithAllowedCharacters: with these.
@rfm could you help with setting up these?

Source/NSURL.m
```diff
@@ -2488,128 +2609,180 @@ - (NSString *) user
...
 - (NSArray *) percentEncodedQueryItems
 {
-  return internal->_percentEncodedQueryItems;
+  NSMutableArray *items = [NSMutableArray arrayWithCapacity: [internal->_queryItems count]];
+  NSEnumerator *en = [internal->_queryItems objectEnumerator];
+  NSURLQueryItem *i = nil;
+
+  while((i = [en nextObject]) != nil)
+    {
+      NSURLQueryItem *ni = [NSURLQueryItem queryItemWithName: [i name]
```
fredkiefer, line 2694: I would expect that the name component needs escaping as well.

gcasa: Do you want to make these changes as I don't want to interfere with any changes you're going to make?

Source/NSURL.m
gcasa: Agreed

Source/NSURL.m
gcasa: I went ahead and made this simple change.  Once it's committed, I leave the branch to you.

fredkiefer: I am fine with either way. I just don't want to waste any more of my time on that PR just to see that another commit has made it impossible to push my changes.

gcasa:
I am going to move some of the tests I have written to the formal test framework.  This should not interfere with other changes made by fred on this branch.  My tests are derived from test code written against macOS.

rfm:
> On 4 Apr 2020, at 11:55, Fred Kiefer <notifications@github.com> wrote:
> 
> 
> @fredkiefer commented on this pull request.
> 
> In Source/NSURL.m:
> 
> > @@ -92,15 +85,12 @@
>  
>  #import "GNUstepBase/NSURL+GNUstepBase.h"
>  
> -#define	GSInternal NSURLComponentsInternal
> -#include "GSInternal.h"
> -GS_PRIVATE_INTERNAL(NSURLComponents)
> -
>  NSString * const NSURLErrorDomain = @"NSURLErrorDomain";
>  NSString * const NSErrorFailingURLStringKey = @"NSErrorFailingURLStringKey";
>  
>  @interface	NSString (NSURLPrivate)
>  - (NSString*) _stringByAddingPercentEscapes;
> 
> We should replace these two methods by proper NSCharactersets for the encoding and call stringByAddingPercentEncodingWithAllowedCharacters: with these.
> @rfm could you help with setting up these?

I can see where you are coming from ... it's elegant to use a single method to do everything.

As an old fashioned system developer the sheer inefficiency of it bugs me though.... new NSCharacterSet  objects use up a fairly large amount of memory and the code to check membership has the overhead of an objc method invocation.  I do know that neither of these things is likely to be relevant, but a mutable string method to append a percent encoded string efficiently would be nice.

Anyway, for now using -stringByAddingPercentEncodingWithAllowedCharacters: makes sense, and we can easily create mutable charactersets using +characterSetWithCharactersInString: and then -invert (depending on whether we want to start with a string of alloed characters or a sring of disallowed characters).
I think the problem is reproducing the OSX behavior for percent escaping;  at work I ignore the OSX/GNUstep methods and implement my own percent escaping, applying an approach where everything that's not definitely legal and non-problematic (a large grey-area) gets escaped.

If Apple stick to the idea that everything outside the ascii range is to be escaped (my reading of the RFCs is that all non-ascii characters are considered unsfe and should always be percent escaped), then we could implement an NSCharacterSet subclass to store all the information we need in a 16-byte (128 bits) bitmap.

rfm:
> On 4 Apr 2020, at 12:07, Gregory Casamento <notifications@github.com> wrote:
> 
> 
> I am going to move some of the tests I have written to the formal test framework. This should not interfere with other changes made by fred on this branch. My tests are derived from test code written against macOS.

Yes please.

fredkiefer:
> I am going to move some of the tests I have written to the formal test framework. This should not interfere with other changes made by fred on this branch. My tests are derived from test code written against macOS.

That sounds like an excellent idea. Please at a lot of cases where encoding is needed. And also file URLs as I am unsure whether the code handles these correctly. Most of all we should have tests where NSURL and NSURLComponents would behave differently. I am almost sure we did not get these correct.

fredkiefer:
I am through with my changes. Looking at the class again I think that we should rather use the absolute string as the base representation to be regenerated and build the url based on that. That way we would be a bit more independent of NSURL.

gcasa:
Just FYI, these two methods can be removed, they are not in the Cocoa implementation...


- (NSString *) percentEncodedScheme
{
  return [internal->_scheme stringByAddingPercentEncodingWithAllowedCharacters:
                    [NSCharacterSet URLPathAllowedCharacterSet]];
}

- (void) setPercentEncodedScheme: (NSString *)scheme
{
  [self setScheme: [scheme stringByRemovingPercentEncoding]];
}

gcasa:
I am going to take out the extraneous methods and add the tests...

gcasa:
The tests revealed a minor issue.  I am going to correct this then we should be good to go.

rfm:
I am aware of no remaining bugs (only a few formatting errors and some cases where we could be doing things a bit more efficiently and/or slightly more clearly).  I think we really want a load of testcases to ensure that what the code is doing matches what OSX does though, particularly, as noted by Fred, in the case of file URLs and perhaps other schemes (it's easy to forget that URLs are not just for HTTP/HTTPS).

gcasa:
There is one remaining issue.  The parsed absolutePath of NSURL always seems to have a trailing / even when there is no path.

gcasa:
I am going to commit the changes I've made to make the tests pass.  Please comment and let me know if you agree or disagree with them.

rfm:
> On 4 Apr 2020, at 16:51, Gregory Casamento <notifications@github.com> wrote:
> 
> 
> There is one remaining issue. The parsed absolutePath of NSURL always seems to have a trailing / even when there is no path.

I recall that there were difficulties getting NSURL to work properly with the case of an empty path because there is no real difference between a URL whose path is empty and one whose path is '/' since a '/' is used both as the delimiter between the host/port and the path and as the root of the path.

gcasa:
I changed it so that it doesn't parse from the string since we have all of the components at that point.  Also it appears there is an issue in the NSURL string parsing as it returns nil for a valid URL.

Source/NSURL.m
```diff
@@ -2306,10 +2405,140 @@ - (id) copyWithZone: (NSZone *)zone
...
+      component = [self percentEncodedFragment];
+      len = [component length];
+      [urlString appendString: @"#"];
+      location += 1;
+      [urlString appendString: component];
+      internal->_rangeOfFragment = NSMakeRange(location, len);
+      location += len;
+    }
+    
+  u = [[NSURL alloc] initWithScheme: [self scheme]
```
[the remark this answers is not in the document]
gcasa: There APPEARS to be an issue parsing the URL for whatever reason in NSURL.  This is not a problem on Cocoa.  I am not sure why the URLFromString method is not working.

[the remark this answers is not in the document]
gcasa: Also, agreed WRT the escaped vs normal strings.

gcasa:
> Having at least a few tests is great. But these are just minimal tests. None of the wired cases gets tested.
> What about a query item where the name contains "&". Or checks that test the differences of NSURL and NSURLComponents.

Agreed.   This is just a start.  Of course these will be expanded.

gcasa:
Unless you feel strongly one way or another, I believe we can go ahead and merge this.   Please feel free to make any additional changes you see fit.

Tests/base/NSURL/components.m
```diff
@@ -0,0 +1,57 @@
...
+  
+  [components setQueryItems: [NSArray arrayWithObjects:
+                                        [NSURLQueryItem queryItemWithName:@"lang" value:@"en"],
+                                      [NSURLQueryItem queryItemWithName:@"response_type" value:@"code"],
+                                      [NSURLQueryItem queryItemWithName:@"uri" value:[[NSURL URLWithString:@"https://some.url.com/path?param1=one&param2=two"] absoluteString]], nil]];
+  // URL
+  PASS([[components string] isEqualToString:
+                                    @"https://user:password@some.host.com/?lang=en&response_type=code&uri=https://some.url.com/path?param1%3Done%26param2%3Dtwo"],
+       "URL string is correct");
+  // NOTE: The slash behind the .com is strictly legal, but this is NOT what is produced by macOS.  I would like to correct this in a
```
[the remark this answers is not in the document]
gcasa: Correct, but according to W3C this is, strictly, valid.   Even if it's not what we might want.   I was on the fence of noting it this way, or leaving the test entirely out.  It seemed better to have a test in place.

rfm:
> On 4 Apr 2020, at 18:01, Fred Kiefer <notifications@github.com> wrote:
> 
> In Source/NSURL.m:
> 
> > +	  if (nil == query)
> +	    {
> +	      query = [[NSMutableString alloc] initWithCapacity: 1000];
> +	    }
> +	  else
> +	    {
> +	      [query appendString: @"&"];
> +	    }
> +	  [query appendString: name];
> +	  if (value != nil)
> +	    {
> +	      [query appendString: @"="];
> +	      [query appendString: value];
> +	    }
> +	}
> +      result = AUTORELEASE([query copy]);
> 
> Why don't we just return the autorelease NSMutableString?

1. The method declares the returned value to be an NSString, not an NSMutableString ... returning the immutable variant, when the API says that's what we do, discourages people from writing bad code using the method and assuming they will get a mutable value returned.
2. You can be space and time efficient by using a mutable string (with a big capacity so appending to it is really quick) to build a value, and then returning an immutable copy (which will use the minimum necessary space).

fredkiefer:
I ist committed the change I suggested. This brings the tests on the same values as Cocoa.
On the other hand I made a few more tests on Cocoa and I am a bit confused about the results. When I set a path without a "/" prefix the created string is nil and the rangeofPath: method results in a segmentation fault. Not sure whether we want to reproduce this behaviour.

gcasa:
> I ist committed the change I suggested. This brings the tests on the same values as Cocoa.
> On the other hand I made a few more tests on Cocoa and I am a bit confused about the results. When I set a path without a "/" prefix the created string is nil and the rangeofPath: method results in a segmentation fault. Not sure whether we want to reproduce this behaviour.

I get this with the current changes....
--- Running tests in Tests/base/NSURL ---

Tests/base/NSURL/components.m:
Failed test:       components.m:21 ... URL string is correct

fredkiefer:
> I get this with the current changes....
> --- Running tests in Tests/base/NSURL ---
> 
> Tests/base/NSURL/components.m:
> Failed test: components.m:21 ... URL string is correct

That is a bit surprising, especially as the travis tests are all passing. Could you please retest your source code and maybe do a "make install". For me this test also required an "install".

gcasa:
> > I get this with the current changes....
> > --- Running tests in Tests/base/NSURL ---
> > Tests/base/NSURL/components.m:
> > Failed test: components.m:21 ... URL string is correct
> 
> That is a bit surprising, especially as the travis tests are all passing. Could you please retest your source code and maybe do a "make install". For me this test also required an "install".

Okay, will do...  WORKED!   I thought the tests are supposed to work even when it's not installed.   I think we are good to go on the merge.

PR #116, opened 2020-03-30
Fixed NSOperation completion block memory management.

triplef:
The NSOperation completion block was not being copied/retained or released.

I’m not totally sure if it’s better to use the Block_copy()/Block_release() functions here, or if I should/can use e.g. the ASSIGNCOPY()/RELEASE() macros instead. GSNotificationObserver uses to former, which is why I went with them.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2020-04-07  Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSOperation.m: Fix completion block memory management.
+
 2020-04-05  Ivan Vucica <ivan@vucica.net>
 
 	* ANNOUNCE:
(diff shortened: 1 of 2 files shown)


triplef:
How can I trigger a CI build for this branch? This used to happen automatically, but didn’t happen here maybe due to being part of the GNUstep organization?

ngrewe:
I can see the CI job over at Travis: https://travis-ci.org/github/gnustep/libs-base/builds/668772398

triplef:
Ah thanks, all green. ✅ I wonder why the build doesn’t show up here in GitHub...

triplef:
I’ve updated the patch to also correctly release a previous completion block when setting a new one.

triplef:
> However, I think the preferred way to do this is to use the ASSIGN and DESTROY macros; as I understand it they are _supposed_ to work for blocks, and should result in code that can be built with/without ARC.

That makes sense, and I’ve updated the patch to use macros instead. I’m using ASSIGNCOPY so that stack blocks are copied to the heap, which I think is important. (Also [Apple’s completionBlock property](https://developer.apple.com/documentation/foundation/nsoperation/1408085-completionblock) is marked as `copy`.)

PR #119, opened 2020-04-13
Normalize annoucement file between core libraries

ivucica:
While preparing the email, I noticed the ANNOUNCE file for `-make` doesn't mention the GPG key, and there are other subtle differences.

I'm creating this pull request across `make`, `base`, `gui` and `back`.


fredkiefer:
@rfm could you please check the failing tests? They are about NSURL and probably related to your latest change in NSURLProtocol.

PR #126, opened 2020-05-04
Bugfix/nsurl components

alotorev:
- Fixed missed autorelease 
- Fixed NSURLComponents when initialized with an NSURL containing percent-encoded query
- Updated NSURLQueryItem initializers to match OSX behavior
- Added test to cover some of the cases (checked with OSX implementation)

### DIFF
--- a/Source/NSURL.m
+++ b/Source/NSURL.m
@@ -2116,17 +2116,26 @@ GS_PRIVATE_INTERNAL(NSURLQueryItem)
   return AUTORELEASE(newQueryItem);
 }
 
-- (instancetype)initWithName:(NSString *)name
-                       value:(NSString *)value
+- (instancetype) init
 {
   self = [super init];
-  if (self != nil)
-    {
-      GS_CREATE_INTERNAL(NSURLQueryItem);
-      
-      ASSIGNCOPY(internal->_name, name);
-      ASSIGNCOPY(internal->_value, value);
-    }
+  if(self != nil)
+  {
+    GS_CREATE_INTERNAL(NSURLQueryItem);
+    ASSIGN(internal->_name, @""); //OSX behaviour is to set an empty string for the name property
+  }
+  return self;
+}
+
+- (instancetype)initWithName:(NSString *)name 
+                       value:(NSString *)value
+{
+  self = [self init];
+  if(self != nil)
+  {
+    if(name)ASSIGNCOPY(internal->_name, name);
+    ASSIGNCOPY(internal->_value, value);
+  }
   return self;
 }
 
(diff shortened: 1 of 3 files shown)


Source/NSURL.m
```diff
@@ -2454,7 +2461,7 @@ - (void) setURL: (NSURL *)url
   [self setUser: [url user]];
   [self setPassword: [url password]];
   [self setPath: [url path]];
-  [self setQuery: [url query]];
+  [self setPercentEncodedQuery:[url query]];
```
alotorev, line 2464: NSURL's query part can contain only characters from NSURLQueryCharacterSet, all other characters should be percent-encoded. Setting query in initializer with -setQuery leads name and value of NSURLQueryItem stay percent-encoded, and query property returns twice percent-encoded string.

Source/NSURL.m
```diff
@@ -2256,12 +2265,10 @@ - (instancetype) init
...
 {
-  self = [self init];
-  if (self != nil)
-    {
-      [self setString: URLString];
-    }
-  return self;
+  //OSX behavior is to return nil for a string which cannot be used to initialize valid NSURL object
+  NSURL* url = [NSURL URLWithString:URLString];
+  if(url) return [self initWithURL:url resolvingAgainstBaseURL:NO];
```
fredkiefer, line 2270: Could you please use curly brackets here?

Tests/base/NSURL/basic.m
```diff
@@ -339,6 +339,23 @@ int main()
...
+  PASS_EQUAL(item.value, nil, "NSURLQueryItem.value should be nil");
+    
+  //OSX behavior is to return query item with an empty string name
+  item = [[NSURLQueryItem alloc] initWithName:nil value:nil];
+  PASS_EQUAL(item.name, @"", "NSURLQueryItem.name should not be nil");
+  PASS_EQUAL(item.value, nil, "NSURLQueryItem.value should be nil");
+    
+  item = [[NSURLQueryItem alloc] initWithName:@"myName" value:@"myValue"];
+  PASS_EQUAL(item.name,  @"myName", "NSURLQueryItem.name should not be nil");
+  PASS_EQUAL(item.value, @"myValue", "NSURLQueryItem.value should be nil");
```
fredkiefer, line 356: A "not" is missing here.

Source/NSURL.m
```diff
@@ -2116,17 +2116,26 @@ + (instancetype)queryItemWithName:(NSString *)name
   return AUTORELEASE(newQueryItem);
 }
 
-- (instancetype)initWithName:(NSString *)name
-                       value:(NSString *)value
+- (instancetype) init
```
fredkiefer, line 2119: I would prefer an "init" method that calls "initWithName: nil value: nil". The later setting name to @"" if it is nil. But your code is correct as well.

Source/NSURL.m
alotorev: Ok

Source/NSURL.m
alotorev: Did the way you suggested.

alotorev:
Are there any other issues with this pull request?

PR #129, opened 2020-05-05
Add NSInputStream support for reading Android assets from main bundle.

triplef:
Adds support for reading Android assets from main bundle using NSInputStream, as was previously done for NSFileHandle in #45.

Sample usage:
```objc
NSString *path = [NSBundle.mainBundle pathForResource:@"file" ofType:@"txt"];
NSInputStream *stream = [NSInputStream inputStreamWithFileAtPath:path];
...
```


fredkiefer:
Would it be possible to reuse some of the code from GSFileHandle here?

triplef:
At a glance it should be possible to swap the entire GSFileInputStream and GSFileOutputStream implementation to forward to GSFileHandle, but that would be a bigger refactoring and I’m not sure it’d be worthwhile considering that these classes are already pretty lean.

I definitely don’t think it would be worthwhile just for Android, as it would add further fragmentation to the code path instead of reducing it. As-is, I basically just added a second (equivalent) code path for open/read/seek/close with an Android asset.

rfm:
I'm happy with that, and agree that using  GSFileHandle would not be worthwhile.

PR #130, opened 2020-05-05
Implement -[NSXMLParser initWithStream:].

triplef:
This uses an I/O parser (streaming the data into the parser) instead of a push parser, which should result in reduced memory requirements when parsing large XML files over `-initWithData:`.

Also [changes](https://github.com/gnustep/libs-base/pull/130/files#diff-049b8082c6e5b42981048909755c7aaeL483-R487) `-[NSXMLParser initWithContentsOfURL:]` to use `-[GSXMLParser initWithSAXHandler:withContentsOfURL:]` for consistency, which internally does the same thing as before, i.e. read the URL into an NSData.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2020-04-07  Frederik Seiffert <frederik@algoriddim.com>
+
+	* Headers/Foundation/NSXMLParser.h:
+	* Headers/GNUstepBase/GSXML.h:
+	* Source/Additions/GSXML.m:
+	* Source/NSXMLParser.m:
+	* Tests/base/NSXMLParser/parse.m:
+	Implement -[NSXMLParser initWithStream:].
+
 2020-04-26  Fred Kiefer <fredkiefer@gmx.de>
 
 	* Source/NSLocale.m: Respect NSLocaleCalendarIdentifier if
--- a/Headers/Foundation/NSXMLParser.h
+++ b/Headers/Foundation/NSXMLParser.h
@@ -36,7 +36,7 @@
 extern "C" {
 #endif
 
-@class NSData, NSDictionary, NSError, NSString, NSURL;
+@class NSData, NSDictionary, NSError, NSInputStream, NSString, NSURL;
 
 /**
  * Domain for errors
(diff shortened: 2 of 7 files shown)


triplef:
I didn’t know there was another XML parser – where can I find it?

What you’re suggesting sounds like a separate piece of work though, and from what I’ve seen of the libxml2 usage in GSXML.m probably much more extensive than this change (unless it’s already implemented to be able to just swap out GSXMLParser for it?).

Wouldn’t it still make sense to add this API extension and tests, so both of these are in place when switching the parser?

rfm:
> On 6 May 2020, at 13:08, Frederik Seiffert <notifications@github.com> wrote:
> 
> 
> I didn’t know there was another XML parser – where can I find it?

It's in NSXMLParser.m

> What you’re suggesting sounds like a separate piece of work though, and from what I’ve seen of the libxml2 usage in GSXML.m probably much more extensive than this change (unless it’s already implemented to be able to just swap out GSXMLParser for it?).

It's the standard implementation of NSXMLParser on systems where libxml2 is not available, and is also present on systems where libxml2 is used, so we can handle documents that libxml2 won't permit (so maybe it should be standard on all systems)

> Wouldn’t it still make sense to add this API extension and tests, so both of these are in place when switching the parser?

Perhaps ...
I'll merge it, then we can look at disabling the libxml2 based parser by default.

triplef:
> It's the standard implementation of NSXMLParser on systems where libxml2 is not available, and is also present on systems where libxml2 is used, so we can handle documents that libxml2 won't permit (so maybe it should be standard on all systems)

Ah I see what you are saying now – I didn’t realize that there were two different implementations for NSXMLParser that are both in use. I agree that this patch should have supported both.

Looking over the native parser it looks like it’s accessing the buffer directly via `this->cp`. Would you have a suggestion how to implement support for NSInputStream in this context? I guess it should read from the stream in chunks somewhere?

rfm:
> On 6 May 2020, at 15:28, Frederik Seiffert <notifications@github.com> wrote:
> 
> 
> It's the standard implementation of NSXMLParser on systems where libxml2 is not available, and is also present on systems where libxml2 is used, so we can handle documents that libxml2 won't permit (so maybe it should be standard on all systems)
> 
> Ah I see what you are saying now – I didn’t realize that there were two different implementations for NSXMLParser that are both in use. I agree that this patch should have supported both.
> 
> Looking over the native parser it looks like it’s accessing the buffer directly via this->cp. Would you have a suggestion how to implement support for NSInputStream in this context? I guess it should read from the stream in chunks somewhere?

Yes I think reading chunks is the way to go.
In the meantime I changed the testcases to test both parsers (there was already code in there to easly to switch which parser you are testing, but I think always testing both makes more sense), restructured the code to make the usage of the two parsers clearer, and added a dummy implementation of -initWithStream: for the native (sloppy) parser, just so that the stream testcases with that parser don't crash.

triplef:
Thank you for fixing this up, much appreciated!

PR #131, opened 2020-05-07
NSLocale improvements

triplef:
- Use "instancetype" for instance-returning methods.
- Fix method signature of -displayNameForKey:value: to match Apple, and since implementation expects NSString.


fredkiefer:
This change looks good to me. The build fails are cased by a previous commit on GSXML that needs to be fixed separately.

PR #133, opened 2020-05-11
Start implementing some secure coding methods.

gcasa:
Just a simple change to return an error when secured coding is asked for.   We should discuss how to actually implement secure coding.


Headers/Foundation/NSKeyedArchiver.h
```diff
@@ -87,6 +87,16 @@ extern "C" {
  */
 + (NSData*) archivedDataWithRootObject: (id)anObject;
 
+  
+/**
```
fredkiefer, line 91: You should add guards for MacOS 10.13 around this method.

Source/NSKeyedArchiver.m
```diff
@@ -460,6 +460,26 @@ + (NSData*) archivedDataWithRootObject: (id)anObject
   return AUTORELEASE(d);
 }
 
++ (NSData *) archivedDataWithRootObject: (id)anObject
```
fredkiefer, line 463: I think this method should not call archivedDataWithRootObject: instead it should be the other way around. This is the more general method and it should just have the code from archivedDataWithRootObject: and set the flag requiresSecureCoding on the created NSKeyedArchiver. As this flag is also still missing there is not much use in adding this method.
There is also no reason to return an error when secure coding is requested.

Headers/Foundation/NSKeyedArchiver.h
gcasa: Done

Source/NSKeyedArchiver.m
gcasa: Done

gcasa:
I am merging this since it seems relatively simple.  This is just a stop gap until we have support.

Source/NSKeyedArchiver.m
```diff
@@ -430,34 +430,49 @@ - (id) _encodeObject: (id)anObject conditional: (BOOL)conditional
 
 @implementation	NSKeyedArchiver
 
++ (NSData *) archivedDataWithRootObject: (id)anObject
+                  requiringSecureCoding: (BOOL)requiresSecureCoding
+                                  error: (NSError **)error
+{
+  NSData *d = nil;
+
+  if (requiresSecureCoding == NO)
```
fredkiefer, line 439: The idea was to call setRequiresSecureCoding: with this value on "a" further down in line 449. And of course any exceptions need to be converted in to NSError.

gcasa: Okay, I will open a "real" secure coding branch for this.   I will do that now.

PR #136, opened 2020-05-15
Set up CI for Windows using MinGW and GCC

triplef:
This adds Windows as a CI target using MinGW-w64 by following these instructions:
https://docs.travis-ci.com/user/reference/windows/#how-do-i-use-msys2

The required packages are taken from the [MSYS2 instructions](http://wiki.gnustep.org/index.php/Installation_MSYS2), with the addition of packages for ICU.

The current status is:
* 64-bit target:
  * builds [successfully](https://travis-ci.org/github/gnustep/libs-base/jobs/687727959#L6167) (albeit with many more warnings than on Unices)
  * running the tests [fails](https://travis-ci.org/github/gnustep/libs-base/jobs/687727959#L6403), but I’m not sure why
* 32-bit target [fails building](https://travis-ci.org/github/gnustep/libs-base/jobs/687727958#L1196)

I’m not sure about the amount of work required to get tests working and 32-bit builds to compile, but I’d appreciate any feedback.

(As a side note, for some reason chaining commands like `make && make install && make check` does not seem to work with the MSYS wrapper, so I’m running make/install/check as separate steps, which means that subsequent steps are still executed even if previous steps fail.)

I’ve currently marked the Windows targets so that they are allowed to fail without failing the whole CI-run, so this could already be merged as is. This would at least allows to verify progress towards getting Windows to work on both architectures and including tests.

This also reworks CI setup to list build matrix explicitly with names.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,73 +1,141 @@
 ---
-os: linux
 language: cpp
 dist: xenial
-compiler:
-    - clang
-    - gcc
-env:
-    - LIBRARY_COMBO=gnu-gnu-gnu
-    - LIBRARY_COMBO=ng-gnu-gnu RUNTIME_VERSION=gnustep-1.9
-    - LIBRARY_COMBO=ng-gnu-gnu BASE_ABI=--disable-mixedabi RUNTIME_VERSION=gnustep-1.9
-    - LIBRARY_COMBO=ng-gnu-gnu RUNTIME_VERSION=gnustep-2.0
-    - LIBRARY_COMBO=ng-gnu-gnu BASE_ABI=--disable-mixedabi RUNTIME_VERSION=gnustep-2.0
 jobs:
-    exclude:
-        - compiler: gcc
-          env: LIBRARY_COMBO=ng-gnu-gnu RUNTIME_VERSION=gnustep-1.9
-        - compiler: gcc
-          env: LIBRARY_COMBO=ng-gnu-gnu BASE_ABI=--disable-mixedabi RUNTIME_VERSION=gnustep-1.9
-        - compiler: gcc
-          env: LIBRARY_COMBO=ng-gnu-gnu RUNTIME_VERSION=gnustep-2.0
-        - compiler: gcc
-          env: LIBRARY_COMBO=ng-gnu-gnu BASE_ABI=--disable-mixedabi RUNTIME_VERSION=gnustep-2.0
-        - compiler: clang
+    include:
+        - name: "Linux GCC"
+          os: linux
+          compiler: gcc
           env: LIBRARY_COMBO=gnu-gnu-gnu
-before_install:
-    - sudo apt-get -qq update
-    - sudo apt-get install -y cmake pkg-config libgnutls28-dev libgmp-dev libffi-dev libicu-dev libxml2-dev libxslt1-dev libssl-dev libavahi-client-dev zlib1g-dev
-    - >
-        if [ $LIBRARY_COMBO = 'gnu-gnu-gnu' ];
-        then
-          if [ $CC = 'gcc' ];
-          then
-            sudo apt-get install -y gobjc;
-          fi;
-          sudo apt-get install -y libobjc-4.8-dev libblocksruntime-dev;
-        else
-          curl -s -o - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -;
-          sudo apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-9 main" && sudo apt-get update -qq;
-          sudo apt-get install -y clang-9 libkqueue-dev libpthread-workqueue-dev;
-          sudo update-alternatives   --install /usr/bin/clang   clang   /usr/bin/clang-9   10 \
-                                     --slave   /usr/bin/clang++ clang++ /usr/bin/clang++-9;
-          export PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/clang-7.0.0\/bin//');
-          if [ "$RUNTIME_VERSION" = "gnustep-2.0" ];
-          then
-             sudo update-alternatives --install "/usr/bin/ld" "ld" "/usr/bin/ld.gold" 10;
-          fi;
-        fi;
-    # libdispatch requires a fairly recent version of cmake
-    - >
-        if [ $LIBRARY_COMBO = 'ng-gnu-gnu' ];
-        then
-          curl -LO https://cmake.org/files/v3.15/cmake-3.15.5-Linux-x86_64.tar.gz;
-          tar xf cmake-3.15.5-Linux-x86_64.tar.gz;
-          mv cmake-3.15.5-Linux-x86_64 $HOME/cmake;
-          export PATH=$HOME/cmake/:$HOME/cmake/bin:$PATH
-        fi;
+        - name: "Linux Clang gnustep-1.9"
+          os: linux
+          compiler: clang
+          env: LIBRARY_COMBO=ng-gnu-gnu RUNTIME_VERSION=gnustep-1.9
+        - name: "Linux Clang gnustep-1.9 disable-mixedabi"
+          os: linux
+          compiler: clang
+          env: LIBRARY_COMBO=ng-gnu-gnu RUNTIME_VERSION=gnustep-1.9 BASE_ABI=--disable-mixedabi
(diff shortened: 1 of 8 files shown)


fredkiefer:
I like these changes in general as they make the build process more explicit compared to the old way. There is one bit that does not appeal to me, that is the rework of the chained commands you write about in your comment above. This makes it harder to spot the issue when the build fails. Maybe we could have the new way only for Windows?

As for the failing tests on Windows 64 bit. The logs say that the compiled library is not found. You should switch on the message output for the "make install" step to see what goes on there.

For Windows 32 bit we seem to require a way to differentiate between old MSYS and the new one. Currently we have this block in GSConfig.h:

'#if defined(__WIN64__)
#include <winsock2.h>
#include <windows.h>
#else
#include <windows.h>
#include <winsock2.h>
#endif'

We should rather decide on whether MSYS or MSYS2 gets used. But this has little to do with your changes and should not block merging these.

triplef:
Thank you both for your feedback. I’ll work on the points mentioned and let you know when they are ready for review.

> We should rather decide on whether MSYS or MSYS2 gets used. But this has little to do with your changes and should not block merging these.

MSYS2 [defines](https://www.msys2.org/wiki/Porting/#platform-checks) the `__MSYS__` preprocessor define. Does one of you know whether MSYS sets that as well, or where one could check this?

fredkiefer:
If this change works for MSYS2 it is good enough for me to merge the whole PR.

triplef:
So unfortunately `__MSYS__` is only defined in the pure MSYS(2) environment, not in MinGW. Both MinGW and MinGW-w64 define `__MINGW32__`, which we already use. I cannot find a way to detect just MinGW-w64.

From what I’ve read most people seem to be using MSYS2/MinGW-w64 these days, so I’m wondering if it might be an option to drop support for the old MSYS/MinGW and assume MinGW-w64?

As mentioned this also doesn’t have to be decided as part of this PR. Unless you have other feedback I’ll just leave the winsock2.h includes untouched for now.

rfm:
Yes,  I had been thinking the same thing.  My impression is that the old 32bit msys stuff has been obsolete for long enough that there is probably no point supporting it.

triplef:
Great! Does that mean we can always import winsock2.h before windows.h, or is there another way to build on Windows that I’m not aware of?

fredkiefer:
We used to have Cygwin support  as well, but I don't know if this is still relevant. Having both MSYS2 supported would be enough for me.
Looking at the warnings we get from NSException.h, maybe it would help to #undef setjmp first? But again, this does not have to be done in your pull request.

triplef:
Somehow the Windows builds are now failing with this error after I deleted the caches... :(

> gpg: keyserver refresh failed: General error
> ==> ERROR: A specified local key could not be updated from a keyserver.

rfm:
> On 18 May 2020, at 17:37, Frederik Seiffert <notifications@github.com> wrote:
> 
> ERROR: A specified local key could not be updated from a keyserver.


I have never seen that message before, but google tells me that it's likely from gpg
I am guessing that's part of the travis CI stuff rather than anything to do with gnustep ... perhaps a transient network error?

rfm:
looking at the latest failure logs I see:

$ ./travis-deps.sh
error: failed to prepare transaction (could not satisfy dependencies)
ERROR: The process "gpg-agent.exe" not found.

Could it be that the last change (Windows CI: attempt setting PATH inside MinGW shell) has broken the PATH information Travis needs?

triplef:
I think it was a case of bad timing that I deleted the Travis caches yesterday just after the MSYS2 runtime was updated using a new package format that doesn’t seem to play well with the pacman version this is using. I’ve written on the Travis forums to ask for help:
https://travis-ci.community/t/error-installing-msys2-runtime-zst-package/8687

triplef:
So the MSYS2 issues got resolved upstream.

However I’m stills struggling to get GNUstep installed in a way where it actually works for the tests, which requires the DLLs to be in the PATH.

I haven’t been able to modify the PATH inside the MinGW environment, as I think it gets overridden by the `RefreshEnv.cmd` that’s called as part of the $mingw prefix, so I’ve been trying to just install it into the existing PATH, which is this:
```
PATH=/mingw64/bin:/usr/local/bin:/usr/bin:/bin:/c/tools/ruby27/bin:/c/Windows/system32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/c/Windows/System32/OpenSSH:/c/ProgramData/GooGet:/c/Program Files/Google/Compute Engine/metadata_scripts:/c/Program Files (x86)/Google/Cloud SDK/google-cloud-sdk/bin:/c/Program Files/Google/Compute Engine/sysprep:/c/Program Files/Docker:/c/ProgramData/chocolatey/bin:/c/Program Files/CMake/bin:/c/Program Files/Git/cmd:/c/Program Files/LLVM/bin:/c/Program Files/dotnet:/c/Program Files (x86)/Windows Kits/10/Windows Performance Toolkit:/c/tools/BCURRAN3:/c/Users/travis/AppData/Local/Microsoft/WindowsApps:/c/ProgramData/chocolatey/lib/mingw/tools/install/mingw64/bin:/
```

The lastest builds are installing into `/mingw64/bin`, which seems to succeed for Make, but when I list /mingw64/bin right after the installation it doesn’t contain `gnustep-config`.

I’m not really sure what else to try. If anyone has suggestions or if there’s anything that should be set up differently here I’m all ears (I haven’t worked with either Travis CI or MSYS before, so I’m likely missing something).

triplef:
Ok this finally works using `setx -m PATH` to edit the PATH inside the MinGW environment (I really should have figured this this much sooner...), and the tests are running on x86_64 (though many of them failing, but that’s for another time). Unfortunately the NSURL tests also time out after 10 minutes without output – maybe there’s a way we can enable additional progress output, or speed them up?

i686 is failing to build Base with the following error (not sure about this – sounds like the system already has this function?):
```
inet_pton.m:66:1: error: conflicting types for 'inet_pton'
   66 | inet_pton(int af, const char *src, void *dst)
      | ^~~~~~~~~
In file included from GSNetwork.h:49,
                 from inet_pton.m:34:
C:/tools/msys64/mingw32/i686-w64-mingw32/include/ws2tcpip.h:458:32: note: previous declaration of 'inet_pton' was here
  458 | WINSOCK_API_LINKAGE INT WSAAPI InetPtonA(INT Family, LPCSTR pStringBuf, PVOID pAddr);
      |                                ^~~~~~~~~
```

In addition to setting up the Windows CI, I also made these changes as discussed to fix some of the the most pervasive warnings:

- Undef setjmp (and longjmp for good measure).
- Always include winsock2.h first in various places.

As before, the Windows builds are marked as allowed to fail, so from my side I’d be happy to merge this as the general setup is now working.

triplef:
Rebased on master to fix conflict in ChangeLog.

rmottola:
> 
> 
> Yes, I had been thinking the same thing. My impression is that the old 32bit msys stuff has been obsolete for long enough that there is probably no point supporting it.

Sorry - I strongly disagree, I missed this discussion! MSYS needs to stay until we have a reliable replacement. Old MSYS is currently developed upstream and also the only target which gives a 100% reliable windows environment including GUI, Network and all applications.
MSYS2 does not, despite all my attempts, also it is geared for 64Bit, so better use it for 64bit and keep MSYS for 32bit.

rmottola:
> 
> 
> Please, do not just revert this change. You could try to improve on it by testing a different guard, eg **MINGW32**.
> The good thing is you are able to test on old MSYS locally and the Travis system will test your pull request with the two new MSYS2 systems. That way the most important Windows build systems will get tested.

I did tackle this issue already more than year ago when making the first ming64 builds.

Unfortunately, I found no way to differentiate reliably the build systems: on one side the projects try to be compatible, on the other side they aren't (reason because they diverged I fear).
I tried "MINGW" which is not defined at all anymore.
__MINGW32__ is instead defined both by MSYS and MSYS2, making the MSYS2-32bit undetectable.

Another way would be to "fix" the error of the includes on either side.. very hard

triplef:
> Unfortunately, I found no way to differentiate reliably the build systems

Yeah that’s what I found too. Maybe we could shift the detection to configure instead and set some define to differentiate MSYS/MSYS2 in sources? Checking `$host` for `-w64` [might work](https://www.msys2.org/wiki/Porting/#platform-checks).

PR #138, opened 2020-05-25
Store NSNumber instead of NSString for NSUserDefaults -setBool:forKey:.

triplef:
Changes `-[NSUserDefaults setBool:forKey:]` to store a bool NSNumber instead of a YES/NO string.

I am not sure why a string was used here, but this seems more correct, and it should make no difference when reading the value via `-[NSUserDefaults boolForKey:]` or `[[NSUserDefaults objectForKey:X] boolValue]`.

We had a code path that was observing NSUserDefaults and was expecting the new value to be an NSNumber (which is the case on Apple platforms).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2020-05-25  Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSUserDefaults.m:
+	Store NSNumber instead of NSString for NSUserDefaults -setBool:forKey:.
+
 2020-05-14  Frederik Seiffert <frederik@algoriddim.com>
 
 	* Headers/Foundation/NSException.h:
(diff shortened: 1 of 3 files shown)


triplef:
That sounds good. Looks like we already test a basic setBool:forKey / boolForKey: roundtrip:
https://github.com/gnustep/libs-base/blob/master/Tests/base/NSUserDefaults/general.m#L40-L42

Would adding a test that checks objectForKey: returns an NSNumber after setting a bool be sufficient here, or was there another kind of test you had in mind?

rfm:
> On 26 May 2020, at 10:44, Frederik Seiffert <notifications@github.com> wrote:
> 
> 
> That sounds good. Looks like we already test a basic setBool:forKey / boolForKey: roundtrip:
> https://github.com/gnustep/libs-base/blob/master/Tests/base/NSUserDefaults/general.m#L40-L42
> 
> Would adding a test that checks objectForKey: returns an NSNumber after setting a bool be sufficient here, or was there another kind of test you had in mind?

Yes, I think that would be sufficient, thanks.

triplef:
I’ve added this test as well as some additional tests for NSUserDefaults along the same lines, and I verified these same tests also pass on macOS 10.15.

rfm:
That's great,  thanks a lot.
Merged.

PR #140, opened 2020-06-03
Fix finding libiconv via cross.config

triplef:
Adds defining `HAVE_ICONV` to the configure cross-compilation branches.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2996,7 +2996,11 @@ int main(int argc,char **argv)
 ,
   found_iconv=no
 ,
-  found_iconv="$cross_found_iconv_libc",
+  found_iconv="$cross_found_iconv_libc";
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -2997,6 +2997,10 @@ int main(int argc,char **argv)
   found_iconv=no
 ,
   found_iconv="$cross_found_iconv_libc",
+  if test "$found_iconv" = "yes"; then
```
fredkiefer, line 3000: I see no lines in configure that are caused by this change. Did you forget to regenerate configure or is this change not needed?

triplef: Good catch! This was caused by the line above incorrectly ending in a comma instead of a semicolon.

PR #142, opened 2020-06-06
Msys fix

rmottola:
Support both old MinGW1 als current MinGW2-32 and MINGW2-64

Clean up another header of "windows hacks" and centralize ifdef's.

### DIFF
--- a/Headers/GNUstepBase/GSConfig.h.in
+++ b/Headers/GNUstepBase/GSConfig.h.in
@@ -306,8 +306,18 @@ typedef	struct {
 #define WINVER Windows2000
 #endif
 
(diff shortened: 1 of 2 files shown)


rmottola:
Frederik says it looks good - the tests have the same issues as before.

rfm:
> On 6 Jun 2020, at 09:45, Riccardo <notifications@github.com> wrote:
> 
> 
> Frederik says it looks good - the tests have the same issues as before.
> 

It looks reasonable to me (wouldn't break anything else)
As I have no test environment (new computer still on order, and they haven't started building it yet, so it may be some time before I do), I can't actually try it out, but I merged it anyway.

PR #145, opened 2020-06-11
Output config.log if build fails on CI.

triplef:
This should help with tracking down compilation issues like the ones we are seeing on Windows 32-bit:
https://travis-ci.org/github/gnustep/libs-base/jobs/697202242#L23769

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,7 +126,7 @@ before_script: |
 script:
     # configure and make need to be executed via MinGW shell on Windows ($mingw is undefined on Linux)
     - $mingw ./configure $BASE_ABI $CONFIGURE_OPTS || (cat config.log && false)
-    - $mingw make && $mingw make install && $mingw make check || (cat Tests/tests.log && false)
+    - ($mingw make || (cat config.log && false)) && $mingw make install && $mingw make check || (cat Tests/tests.log && false)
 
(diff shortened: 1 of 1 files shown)


rmottola:
I think this one is obsolete now?

triplef:
Only as long as there are no issues building. 😉 I think this it’s helpful to see the config log when the build fails, so I’d still be happy to see this merged.

triplef:
I’ve rebased the branch and fixed the conflict.

fredkiefer:
I am still in favour of merging this change but we need to fix the base tests before doing so. Just to be sure that this change does not produce any issues on one of our platforms.

rfm:
>  but we need to fix the base tests before doing so.
What fixes are needed?

fredkiefer:
> but we need to fix the base tests before doing so.
> What fixes are needed?

base/NSOperation/threads.m:
2081Failed test:     threads.m:213 ... thread exit occurs after six seconds

rfm:
How can we make that testcase fail (it always passes for me)?
I'm guessing that, since it's a timing related test, it may be giving up too soon on slow systems.
So I changed the testcase to allow up to 10 seconds rather than giving up waiting after six.

triplef:
Thanks for looking into this! I’ve rebased this change and the test all pass now. 👍

PR #146, opened 2020-06-22
Fix MSYS2-32

rmottola:
after some discussion with Fred and Frederick and several local test, I come up with this.
Although it is a mystery (the whole cause is that in 32bit MSYS2, inet_ntop is not found in the same library where it is found in 64bit version, so we need a replacement.

Thus I created a replacement which does not clash and works better. If we can in future make the configure check better, then fine.
This pull-request should make #34 obsolete too and concentrate on why URL stuff times out (now coherently in 32bit and 64bit MSYS2).


rmottola:
@rfm would you mind approving or either commenting this fix? then we can delete this branch too.

PR #147, opened 2020-06-23
Don't return from _setLocaleIdentifier: if my->cal is NULL

nilium:
Add a check for my->cal being non-NULL to _setLocaleIdentifier to
prevent faulty initialization of NSCalendar* using ICU.

Due to a quirk of some environments (like my weird system), a locale
isn't necessarily going to be a non-empty value. When this happens, the
locale ID ends up being an empty string, for which isEqualToString:nil
returns true. So, following from this, when _setLocaleIdentifier: is
sent as part of initWithCalendarIdentifier:, isEqualToString:nil returns
true and my->cal remains NULL, because it returns early.

Later on, in NSCalendar/basic.m tests, this causes a segfault because
ucal_equivalentTo is called with two NULL pointers. This happens
because my->cal is not initialized.

So, to ensure my->cal is initialized, check that it's not NULL before
checking if an early return is possible from _setLocaleIdentifier:.

### DIFF
--- a/Source/NSCalendar.m
+++ b/Source/NSCalendar.m
@@ -210,7 +210,7 @@ static NSRecursiveLock *classLock = nil;
 
 - (void) _setLocaleIdentifier: (NSString *) identifier
 {
-  if ([identifier isEqualToString: my->localeID])
+  if (my->cal != NULL && [identifier isEqualToString: my->localeID])
     {
       return;
     }


rfm:
Thanks, but it seems to me that the issue here might be a bug in -isEqualToString: since it seems odd that this method should return YES for nil.  I need to find out what this method is supposed to do when given a nil argument.

nilium:
That was bothering me, but it seemed externally consistent (the bytes and range of a nil NSString should be all zeroes, so I'm still wrestling with it mentally). Happy to close this if it turns out isEqualToString: should be changed instead.

rfm:
I just added a couple of testcases (for -isEqual: and -isEqualToString:) comparing an empty string with nil.  Running them on OSX tells me that the methods should return NO when given a nil argument.  I will fix them.

nilium:
Sounds good. Thanks for checking.

rfm:
OK, changes committed.  I think the old string implementations were correct and the problem was with the tiny strings used by clang and the more modern objc runtime (and possibly also with string subclasses using the abstract base class implementation).  Please check out the changes and see if they fix your issue.

nilium:
Confirmed NSCalendar's no longer segfaulting (`make check` also passes now). Thanks!

PR #149, opened 2020-07-08
Improve NSLog output on Android.

triplef:
Always set GSLogSyslog flag on Android instead of using #ifdefs. This enables other checks for this flag (e.g. not logging date/time to syslog) that were previously inactive on Android.

This is an improvement over #43.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2020-07-08  Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSLog.m,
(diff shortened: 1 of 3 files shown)


rfm:
I was a bit worried, looking at the diff, but then I realised that the change to set the flag *is* in a section of code already ifdeffed to be android specific.

triplef:
Thanks! Yeah process initialization on Android is a bit special, and as there are no process arguments on Android this seemed like the cleanest way to enable using syslog throughout.

PR #152, opened 2020-07-13
Update NSURLRequest header to use "instancetype".

triplef:


fredkiefer:
Shouldn't we update NSURLRequest.m as well?

triplef:
> Shouldn't we update NSURLRequest.m as well?

Yes you’re absolutely right. I’ve updated the commit.

PR #154, opened 2020-08-03
Add various NSURL methods.

triplef:
- URLByAppendingPathComponent:isDirectory:
- isFileReferenceURL (always returns NO)
- fileReferenceURL
- filePathURL

Note that this is not an actual implementation of "file reference URLs" (I am not sure if there would be a way to support these in a cross-platform manner).

### DIFF
--- a/Source/NSURL.m
+++ b/Source/NSURL.m
@@ -1539,6 +1539,29 @@ static NSUInteger	urlAlign;
   return [[self path] lastPathComponent];
 }
 
+- (BOOL) isFileReferenceURL
+{
+  return NO;
+}
+
+- (NSURL *) fileReferenceURL
+{
+  if ([self isFileURL]) 
+    {
+      return self;
+    }
+  return nil;
+}
+
+- (NSURL *) filePathURL
+{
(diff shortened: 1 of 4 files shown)


Source/NSURL.m
```diff
@@ -1539,6 +1539,29 @@ - (NSString*) lastPathComponent
...
 }
 
+- (BOOL) isFileReferenceURL
+{
+  return NO;
+}
+
+- (NSURL *) fileReferenceURL
+{
+  if ([self isFileURL])
```
fredkiefer, line 1549: The check here should be for isFileReferenceURL. For isFileURL a new reference should be created and currently we don't have any means to do so.

triplef: You’re of course right in terms of strictly copying Apple APIs. However, given that we cannot reproduce this exact functionality and file reference URLs are just "special URLs", my thought was that returning self in case of file URLs would be better in terms of API compatibility, as clients wouldn’t expect this method to return nil for a file URL. I’m open to your thoughts tho.

rfm (approved):
Seems reasonable to me.

PR #159, opened 2020-10-12
[NSData rangeOfData:options:range]

adamfox-tp:
This implementation was adapted from [NSString rangeOfString:options:range].


Source/NSData.m
```diff
@@ -1082,6 +1082,119 @@ - (NSData*) subdataWithRange: (NSRange)aRange
...
+          if ((mask & NSDataSearchBackwards) == NSDataSearchBackwards)
+            {
+              searchRange.location = NSMaxRange(searchRange) - countOther;
+              searchRange.length = countOther;
+            }
+          else
+            {
+              searchRange.length = countOther;
+            }
+          if (memcmp(&bytesSelf[0], &bytesOther[0], countOther) == 0)
```
fredkiefer, line 1136: Why not `memcmp(bytesSelf, &bytesOther, countOther)`?

I think that this is also wrong as the anchor search could start from the end, so we should not just compare the start of self.

Source/NSData.m
```diff
@@ -1082,6 +1082,119 @@ - (NSData*) subdataWithRange: (NSRange)aRange
...
+  if (searchRange.length < countOther)
+    {
+      /* Range to search is smaller than data to look for.
+       */
+      result = NSMakeRange(NSNotFound, 0);
+    }
+  else
+    {
+      if ((mask & NSDataSearchAnchored) == NSDataSearchAnchored
+        || searchRange.length == countOther)
```
fredkiefer, line 1123: Or do you mean `&&` here?

Tests/base/NSData/search.m
```diff
@@ -0,0 +1,81 @@
+#import <Foundation/Foundation.h>
+
+#import "Testing.h"
+
+static BOOL rangesEqual(NSRange r1, NSRange r2)
```
fredkiefer, line 5: You could use NSEqualRanges instead.

Tests/base/NSData/search.m
```diff
@@ -0,0 +1,81 @@
...
+  if (r1.length == 0 && r2.length == 0)
+    return YES;
+
+  return (r1.length == r2.length && r1.location == r2.location);
+}
+
+static void dataRange(char *s0, char *s1, NSDataSearchOptions opts,
+ 		     NSRange range, NSRange want)
+{
+  NSData *d0 = [NSData dataWithBytes:s0 length:strlen(s0)];
```
fredkiefer, line 19: There should be spaces after the colons.

Source/NSData.m
```diff
@@ -1082,6 +1082,119 @@ - (NSData*) subdataWithRange: (NSRange)aRange
...
+{
+  NSUInteger  length = [self length];
+  NSUInteger  countOther = [dataToFind length];
+  const void* bytesSelf = [self bytes];
+  const void* bytesOther = [dataToFind bytes];
+  NSRange     result;
+
+  GS_RANGE_CHECK(searchRange, length);
+  if (dataToFind == nil)
+    [NSException raise: NSInvalidArgumentException format: @"range of nil"];
```
fredkiefer, line 1100: In GNUstep code we prefer {} even if it is not strictly needed.

fredkiefer (changes requested):
There  are a few minor improvements that should be done before merging.

Tests/base/NSData/search.m
adamfox-tp: I can certainly do that, it's just what the NSString/test00.m did, so I wasn't sure if there was a reason for a special function.

Source/NSData.m
adamfox-tp: The indices were just specified because that's how it was done in [NSString rangeOfString:options:range], but I agree it would be clearer to change it.

I'm not quite sure it's wrong yet though, unless rangeOfString is wrong too.  Certainly I could have made a mistake and I'll add some tests to try to work out how it should behave with both NSDataSearchAnchored and NSDataSearchBackwards, but after a quick review it still appears to me to be working the same as rangeOfString.

Thank you!

Source/NSData.m
adamfox-tp: Similar to the prior comment, this was adapted directly from rangeOfString, and I think the code is assuming that an anchored search is the same as a search where self and other are the same length (of course just not caring if it's not true for self).  It does seem off with regard to a backwards search, so I'll probably have better feedback here after addressing the prior comment.

Source/NSData.m
fredkiefer: I never reviewed the code in NSString. A quick look shows that it probably used the indices to keep it similar to the other places where the compare starts from a specific index. I must admit consistency is good enough a reason to do it this way.

rfm: > The indices were just specified because that's how it was done in ...
I think there is a problem in that the code here, not in that it uses indices, but that it uses the wrong indices.
In NSString.m the indices used are zero because the relevant range of characters has been extracted from the string and placed in new unichar buffers, and we are doing a comparison of those unichar sequences.
In NSData.m bytesSelf and bytesOther are pointers to the start of the raw data, so the correct offsets must be used.  For the data being checked with an anchored search, that offset is searchRange.location which may not be zero.

Tests/base/NSData/search.m
rfm: The two things are subtly different.
NSEqualRanges checks to see if two ranges both have the same length and location,  but the function allows two ranges of zero length to be counted as equal.  I suspect that this comes from days when OSX would consistently return a zero length range when a string was not found, but would leave the location indeterminate.
I think using NSEqualRanges is preferable, as long as the testcase pass on OSX.

fredkiefer:
Are you still working on this patch? It would be nice to have an implementation of this method in GNUstep.

rfm (changes requested):
Correcting the memcmp call line 1136 to use searchRange.location is the only issue I see.

fredkiefer:
> Correcting the memcmp call line 1136 to use searchRange.location is the only issue I see.

if (memcmp(&bytesSelf[searchRange.location], &bytesOther[0], countOther) == 0)

PR #161, opened 2020-10-27
Add methods from 10.5

gcasa:
Add missing 10.5 methods.

### DIFF
--- a/Headers/Foundation/NSRunLoop.h
+++ b/Headers/Foundation/NSRunLoop.h
@@ -34,12 +34,18 @@ extern "C" {
 
 @class NSTimer, NSDate, NSPort;
 
+typedef NSString* NSRunLoopMode;
+  
 /**
  * Run loop mode used to deal with input sources other than NSConnections or
  * dialog windows.  Most commonly used. Defined in
  * <code>Foundation/NSRunLoop.h</code>.
  */
-GS_EXPORT NSString * const NSDefaultRunLoopMode;
+GS_EXPORT NSRunLoopMode const NSDefaultRunLoopMode;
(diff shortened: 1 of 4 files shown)


Headers/Foundation/NSRunLoop.h
```diff
@@ -34,12 +34,18 @@ extern "C" {
 
 @class NSTimer, NSDate, NSPort;
 
+typedef NSString* NSRunLoopMode;
```
fredkiefer, line 37: Maybe we should use this new type for all the cases where mode is used here (and in the long run all the places where we call a run loop).

Source/NSURLConnection.m
```diff
@@ -157,13 +157,37 @@ + (NSURLConnection *) connectionWithRequest: (NSURLRequest *)request
...
+  NSArray *modes = [NSArray arrayWithObject: mode];
+  [aRunLoop performSelector: @selector(start)
+                     target: self
+                   argument: nil
+                      order: 0
+                      modes: modes];
+}
+
+- (void) unscheduleFromRunLoop: (NSRunLoop *)aRunLoop 
+                       forMode: (NSRunLoopMode)mode
```
fredkiefer, line 184: It looks wrong that the mode argument gets ignored, but I see no way of using it.

fredkiefer (approved):
Looks good to me.

Headers/Foundation/NSRunLoop.h
gcasa: Okay, I will make this change.

Source/NSURLConnection.m
gcasa: That was my confusion as well.

PR #164, opened 2020-11-17
Add NSFileHandle URL initializers.

triplef:
* -fileHandleForReadingFromURL:error:
* -fileHandleForWritingToURL:error:
* -fileHandleForUpdatingURL:error:

Also use "instancetype" for all initializers.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2020-11-17  Frederik Seiffert <frederik@algoriddim.com>
+
+	* Headers/Foundation/NSFileHandle.h,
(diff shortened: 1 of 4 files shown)


fredkiefer (approved):
I think for now this patch is good enough. In the long run we should implement the methods the other way around and make the new methods the base ones.

triplef:
Yeah agreed, I just didn’t want to move too much around at this point.

I just pushed a small update that adds a missing `@class NSError` in NSFileHandle.h.

rfm:
Also added @class NSURL

PR #165, opened 2020-11-27
Windows fixes

triplef:
This PR makes two changes to fix and improve building on Windows with MinGW-w64 and libobjc2. Posting this as a draft since my Autoconf knowledge is very limited and there are probably issues with my changes or better ways to do this. Feedback welcome.

1. Detect objc.dll and temporarily copy it to source directory.
This fixes the following error during configure, which happens because conftest.exe can’t find objc.dll at runtime. By default Windows executables search for dlls in the working directory, so we simply copy and remove it at the end (but only if configure succeeds – is there a way to have a hook that gets executed at the end?).

```
checking whether objc really works... no
I don't seem to be able to use your Objective-C compiler to produce
working binaries!  Please check your Objective-C compiler installation.
If you are using gcc-3.x make sure that your compiler's libgcc_s and libobjc
can be found by the dynamic linker - usually that requires you to play
with LD_LIBRARY_PATH or /etc/ld.so.conf.
Please refer to your compiler installation instructions for more help.
configure: error: The Objective-C compiler does not work or is not installed properly.
```

2. Prefer `dispatch/dispatch.h` over `dispatch.h`.
Recent libdispatch releases put headers in `dispatch/dispatch.h`, so I think this should be searched first. This also fixes the following noticeable warning with MinGW-w64:

```
configure: WARNING: dispatch.h: present but cannot be compiled
configure: WARNING: dispatch.h:     check for missing prerequisite headers?
configure: WARNING: dispatch.h: see the Autoconf documentation
configure: WARNING: dispatch.h:     section "Present But Cannot Be Compiled"
configure: WARNING: dispatch.h: proceeding with the compiler's result
```

This is because of MinGW-w64 containing the following unrelated `dispatch.h` in `C:\msys64\mingw64\x86_64-w64-mingw32\include`:
```c
/**
 * This file has no copyright assigned and is placed in the Public Domain.
 * This file is part of the mingw-w64 runtime package.
 * No warranty is given; refer to the file DISCLAIMER.PD within this package.
 */
#ifndef RC_INVOKED
#warning Your code should include oleauto.h instead of dispatch.h
#endif

#include <oleauto.h>
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3392,12 +3392,13 @@ if test $enable_libdispatch = yes; then
     GS_ADD_LIBRARY_PATH([${dispatch_libdir}])
   fi
 
-  AC_CHECK_HEADERS(dispatch.h, have_dispatch=yes, have_dispatch=no)
-  if test "$have_dispatch" = "no"; then
-    AC_CHECK_HEADERS(dispatch/dispatch.h, have_dispatch=yes, have_dispatch=no)
+  AC_CHECK_HEADERS(dispatch/dispatch.h, have_dispatch=yes, have_dispatch=no)
+  if test "$have_dispatch" = "yes"; then
     # check for private header which includes runloop integration functions in
     # the Swift corelibs libdispatch release
     AC_CHECK_HEADERS(dispatch/private.h)
+  else
+    AC_CHECK_HEADERS(dispatch.h, have_dispatch=yes, have_dispatch=no)
   fi
   if test "$have_dispatch" = "yes"; then
     AC_CHECK_LIB(dispatch, dispatch_queue_create, have_dispatch=yes, have_dispatch=no)
--- a/configure.ac
+++ b/configure.ac
@@ -1295,7 +1295,7 @@ else
 fi
 
 if test -f "$GNUSTEP_HDIR/objc/objc.h"; then
-  if test -f "$GNUSTEP_LDIR/libobjc.a" -o -f "$GNUSTEP_LDIR/libobjc.so" -o -f "$GNUSTEP_LDIR/libobjc.dll.a" -o -f "$GNUSTEP_LDIR/libobjc-gnu.dylib"; then
+  if test -f "$GNUSTEP_LDIR/libobjc.a" -o -f "$GNUSTEP_LDIR/libobjc.so" -o -f "$GNUSTEP_LDIR/libobjc.dll.a" -o -f "$GNUSTEP_LDIR/libobjc-gnu.dylib" -o -f "$GNUSTEP_LDIR/objc.lib"; then
     gs_cv_objc_libdir="$GNUSTEP_LDIR"
     gs_cv_objc_incdir="$GNUSTEP_HDIR"
   fi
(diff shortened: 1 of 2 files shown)


fredkiefer (commented):
I like most of your changes. The only thing that seems wrong is copying the objc.dll into the current directory. I would expect that installing libobjc2 includes a step where the DLL gets copied into a directory that is on the path. If that is not the case, how would applications later be able to find this DLL at runtime?

triplef:
As far as I understand, DLL search paths on Windows work somewhat differently and putting the DLL next to an executable is a common, simple solution to make Windows find it at runtime. David had once said the following about this on the mailing list:

> Note that, on Windows, the linker does not need to find objc.dll, it needs to find objc.lib.  You don't need objc.dll until you try to run a program (by default, Windows searches in the same folder as the .exe, so to run the tests CMake copies objc.dll into the Tests directory). 

That being said, I’m wondering if there’s a more Autoconf-y instead of manually copying the DLL into the output directory, although I have not found one. The only thing I might change is to add objc.dll to .gitignore and just leave it as part of the output files, as deleting it at the end is brittle anyway (e.g. does not work if configure fails). Thoughts?

triplef:
One other thing I noticed: while make also [has a list of libobjc library names](https://github.com/gnustep/tools-make/blob/528747eac1e53c095194475d100b0095c079cbf3/m4/gs_objc_runtime.m4#L93), it does not list the following that are listed in Base:

`libobjc_gc.a libobjc_gc.so libobjc_gc.dll.a libobjc_gc-gnu.dylib`

Can we assume that these are obsolete then and can be removed in Base as well? Or should we even just copy the `GS_CUSTOM_OBJC_RUNTIME_DOMAIN` function from Make into Base?

triplef:
Ok after reading up on DLL search paths the obvious solution is to add libdir to the path, which works fine and is also [already done in Make](https://github.com/gnustep/tools-make/blob/528747eac1e53c095194475d100b0095c079cbf3/m4/gs_objc_runtime.m4#L226). Sorry for the noise.

With this, the change from multiple `test` calls to the for loop is no longer necessary. Let me know if I should revert it to using the previous style and just add objc.lib there in order to make this change more minimal.

I also changed it to look for objc.lib instead of objc.dll, as objc.lib is what’s needed by the linker. (I will update Make as well to reflect that.)

rfm (approved):
The changes look good.  I took the suggestion of removing the obsolete gc lib from the base configure script.  I'm not sure about copying GS_CUSTOM_OBJC_RUNTIME_DOMAIN ... I don't really understand why we have the current split of obj feature detection between make and base, and wonder if we really need to look into changing that somehow.

configure.ac
```diff
@@ -3551,3 +3567,8 @@ AC_SUBST(GCC_VERSION)
 #--------------------------------------------------------------------
 AC_CONFIG_FILES([config.mak base.make Headers/GNUstepBase/GSConfig.h])
 AC_OUTPUT
+
+# remove copied objc.dll from source directory on Windows
```
fredkiefer, line 3571: This seems to be a leftover. Could you please remove it?

fredkiefer (approved):
Apart from the now obsolete delete statement this change looks good to me.

configure.ac
triplef: Good catch, thanks.

triplef:
Thanks for your feedback. I updated the patch to stick with the old way of checking for the libraries, and also removed the obsolete gc lib when checking in network and local domains which I think were missed in d015cebbf343990f51e7d6662d6c57553bc9089f.

triplef:
I’ve updated the branch to resolve conflicts with the latest changes on master.

I’d appreciate if you guys could take another look at it once more to make sure I didn’t do anything silly, as the patch changed quite a bit (but is much smaller now) from your last review. Thanks!

fredkiefer (approved):
Still good to merge

PR #166, opened 2020-12-15
Disable libcurl support when headers are not found

triplef:
This moves the check for `curl/curl.h` further up so that `HAVE_LIBCURL` is only enabled if both the library and the headers were found.

Fixes cross-compiling in an environment where the host has curl-config, but the target doesn’t have libcurl, e.g. when building for Android on a macOS host.

I’m no expert at Autoconf, so please let me know if there there’s better way to do this and I’m happy do change the patch.


fredkiefer (approved):

rfm (approved):
It seems to make sense.

PR #167, opened 2021-01-23
use built-in strnstr

rmottola:


fredkiefer:
You did see that Richard already fixed this issue even before you started to work on it?

rmottola:
> You did see that Richard already fixed this issue even before you started to work on it?

Actually.... that is incorrect. I wrote him that I found the issue and already a fix, but I forgot to actually create the pull-request. My solution is alternative, but already tested. Up to him to see and check. He can just discard the PR.

PR #168, opened 2021-01-29
Support building on Windows with Clang MSVC target

triplef:
Added support for building on Windows with Clang MSVC target and libobjc2. This requires using a standard (non-MinGW) Clang build that e.g. comes with Visual Studio or is available as pre-built binary from the LLVM website, and requires passing a host to configure like `--host=x86_64-pc-windows`.

The build is best done in an MSYS2 shell that does not have any additional *-devel packages installed that might get picked up by configure. Alternatively `--disable-xxx` flags can be used to prevent these dependencies to be picked up.

Annotates all ObjC class interfaces with `GS_EXPORT_CLASS` for DLL export/import declarations, and adds missing `GS_DECLARE` annotations in externs.m, in order for these symbols to be correctly exported in the DLL.

I’ll send more info on the mailing list.

Requires https://github.com/gnustep/tools-make/pull/14.


rfm (approved):
This all makes sense t me, and looks straightforward, so I don't see any problems.
I have now had a chance to try it out on various platforms, just to make sure nothing has been broken.

triplef:
Thanks @rfm! I’ll rebase the branch and delete the temporary CI change commit after merging the Make PR and before merging this one. 👍

gcasa:
While I understand and approve of these changes I am wondering why it isn't possible to change the build of libobjc2 to simply work on MinGW or MSYS2?

triplef:
MinGW support is also underway in https://github.com/gnustep/libobjc2/pull/190, but it requires fixes in Clang and I haven’t figured out yet how to build Clang with the MinGW patches in a way that David can debug it in Visual Studio.

Hopefully we can support both going forward, as they each have their pros and cons:

**MSVC version**
- Pros: can be used with the standard Visual Studio toolchains and tooling, and with any other library that’s built with MSVC.
- Cons: more involved to build dependencies, as they are not readily available as packages, and each require their own special way to build with MSVC. Hopefully we can remedy this with a set of good scripts akin to [tools-android](https://github.com/gnustep/tools-android).

**MinGW version**
- Pros: should be easy to build, with dependencies available as packages.
- Cons: can only be used with a MinGW toolchain, e.g. all other libraries must also be built with MinGW.

For us, the MSVC version is much more attractive for the above reasons, and it should combine nicely with projects such as [C++/WinRT](https://github.com/microsoft/cppwinrt) that are not compatible with MinGW.

PR #175, opened 2021-03-12
Windows MSVC fixes

triplef:
- Add NSFileManager support on Windows MSVC using `dirent.h` from https://github.com/tronkko/dirent/blob/master/include/dirent.h (MIT licensed).
- Unifies _CHAR in NSFileManager with GSNativeChar, as they are internally required to be the same.
- Fix libdispatch integration compilation error on Windows because `ET_RDESC` is undefined.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2021-03-09 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSRunLoop.m:
+	Fix libdispatch integration compilation error on Windows.
+
 2021-03-05 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Source/GSHTTPURLHandle.m:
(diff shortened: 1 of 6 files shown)


triplef:
FYI if anyone wants to try this out I’ve been working on a [set of scripts](https://github.com/triplef/tools-windows-msvc) (similar to the ones for Android) to build a toolchain using the MSVC ABI.

fredkiefer:
Looks good to me. Would it be possible to add MSVC to the Travis CI as well?

triplef:
Yeah good idea to add MSVC to CI. It would probably require using the [MSVC toolchain scripts](https://github.com/gnustep/tools-windows-msvc) to build all the dependencies, as some of them require a lot of fiddling with build parameters etc. and one needs to jump between CMD and Bash, but once those are installed the existing part of the CI script should be able to build Base and the tests normally. I’ll see what I can do there.

PR #176, opened 2021-03-22
Add missing GS_EXPORT_CLASS annotations.

triplef:
Looks like I missed these.

### DIFF
--- a/Headers/Foundation/NSLock.h
+++ b/Headers/Foundation/NSLock.h
@@ -121,6 +121,7 @@ GS_EXPORT_CLASS
 /**
  * NSCondition provides an interface to POSIX condition variables.
  */
+GS_EXPORT_CLASS
 @interface NSCondition : NSObject <NSLocking>
 {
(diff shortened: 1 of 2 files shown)


triplef:
I took the liberty to merge this without review, hope that’s ok. The failing tests were the same as before (NSFileHandle/singleton).

rfm:
> On 23 Mar 2021, at 08:52, Frederik Seiffert ***@***.***> wrote:
> 
> 
> I took the liberty to merge this without review, hope that’s ok. The failing tests were the same as before (NSFileHandle/singleton).

I had meant to approve it ... but got distracted looking at the NSFileHandle thing.
I read the code there, and it's quite clear that the use of -retain in -dealloc was intended to ensure that these objects, which should never be deallocated, are brought back to having a retain count.
But, given that it's not an important feature, perhaps we can drop it and change the tests.
I actually feel the libobjc bug where it is printing spurious information for exceptions is more worrying.

PR #177, opened 2021-03-26
Add missing GS_EXPORT(_CLASS) annotations

triplef:
Somehow I had missed a couple more... I hope this should be it tho.

Note that we’re currently not exporting the GS* classes, not sure if we should?

I did add `GS_EXPORT` to the `GSTLS*` string constants, as they are also already defined like this in NSFileHandle.h and the definitions should match.


triplef:
I also added a change that should fix the NSFileHandle singleton tests as discussed on the mailing lists and in https://github.com/gnustep/libs-base/pull/176#issuecomment-804738429.

rfm (approved):

PR #178, opened 2021-03-26
Export objc_enumerationMutation() in public header

triplef:
Fixes linker error on Windows, as I didn’t realize that the previous declaration in GSFastEnumeration.h was not public, and therefore didn't expose the dllspec attributes.

### DIFF
--- a/Headers/Foundation/NSEnumerator.h
+++ b/Headers/Foundation/NSEnumerator.h
@@ -57,6 +57,11 @@ GS_EXPORT_CLASS
 - (GS_GENERIC_TYPE(IterT)) nextObject;
 @end
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wattributes"
+GS_EXPORT void objc_enumerationMutation(id);
+#pragma GCC diagnostic pop
+
 #if	defined(__cplusplus)
 }
 #endif
--- a/Source/GSFastEnumeration.h
+++ b/Source/GSFastEnumeration.h
@@ -1,9 +1,4 @@
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wattributes"
-GS_EXPORT void objc_enumerationMutation(id);
-#pragma GCC diagnostic pop
-
 #ifdef __clang__
 #define FOR_IN(type, var, collection) \
   for (type var in collection)\
--- a/Source/NSEnumerator.m
+++ b/Source/NSEnumerator.m
@@ -102,6 +102,7 @@
  * objc_enumerationMutation() is called whenever a collection mutates in the
  * middle of fast enumeration.
  */
+GS_DECLARE
 void objc_enumerationMutation(id obj)
 {
   [NSException raise: NSGenericException 


fredkiefer:
Things here are a bit more complicated. The function objc_enumerationMutation actually belongs into libobjc2 and is present there as a weak symbol. It should get added in base for old versions of libobjc, where this is missing and to override the version from libobjc2 to throw an exception instead of just aborting the program.
I am not sure on how to best express all of this with C declarations. To me what you are doing here and in the last change looks correct. I just wanted to bring up the background to make sure this is taken care of.

rfm (commented):
This is not part of the public API, so ideally shouldn't be in NSEnumerator.h (though I don't think it's terribly harmful).
Would it make sense to leave the declaration where it is and explicitly use extern __declspec(dllexport) on windows, rather than using GS_EXPORT ?

triplef:
> Things here are a bit more complicated. The function objc_enumerationMutation actually belongs into libobjc2 and is present there as a weak symbol. It should get added in base for old versions of libobjc, where this is missing and to override the version from libobjc2 to throw an exception instead of just aborting the program.
> I am not sure on how to best express all of this with C declarations. To me what you are doing here and in the last change looks correct. I just wanted to bring up the background to make sure this is taken care of.

You’re right we were also missing a `GS_DECLARE` annotation on the implementation. I tested the latest version on Windows MSVC and it’s now correctly being found.

> This is not part of the public API, so ideally shouldn't be in NSEnumerator.h (though I don't think it's terribly harmful).
> Would it make sense to leave the declaration where it is and explicitly use extern __declspec(dllexport) on windows, rather than using GS_EXPORT ?

The problem is that `GSFastEnumeration.h` is not imported as part of `Foundation.h`, but the compiler must be able to see the declaration when compiling code that is referencing this function (i.e. whenever using fast enumeration with Clang). That’s why it has to be in a public header, as importing Foundation should be sufficient to use fast enumeration.

fredkiefer:
> The problem is that `GSFastEnumeration.h` is not imported as part of `Foundation.h`, but the compiler must be able to see the declaration when compiling code that is referencing this function (i.e. whenever using fast enumeration with Clang). That’s why it has to be in a public header, as importing Foundation should be sufficient to use fast enumeration.

This last bit worries me somewhat. This functions should be exported by libobjc2, but it isn't. The companion function objc_setEnumerationMutationHandler is missing there as well. Instead of trying to work around this in base, we should find a suitable way forward together with @davidchisnall. Your PR may still be a short-term workaround to get things usable on Windows.

triplef:
Ah that’s a good point, you’re absolutely right. I think we’re just missing an `OBJC_PUBLIC` attribute here:
https://github.com/gnustep/libobjc2/blob/master/mutation.m#L7

I’ll give that a try.

triplef:
This is now fixed in the runtime via https://github.com/gnustep/libobjc2/pull/202. Thanks for thinking this through Fred!

I’ll close this and also revert https://github.com/gnustep/libs-base/commit/c52f1e3223d5789b3f677bfb7a6a7e958436e060.

PR #179, opened 2021-03-26
Export string constants and annotate exported function implementations

triplef:
- adds missing exception name declarations in `NSConnection.h`
- adds missing `GS_DECLARE` annotations to many public C function implementations whose declarations are annotated with `GS_EXPORT`
- moves string constants to `externs.m` and annotates them with `GS_DECLARE`
- changes string constants in `externs.m` to `NSString* const` so they match their declarations

Unfortunately the last point causes "assignment of read-only variable" errors on CI (somehow with GCC and libobjc 1.9 runtime only). However without this change I was getting errors like this:
```
externs.m:335:22: error: redefinition of 'NSBundleDidLoadNotification' with a different type: 'NSString *' vs
      'NSString *const'
GS_DECLARE NSString* NSBundleDidLoadNotification = @"NSBundleDidLoadNotification";
                     ^
../Headers\Foundation/NSBundle.h:65:27: note: previous declaration is here
GS_EXPORT NSString* const NSBundleDidLoadNotification;
```

I think this mismatch between using const in the headers and non-const in externs.m worked for the existing strings because none of their const declarations were visible to the compiler, as externs.m is only importing few headers. But now I think we can’t avoid indirectly importing some of the headers declaring the same strings defined in externs.m, so I’m not sure how to resolve this.

The reason the strings in externs.m are declared writable is that they get replaced in `GSPrivateBuildStrings()` at runtime, supposedly so they can cache their hash values. This code seems to be about 20 years old and is only implemented for a few seemingly random string constants, and I’m not sure if the optimization is still necessary today. Removing this code would solve the above issue.

I’d appreciate any thoughts on the above.


rfm:
> On 26 Mar 2021, at 22:09, Frederik Seiffert ***@***.***> wrote:
> 
> 	• adds missing exception name declarations in NSConnection.h
> 	• adds missing GS_DECLARE annotations to many public C function implementations whose declarations are annotated with GS_EXPORT
> 	• moves string constants to externs.m and annotates them with GS_DECLARE
> 	• changes string constants in externs.m to NSString* const so they match their declarations
> Unfortunately the last point causes "assignment of read-only variable" errors on CI (somehow with GCC and libobjc 1.9 runtime only). However without this change I was getting errors like this:
> 
> externs.m:335:22: error: redefinition of 'NSBundleDidLoadNotification' with a different type: 'NSString *' vs
>       'NSString *const'
> GS_DECLARE NSString* NSBundleDidLoadNotification = @"NSBundleDidLoadNotification";
>                      ^
> ../Headers\Foundation/NSBundle.h:65:27: note: previous declaration is here
> GS_EXPORT NSString* const NSBundleDidLoadNotification;
> 
> I think this mismatch between using const in the headers and non-const in externs.m worked for the existing strings because none of their const declarations were visible to the compiler, as externs.m is only importing few headers. But now I think we can’t avoid indirectly importing some of the headers declaring the same strings defined in externs.m, so I’m not sure how to resolve this.
> 
> The reason the strings in externs.m are declared writable is that they get replaced in GSPrivateBuildStrings() at runtime, supposedly so they can cache their hash values. This code seems to be about 20 years old and is only implemented for a few seemingly random string constants, and I’m not sure if the optimization is still necessary today. Removing this code would solve the above issue.


You are right, the code dates from long ago, when that optimisation was important.
It has been irrelevant for many years now (so none of the newer constants use it).
We should remove the obsolete code.

triplef:
> You are right, the code dates from long ago, when that optimisation was important. It has been irrelevant for many years now (so none of the newer constants use it). We should remove the obsolete code.

Great, I’ve updated the patch to remove `GSPrivateBuildStrings()`.

rfm (approved):

PR #180, opened 2021-03-28
Add support for libobjc2 runtime in NSObject on Windows

triplef:
Windows doesn’t support resolving weak symbols, so we need to guard the retain/release libobjc2 support on `__GNUSTEP_RUNTIME__` instead of `SUPPORT_WEAK`. And we also need a `__declspec(dllimport)` annotation on the libobjc2 functions so they are resolved from the DLL.

The new `GS_IMPORT` macro should be added in other places too where we declare externally defined functions.


triplef:
I now also removed the obsolete `SUPPORT_WEAK` definition.

rfm (approved):

PR #184, opened 2021-06-02
Fix compilation without GNUTLS

triplef:
e42d9fdc6ba1a632ff16741fe9a0f919205e2007 recently broke building without GNUTLS, because it unconditionally references `GSTLSSession`, which [GSTLS.h only defines with `HAVE_GNUTLS`](https://github.com/gnustep/libs-base/blob/master/Source/GSTLS.m#L76).

To fix this I’ve added guards around any usage of GSTLSSession in NSSocketPort.m.

### DIFF
--- a/Source/NSSocketPort.m
+++ b/Source/NSSocketPort.m
@@ -243,7 +243,9 @@ typedef enum {
   NSSocketPort		*sendPort;
   struct sockaddr	sockAddr;	/* Far end of connection.	*/
   NSString		*defaultAddress;
+#if     defined(HAVE_GNUTLS)
   GSTLSSession		*session;	/* Session for encryption.	*/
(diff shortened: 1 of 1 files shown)


triplef:
Thanks, looks like this was fixed in 2fc16eae5eeb8cccb858d1ef408d4c51e5b6b70d.

rfm:
Thanks.  By the time I saw this I'd already changed stuff somewhat (so the patch wouldn't apply), but your patch prompted me to fix a few things.
FYI this change is still a work in progress:  while encrypted DO seems to work fine in my simple testcases, trying in on real-world applications with a lot of cooperating DO servers is not working properly (messages not getting to their destinations), so there are clearly bugs to fix.

PR #185, opened 2021-06-17
Fix lookup of string encodings with high numeric values in encoding table

triplef:
Using Clang MSVC on 64-bit Windows was interpreting the constants in the `NSStringEncoding` enum as `int` (which is [according to the C standard](https://stackoverflow.com/q/366017/1534401)), which meant that the `NSUTF16*EndianStringEncoding` and `NSUTF32*StringEncoding` were interpreted as negative numbers due to their topmost bit being set, which caused the `EntryForEncoding()` function to always fail for these constants.

I’m not sure in how far this affects other compilers/platforms, although it seems like it might affect all.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2021-06-17 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/Additions/Unicode.m:
+	Fix lookup of string encodings with high numeric values in
+	encoding table when they are interpreted as negative numbers
+	for some platforms/compilers (pertains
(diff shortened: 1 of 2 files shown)


rfm:
Thanks ... fix looked good, so I merged.

triplef:
Interestingly some compilers seem to consider `NSStringEncoding` an unsigned type and happily go into the `if (enc > 0)` branch for one of the above mentioned string encodings. I’ve verified this behavior on Android x86 and arm64 (both using Clang).

Not sure if the signed type is a specific behavior of Clang in MSVC mode, but the change should still make the encoding lookup more robust regarding these different behaviors.

PR #187, opened 2021-06-29
Rename first argument to NSDebug[MF]LLog to make it easier to understand

iamleeg:
The argument was named "level" in the docs but it's preferable if it's called "key" because then you can see the parallels with `NSDebugLLog`.


fredkiefer (approved):
This surely is an improvement but wouldn't it be even better to also change the name of the argument in NSDebug+GNUstepBase.h?

iamleeg:
Good point, we only looked in the documentation and not in the header. I'll send that PR too!

PR #188, opened 2021-06-30
Reword NSDebug*LLog "level" argument to "key"

iamleeg:
The motivation is that these are more like keys to specific logging domain names, not log levels. If you think of log levels, you usually think of the syslog style debug/info/warn/error levels where enabling logs at any one level implies enabling those at all levels above. Our log keys aren't ordered in that way. The texinfo docs were inconsistent in using "level" and "key" as the argument and we already changed that to use "key" throughout" this change makes the header and auto docs consistent.


rfm:
yes, good change.

PR #189, opened 2021-07-02
Fix possible heap corruption when converting to NSNonLossyASCIIStringEncoding

triplef:
Possible heap corruption in `GSFromUnicode()` because `sprintf()` always appends NULL terminator, which is written past the end of the `ptr` buffer for the last character in the string.

Was exposed by existing `NSString/basic` test when checking for "latin1 in lossy encoding", as the last character of that string is running into the affected code. When running tests using the Windows MSVC toolchain this was showing a "HEAP CORRUPTION DETECTED" error.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2021-07-02 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/Additions/Unicode.m:
+	Fix possible heap corruption when converting to
+	NSNonLossyASCIIStringEncoding.
+
 2021-06-19 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Headers/Foundation/Port.h:
--- a/Source/Additions/Unicode.m
+++ b/Source/Additions/Unicode.m
@@ -2242,12 +2242,22 @@ GSFromUnicode(unsigned char **dst, unsigned int *size, const unichar *src,
                         }
                       else
                         {
-                          dpos += sprintf((char*)&ptr[dpos], "\\%03o", u);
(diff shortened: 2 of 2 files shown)


rfm (changes requested):
As I understand it, the problem is that if we need an ocat or hex escape sequence for the last character in the string, the nul terminator of the sprintf will overrun the buffer.

Unless I miss something about snprintf, replacing each call to sprintf with one to snprintf should fix that (no need to inefficiently write into a separate buffer and then copy back).

But with efficiency in mind, we could go better than that and generate the octal or hex sequence directly.
eg
ptr[dpos + 3] = "01234567"[u / 8];
u /= 8;
ptr[dpos + 2] = "01234567"[u / 8];
...

Of course that's a lot more lines of code and a chance to make mistakes, so perhaps just replacing sprintf with snprintf is best?

What do you think?

rfm:
Given that I managed to make an error (u / 8 when I meant u % 8) in that snippet of code, snprintf seems the safer option.

triplef:
If I read the man page correctly snprintf() always adds a nul terminator, so I don't think there's a way to use either sprintf() or snprintf() without a separate buffer.

It does seem not so great performance wise though. I'll see if I can come up with a concise way to build these strings manually.

triplef:
How is this?

rfm (approved):
That looks fne.
You are right about snprintf (its behavior just seems really unintuitive to me): I actually had to write a test program to convince myself about that reading of the documentation!

triplef:
Yeah same, had to write a test program too...

PR #190, opened 2021-07-08
Add support for native locks and threading on Windows

triplef:
This PR adds support for native locks on Windows, as previously suggested by David on the mailing list. I subsequently plan to also add native threading support in NSThread and thereby remove the dependency on the pthread library on Windows.

While the second part is still outstanding, I wanted to get feedback on the changes early on to ensure this is going in the right direction, and I’d appreciate your initial feedback @rfm and @davidchisnall.

The support for native locks uses the following Windows APIs:

- [Slim Reader/Writer (SRW) locks](https://docs.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks) in exclusive mode for any non-recursive locks, which should be the fastest kind of locks available on Windows from what I’ve read
- [Critical Section objects](https://docs.microsoft.com/en-us/windows/win32/sync/critical-section-objects) for recursive locks
- [Condition Variables](https://docs.microsoft.com/en-us/windows/win32/sync/condition-variables)

One issue with using these APIs is that SRWLocks and Critical Sections don’t seem to support a way to detect deadlocks akin to the `PTHREAD_MUTEX_ERRORCHECK` attribute. This means that with this change, `NSLock` and `NSCondition` will no longer be able to detect deadlocks and throw exceptions when they occur. Please let me know if you can think of a way to work around this, or whether this would still be an acceptable change.

To implement these APIs I’ve defined the following new types and a number of defines for initializing/locking/unlocking/signalling in `GSPThread.h`, which cover all "simple" locking cases throughout the code base outside the more specific needs of NSLock:

- `gs_mutex_t`: defined as `SRWLOCK` on Windows, otherwise `pthread_mutex_t`
- `gs_recursive_mutex_t`: defined as `CRITICAL_SECTION` on Windows, otherwise `pthread_mutex_t`
- `gs_cond_t`: defined as `CONDITION_VARIABLE` on Windows, otherwise `pthread_cond_t`

_(I’ve renamed the existing types `gs_cond_t` and `gs_mutex_t` in `GSConfig.h` to `gs_cond_ivar_t` and `gs_mutex_ivar_t`, and added a 3rd type `gs_recursive_mutex_ivar_t` there as well. Let me know if a different naming would be preferable.)_

Other than the missing deadlock detection, the change seems to work well in my limited testing so far, but I plan to test it more extensively with MinGW and MSVC.

### DIFF
--- a/Source/GSFFCallInvocation.m
+++ b/Source/GSFFCallInvocation.m
@@ -26,6 +26,7 @@
 #import "Foundation/NSCoder.h"
 #import "Foundation/NSDistantObject.h"
 #import "GSInvocation.h"
+#import "GSPThread.h"
 #import <avcall.h>
 #import <callback.h>
 #import "callframe.h"
@@ -33,10 +34,6 @@
 #if !defined (__GNU_LIBOBJC__)
 #  include <objc/encoding.h>
 #endif
-
-#include <pthread.h>
-
-#import "GSInvocation.h"
 
 #ifndef GS_STATIC_INLINE
 #define GS_STATIC_INLINE static inline
@@ -140,7 +137,7 @@ static GSIMapTable_t ff_callback_map;
 
 /* Lock that protects the ff_callback_map */
 
-static pthread_mutex_t ff_callback_map_lock = PTHREAD_MUTEX_INITIALIZER;
+static gs_mutex_t ff_callback_map_lock = GS_MUTEX_INIT_STATIC;
 
 /* Static pre-computed return type info */
 
@@ -486,7 +483,7 @@ static IMP gs_objc_msg_forward (SEL sel)
       GSIMapNode node;
 
       // Lock
-      pthread_mutex_lock (&ff_callback_map_lock);
+      GS_MUTEX_LOCK(ff_callback_map_lock);
 
       node = GSIMapNodeForKey (&ff_callback_map,
 	(GSIMapKey) ((void *) &returnInfo));
@@ -512,7 +509,7 @@ static IMP gs_objc_msg_forward (SEL sel)
 	    (GSIMapVal) forwarding_callback);
 	}
       // Unlock
-      pthread_mutex_unlock (&ff_callback_map_lock);
+      GS_MUTEX_UNLOCK(ff_callback_map_lock);
     }
   return forwarding_callback;
 }
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,45 @@
 	Changes to add support for 64bit transitions used by v2+ of the
 	timezone format.
 
+2021-07-14 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Headers/Foundation/NSLock.h:
+	* Headers/Foundation/NSThread.h:
+	* Headers/GNUstepBase/GSConfig.h.in:
+	* Headers/GNUstepBase/config.h.in:
+	* Source/Additions/GCObject.m:
+	* Source/Additions/GSObjCRuntime.m:
+	* Source/Additions/Unicode.m:
+	* Source/GSAtomic.h:
+	* Source/GSFFCallInvocation.m:
+	* Source/GSFFIInvocation.m:
+	* Source/GSPThread.h:
+	* Source/NSArray.m:
+	* Source/NSCharacterSet.m:
+	* Source/NSDebug.m:
+	* Source/NSException.m:
+	* Source/NSLock.m:
+	* Source/NSObject.m:
+	* Source/NSScanner.m:
+	* Source/NSString.m:
+	* Source/NSThread.m:
+	* Source/NSTimeZone.m:
+	* Source/NSZone.m:
+	* Tests/base/NSThread/GNUmakefile.preamble:
+	* Tests/base/NSThread/late_unregister.m:
+	* Tests/base/NSThread/lazy_thread.m:
+	* config/config.initialize.m:
+	* configure:
+	* configure.ac:
+	Use native threading and locking APIs on Windows, removing dependency
+	on pthread library and using fast Slim Reader/Writer (SRW) locks for
+	NSLock/NSRecursiveLock/NSCondition/NSConditionLock as well as all
+	internal locks. Adds GS_MUTEX_*() macros in GSPThread.h, that are being
+	used for all internal locking instead of pthread APIs.
+	Also adds support for thread priorities on Windows, fixes method
+	signature of +[NSThread setThreadPriority:] to match Apple platforms,
+	and adds error handling in same method.
+
 2021-07-16 Frederik Seiffert <frederik@algoriddim.com>
 
 	* Source/GSICUString.h:
--- a/Headers/Foundation/NSLock.h
+++ b/Headers/Foundation/NSLock.h
@@ -74,7 +74,7 @@ GS_EXPORT_CLASS
 {
 #if	GS_EXPOSE(NSLock)
 @protected
-  gs_mutex_t	_mutex;
+  gs_mutex_public_t	_mutex;
   NSString	*_name;
 #endif
 }
@@ -126,8 +126,8 @@ GS_EXPORT_CLASS
 {
 #if	GS_EXPOSE(NSCondition)
 @protected
-  gs_cond_t	_condition;
-  gs_mutex_t	_mutex;
+  gs_cond_public_t	_condition;
+  gs_mutex_public_t	_mutex;
   NSString	*_name;
 #endif
 }
@@ -278,7 +278,7 @@ GS_EXPORT_CLASS
 {
 #if	GS_EXPOSE(NSRecursiveLock)
 @protected
-  gs_mutex_t	_mutex;
+  gs_mutex_public_t	_mutex;
   NSString      *_name;
 #endif
 }
--- a/Headers/Foundation/NSThread.h
+++ b/Headers/Foundation/NSThread.h
@@ -134,7 +134,7 @@ GS_EXPORT_CLASS
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_2,GS_API_LATEST) \
   && GS_API_VERSION( 10200,GS_API_LATEST)
-+ (void) setThreadPriority: (double)pri;
++ (BOOL) setThreadPriority: (double)pri;
 + (double) threadPriority;
 #endif
 
--- a/Headers/GNUstepBase/GSConfig.h.in
+++ b/Headers/GNUstepBase/GSConfig.h.in
@@ -254,10 +254,13 @@ typedef	gsuaddr gsaddr;
  */
 typedef	struct {
   uint8_t	dummy[@GS_SIZEOF_COND_T@];
-} gs_cond_t	__attribute__((aligned (@GS_ALIGNOF_COND_T@)));
+} gs_cond_public_t	__attribute__((aligned (@GS_ALIGNOF_COND_T@)));
+typedef	struct {
+  uint8_t	dummy[@GS_SIZEOF_COND_MUTEX_T@];
+} gs_cond_mutex_public_t	__attribute__((aligned (@GS_ALIGNOF_COND_MUTEX_T@)));
 typedef	struct {
   uint8_t	dummy[@GS_SIZEOF_MUTEX_T@];
-} gs_mutex_t	__attribute__((aligned (@GS_ALIGNOF_MUTEX_T@)));
+} gs_mutex_public_t	__attribute__((aligned (@GS_ALIGNOF_MUTEX_T@)));
 
 #define	OBJC2RUNTIME @OBJC2RUNTIME@
 #define BASE_NATIVE_OBJC_EXCEPTIONS     @BASE_NATIVE_OBJC_EXCEPTIONS@
@@ -272,6 +275,7 @@ typedef	struct {
 #define GS_USE_LIBDISPATCH_RUNLOOP @HAVE_LIBDISPATCH_RUNLOOP@
 #define GS_HAVE_NSURLSESSION @GS_HAVE_NSURLSESSION@
 #define GS_HAVE_OBJC_ROOT_CLASS_ATTR @GS_HAVE_OBJC_ROOT_CLASS_ATTR@
+#define GS_USE_WIN32_THREADS_AND_LOCKS @HAVE_WIN32_THREADS_AND_LOCKS@
 
 #ifndef __has_include
 #  define __has_include(x) 0
--- a/Headers/GNUstepBase/config.h.in
+++ b/Headers/GNUstepBase/config.h.in
@@ -3,11 +3,20 @@
 /* Define if building universal (internal helper macro) */
 #undef AC_APPLE_UNIVERSAL_BUILD
 
+/* The normal alignment of `CONDITION_VARIABLE', in bytes. */
+#undef ALIGNOF_CONDITION_VARIABLE
+
+/* The normal alignment of `gs_mutex_t', in bytes. */
+#undef ALIGNOF_GS_MUTEX_T
+
 /* The normal alignment of `pthread_cond_t', in bytes. */
 #undef ALIGNOF_PTHREAD_COND_T
 
 /* The normal alignment of `pthread_mutex_t', in bytes. */
 #undef ALIGNOF_PTHREAD_MUTEX_T
+
+/* The normal alignment of `SRWLOCK', in bytes. */
+#undef ALIGNOF_SRWLOCK
 
 /* Define if SO_REUSEADDR is broken on this system */
 #undef BROKEN_SO_REUSEADDR
@@ -861,11 +870,17 @@
 /* Define to 1 if the `setpgrp' function takes no argument. */
 #undef SETPGRP_VOID
 
+/* The size of `CONDITION_VARIABLE', as computed by sizeof. */
+#undef SIZEOF_CONDITION_VARIABLE
+
 /* The size of `double', as computed by sizeof. */
 #undef SIZEOF_DOUBLE
 
 /* The size of `float', as computed by sizeof. */
 #undef SIZEOF_FLOAT
+
+/* The size of `gs_mutex_t', as computed by sizeof. */
+#undef SIZEOF_GS_MUTEX_T
 
 /* The size of `int', as computed by sizeof. */
 #undef SIZEOF_INT
@@ -884,6 +899,9 @@
 
 /* The size of `short', as computed by sizeof. */
 #undef SIZEOF_SHORT
+
+/* The size of `SRWLOCK', as computed by sizeof. */
+#undef SIZEOF_SRWLOCK
 
 /* The size of `void*', as computed by sizeof. */
 #undef SIZEOF_VOIDP
--- a/Source/Additions/GCObject.m
+++ b/Source/Additions/GCObject.m
@@ -37,7 +37,7 @@
 
 #import "GNUstepBase/GCObject.h"
 
-#include <pthread.h>
+#import "../GSPThread.h"
 
 /*
  * The head of a linked list of all garbage collecting objects  is a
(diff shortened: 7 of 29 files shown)


davidchisnall:
You can implement a recursive mutex as a struct containing a SRWLOCK and a thread ID and a recursion count.  Something like this:

```
struct RecursiveLock
{
  SRWLOCK lock;
  DWORD owner;
  uint64_t depth;
};
```

Lock acquires `.lock` and stores `GetCurrentThreadId()` in owner.  If acquiring `.lock` fails then do an atomic load of `.owner` and compare it to the result of `GetCurrentThreadId()`.  If it is the same, lock succeeded.  If you succeeded in either path, increment `.depth`.

Unlock decrements `.depth`.  If the result is 0, then it sets `.owner` to some invalid value (0 or -1?) and releases `.lock`.

Once you have this primitive, you can implement error-checked locks very easily: check `.owner` before unlocking and return an error.

triplef:
Thanks David, that is very helpful input and seems manageable.

I even found an implementation that does almost exactly what you are describing:
https://github.com/libsdl-org/SDL/commit/548cb9089353d07323bb079976524562c1a5c8ed

One difference here is that they lock before setting the owner instead of doing an atomic load, which might be slightly simpler. What are your thoughts on that, should that work just as well?


An additional point that I forgot to mention is that SRW locks require Windows Vista. @rfm would that be ok to not support 2000/XP, or should this be conditional somehow and still allow using pthreads?

davidchisnall:
Sorry, I don't think I was clear.  Pseudocode for what I was suggesting:

```c++
void lock(RecursiveLock *l)
{
  if (try_lock(l->lock))
  {
    assert(l->depth == 0);
    l->depth = 1;
    return;
  }
  // Needs to be atomic because another thread can concurrently set it.  
  auto owner = atomic_load(&l->owner);
  if (owner == my_thread_id)
  {
    l->depth++; // Doesn't need to be atomic, protected by lock
    return;
  }
  lock(&l->lock);
  assert(l->depth == 0);
  l->depth = 1;
  // Must be visible to other threads, data races are UB in C/C++11.
  atomic_store(&l->owner, my_thread_id);  
}
```

I think line 118 in the SDL implementation is undefined behaviour in C/C++.  If the lock is held, then it's a non-racy access and it's well defined, but if the lock is not held then this is a non-atomic access that can happen concurrently with a the store from another thread.  In general, this is undefined behaviour (C programs must be data-race free to be well defined) and so the implementation is free to do pretty much anything at this point.

triplef:
So I spent a bit more time trying to figure out what this could look like using David’s suggestion. I think ideally we’d be able to use this to replicate pthread’s behavior for recursive and non-recursive locks, and also support error checking ([PTHREAD_MUTEX_ERRORCHECK](https://linux.die.net/man/3/pthread_mutexattr_settype)) as that’s used in NSLock to throw exceptions instead of deadlocking:

> This type of mutex provides error checking. A thread attempting to relock this mutex without first unlocking it shall return with an error. A thread attempting to unlock a mutex which another thread has locked shall return with an error. A thread attempting to unlock an unlocked mutex shall return with an error.

Recursive locks (PTHREAD_MUTEX_RECURSIVE) are documented to have the same error checking, whereas the default behavior (PTHREAD_MUTEX_DEFAULT, which is not used in NSLock, but is used for all the locks protecting state inside GNUstep) is to exhibit undefined behavior in error cases. Finally there’s the `PTHREAD_MUTEX_NORMAL` type, which deadlocks if trying to take a lock recursively.

With the above in mind and using David’s template I sketched out the following implementations for lock/trylock/unlock (completely untested), adding another field `attr` to the lock struct for the above flags (normal/errorcheck/recursive, I left out "default" since I think we can default to the "normal" behavior instead).

@davidchisnall I’d appreciate if you could look this over to see if there’s anything you would change or you can spot any obvious errors. The unlock implementation I’m most unsure about, as it I think there could be a data race for error checking locks if another thread sets the owner right after we do the owner check. Not sure if there’s a way to fix this? Thanks for your help!

```c
typedef enum {
  gs_mutex_attr_default = 0,
  gs_mutex_attr_errorcheck = 1,
  gs_mutex_attr_recursive = 2
} gs_mutex_attr_t;

typedef struct
{
  SRWLOCK lock;
  DWORD owner;
  DWORD depth;
  gs_mutex_attr_t attr;
} gs_mutex_t;

#define GS_MUTEX_INIT_DEFAULT    {SRWLOCK_INIT, 0, 0, gs_mutex_attr_default}
#define GS_MUTEX_INIT_ERRORCHECK {SRWLOCK_INIT, 0, 0, gs_mutex_attr_errorcheck}
#define GS_MUTEX_INIT_RECURSIVE  {SRWLOCK_INIT, 0, 0, gs_mutex_attr_recursive}

int gs_mutex_lock(gs_mutex_t *l)
{
  DWORD thisThread = GetCurrentThreadId();
  
  // fast path if lock is not taken
  if (TryAcquireSRWLockExclusive(l->lock))
  {
    assert(l->depth == 0);
    l->depth = 1;
    atomic_store(&l->owner, thisThread);
    return 0;
  }
  
  // Needs to be atomic because another thread can concurrently set it.
  DWORD ownerThread = atomic_load(&l->owner);
  if (ownerThread == thisThread)
  {
    // this thread already owns this lock
    switch (l->attr) {
      case gs_mutex_attr_default:
        // deadlock
        assert(l->depth == 1);
        AcquireSRWLockExclusive(&l->lock);
        assert(false); // not reached
        return 0;
        
      case gs_mutex_attr_errorcheck:
        // return deadlock error
        assert(l->depth == 1);
        return EDEADLK;
      
      case gs_mutex_attr_recursive:
        // recursive lock
        l->depth++;
        return 0;
    }
  }
  else
  {
    // wait for another thread to release the lock
    AcquireSRWLockExclusive(&l->lock);
    assert(l->depth == 0);
    l->depth = 1;
    atomic_store(&l->owner, thisThread);
    return 0;
  }
}

int gs_mutex_trylock(gs_mutex_t *l)
{
  DWORD thisThread = GetCurrentThreadId();
  
  if (TryAcquireSRWLockExclusive(l->lock))
  {
    assert(l->depth == 0);
    l->depth = 1;
    atomic_store(&l->owner, thisThread);
    return 0;
  }
  
  // Needs to be atomic because another thread can concurrently set it.
  DWORD ownerThread = atomic_load(&l->owner);
  if (ownerThread == thisThread && l->attr == gs_mutex_attr_recursive)
  {
    // this thread already owns this lock and it's recursive
    assert(l->depth > 0);
    l->depth++;
    return 0;
  }
  else
  {
    // lock is taken
    return EBUSY;
  }
}

int gs_mutex_unlock(gs_mutex_t *l)
{
  DWORD thisThread = GetCurrentThreadId();
  DWORD ownerThread = atomic_load(&l->owner);
  
  switch (l->attr) {
    case gs_mutex_attr_default:
      break;
    case gs_mutex_attr_errorcheck:
    case gs_mutex_attr_recursive:
      // return error if lock is not held by this thread
      if (ownerThread != thisThread) {
        return EPERM;
      }
      break;
  }
  
  if (l->attr == gs_mutex_attr_recursive && l->depth > 1)
  {
    // recursive lock releasing inner lock
    l->depth--;
    return 0;
  }
  else
  {
    assert(l->depth == 1);
    l->depth = 0;
    atomic_store(&l->owner, 0);
    ReleaseSRWLockExclusive(&l->lock);
    return 0;
  }
}
```

davidchisnall:
Looks fine to me.  One minor nit: Note that the compiler may not reorder atomic loads or stores, so in the unlock function you're unconditionally loading the value even though you are not always using it.  Moving this initialisation into the case statement would improve things slightly, something like:

```c
  switch (l->attr) {
    case gs_mutex_attr_default:
      break;
    case gs_mutex_attr_errorcheck:
    case gs_mutex_attr_recursive: {
      DWORD thisThread = GetCurrentThreadId();
      DWORD ownerThread = atomic_load(&l->owner);
      // return error if lock is not held by this thread
      if (ownerThread != thisThread) {
        return EPERM;
      }
      break;
    }
  }
```

I'm assuming you're using the C11 `atomic_load` so these are defaulting to sequentially consistent?  I think this is slightly stronger than you need, but should generate the same code on x86 and Aarch64, so it's better to use them than get a weaker ordering wrong.

triplef:
> One minor nit...

Thanks, that change makes total sense.

> I'm assuming you're using the C11 `atomic_load` so these are defaulting to sequentially consistent? I think this is slightly stronger than you need, but should generate the same code on x86 and Aarch64, so it's better to use them than get a weaker ordering wrong.

I’m not so familiar with different ways of atomics and just assumed that’s what you meant in your pseudo code. Is there something else that you would recommend instead?

Using `atomic_load` I also had to change `DWORD owner` to use `atomic_ulong owner`:
```c
typedef struct {
  SRWLOCK lock;
  atomic_ulong owner;
  unsigned long depth;
  gs_mutex_attr_t attr;
} gs_mutex_t;
```

Source/GSFFCallInvocation.m
```diff
@@ -512,7 +509,7 @@ static IMP gs_objc_msg_forward (SEL sel)
 	    (GSIMapVal) forwarding_callback);
 	}
       // Unlock
-      pthread_mutex_unlock (&ff_callback_map_lock);
+      GS_MUTEX_LOCK(ff_callback_map_lock);
```
fredkiefer, line 512: This should be GS_MUTEX_UNLOCK

triplef: Good catch, thank you!

triplef:
I implemented David’s suggestion for recursive locks based on SRWLocks based on the discussion above. The change now feels pretty well-contained, as it leaves almost all the code around locking in place as-is and just replaces the `pthread_mutex/cond_*()` functions with `GS_MUTEX/COND_*()` macros [defined in GSPThread.h](https://github.com/gnustep/libs-base/blob/3e879c149e9559c04c76974d1e9e308764c51700/Source/GSPThread.h).

In addition to the pthread-like functions already pasted above, I had to also add an implementation for `pthread_cond_timedwait()`:
https://github.com/gnustep/libs-base/blob/3e879c149e9559c04c76974d1e9e308764c51700/Source/NSLock.m#L1027-L1051

Windows pre-7 support should also be restored as configure will now check whether SRWLocks are available, and all related code is guarded by `GS_USE_WINDOWS_LOCKS`.

On Windows MSVC the affected tests (e.g. NSLock, NSThread, NSOperation) are passing. Hopefully by the time this is ready we’ll have the CI here back up as well.

davidchisnall:
Windows 7 is no longer supported for security updates, so I think it's fairly irresponsible for any project to maintain support for older versions of Windows for new code.

Using the C11 atomics support is probably the right thing to do.  I don't know which windows toolchains have a `stdatomic.h` though.  If there's a problem, I wrote the 2-clause BSDL version in FreeBSD that includes fallback support for old GCC, so we could use that as an internal header if necessary.  C11 atomics aren't great (they're a back-port of C++11s atomic support with a number of very rough edges) but they're better than anything else in C.  They're one of the main reasons that I've given up on C as a useful programming language for this century.

rfm:
> On 13 Jul 2021, at 11:22, Frederik Seiffert ***@***.***> wrote:
> 
> 
> I implemented David’s suggestion for recursive locks based on SRWLocks based on the discussion above. The change now feels pretty well-contained, as it leaves almost all the code around locking in place as-is and just replaces the pthread_mutex/cond_*() functions with GS_MUTEX/COND_*() macros defined in GSPThread.h.
> 
> In addition to the pthread-like functions already pasted above, I had to also add an implementation for pthread_cond_timedwait():
> https://github.com/gnustep/libs-base/blob/3e879c149e9559c04c76974d1e9e308764c51700/Source/NSLock.m#L1027-L1051

Yes, this is looking good now.  

> Windows pre-7 support should also be restored as configure will now check whether SRWLocks are available, and all related code is guarded by GS_USE_WINDOWS_LOCKS.

OK, though support for old systems in this seems unnecessary; my understanding was that this is only for use with the microsoft toolchain, and I'm not sure why anyone would want to move old windows-xp code to that.

> On Windows MSVC the affected tests (e.g. NSLock, NSThread, NSOperation) are passing. Hopefully by the time this is ready we’ll have the CI here back up as well.

:-)

triplef:
> Windows 7 is no longer supported for security updates, so I think it's fairly irresponsible for any project to maintain support for older versions of Windows for new code.

Yeah not sure what GNUstep’s stance is on this, but I didn’t want to break it without need and just keep the pthread support as a fallback.

> Using the C11 atomics support is probably the right thing to do. I don't know which windows toolchains have a `stdatomic.h` though.

Unfortunately it looks like GCC (v10.2.0) doesn’t provide atomics under MinGW (some details [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65467#c5)):

```
C:/tools/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/10.2.0/include/stdatomic.h:40:1: sorry, unimplemented: '_Atomic' in Objective-C
   40 | typedef _Atomic _Bool atomic_bool;
      | ^~~~~~~
```

>  If there's a problem, I wrote the 2-clause BSDL version in FreeBSD that includes fallback support for old GCC, so we could use that as an internal header if necessary.

Not sure if that will work since GCC already complains about the `_Atomic` keyword, which is also used by the [FreeBSD implementation](https://github.com/freebsd/freebsd-src/blob/main/sys/sys/stdatomic.h). I guess alternatively we can just have configure check for a working `stdatomic.h` and otherwise keep using pthreads.

> OK, though support for old systems in this seems unnecessary; my understanding was that this is only for use with the microsoft toolchain, and I'm not sure why anyone would want to move old windows-xp code to that.

This was meant for all Windows toolchains as it should be more performant and reduce dependencies (once threading also uses native APIs), but if we can’t get it to work or it’s preferred to keep using pthreads under MinGW I’m fine with that too.

davidchisnall:
C11 provides two syntaxes for `_Atomic`, I prefer the one that uses brackets because if you write `_Atomic T *` I need to think to understand if this is a pointer to an atomic `T` or an atomic pointer to a `T`, whereas if you write `_Atomic(T) *` then it's obviously a pointer to an atomic `T` and if you write `_Atomic(T*)` it's obviously an atomic pointer to a `T`.  The FreeBSD header implements a macro that turns `_Atomic(T)` into a `struct` containing a `T` and implements the other functions taking this as arguments.

triplef:
After copying various parts from FreeBSD stdatomic.h I came up with the following, which compiles with GCC and has all relevant tests passing under MinGW. It’s still kind of a mouthful and I can’t say I understand what’s happening here (just e.g. what does `(void)(order)` do exactly in `atomic_exchange_explicit()`?), but together with the native thread API support I just pushed it would allow us to remove the pthread dependency completely on Windows.

```c
#if __has_extension(c_atomic) || __has_extension(cxx_atomic)
#define	atomic_load_explicit(object, order) \
  __c11_atomic_load(object, order)
#define	atomic_store_explicit(object, desired, order) \
  __c11_atomic_store(object, desired, order)
#else
/*
 * No native support for _Atomic(). Place object in structure to prevent
 * most forms of direct non-atomic access.
 */
#define	_Atomic(T) struct { T volatile __val; }
/*
 * __sync_lock_test_and_set() is only an acquire barrier in theory (although in
 * practice it is usually a full barrier) so we need an explicit barrier before
 * it.
 */
#define	atomic_exchange_explicit(object, desired, order) \
__extension__ ({ \
  __typeof__(object) __o = (object); \
  __typeof__(desired) __d = (desired); \
  (void)(order); \
  __sync_synchronize(); \
  __sync_lock_test_and_set(&(__o)->__val, __d); \
})
#define	atomic_load_explicit(object, order) \
  ((void)(order), __sync_fetch_and_add(&(object)->__val, 0))
#define	atomic_store_explicit(object, desired, order) \
  ((void)atomic_exchange_explicit(object, desired, order))
#endif

#ifndef __ATOMIC_SEQ_CST
#define __ATOMIC_SEQ_CST 5
#endif

#define	atomic_load(object) \
  atomic_load_explicit(object, __ATOMIC_SEQ_CST)
#define	atomic_store(object, desired) \
  atomic_store_explicit(object, desired, __ATOMIC_SEQ_CST)

typedef struct {
  SRWLOCK lock;
  _Atomic(DWORD) owner;
  DWORD depth;
  gs_mutex_attr_t attr;
} gs_mutex_t;
```

And if we go with this, would I have to add the full BSD license from FreeBSD’s stdatomic.h?

davidchisnall:
I believe I own the copyright on all of the code you've taken from FreeBSD.  

The `(void)order` is just to avoid warnings that `order` is unused when compiling with a compiler that doesn't implement it. 

I don't know how old a gcc we need to support.  If we don't care about ones more than about 10 years old for Windows (seems reasonable?) then we can use the [more modern GCC atomics](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#g_t_005f_005fatomic-Builtins), something like this:

```
#define	atomic_load_explicit(object, order) \
  __atomic_load_n(object, order)
#define	atomic_store_explicit(object, desired, order) \
  __atomic_store_n(object, desired, order)
```

triplef:
> I believe I own the copyright on all of the code you've taken from FreeBSD.

Ok great, so I won’t include the BSD license then.

> The `(void)order` is just to avoid warnings that `order` is unused when compiling with a compiler that doesn't implement it.

Ah makes sense. Thanks for the clarification.

> I don't know how old a gcc we need to support. If we don't care about ones more than about 10 years old for Windows (seems reasonable?) then we can use the [more modern GCC atomics](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#g_t_005f_005fatomic-Builtins), something like this:

I had tried this, but these seem to require the native `_Atomic` keyword (I tried with and without brackets), which causes this error when used within Objective-C:

`sorry, unimplemented: '_Atomic' in Objective-C`

Not sure if there’s a way around this, but if not the fallback implementation seems to work just fine in my testing.

triplef:
I’ve squashed the changes and added a change log entry, this is now ready for review. Thank you David for all your insightful help with this.

On a higher level the changes are:

- New `GS_MUTEX_*()` and `GS_COND_*()` macros and related typedefs in `GSPThread.h` as an abstraction of mutex/condition variable APIs. These are now used throughout the code base instead of direct pthread API usage. On Windows these map to our own new locking primitives developed with David’s help based on SRW locks.
- New `GS_THREAD_KEY_*()` macros in `GSPThread.h` as an abstraction of thread-specific data APIs. These are now used in NSThread and GSFFIInvocation instead of direct pthread API usage. On Windows these map to Fiber Local Storage (FLS) APIs, as they provide a destructor callback like the pthread equivalent and can be used without using fibers.
- Specific changes in NSLock and NSThread to use native Windows APIs.
- Updates to NSThread tests and the `config.initialize.m` file to use native Windows APIs.
- Updates to configure to check whether `GSPThread.h` compiles using native APIs on Windows. Otherwise pthreads are used as a fallback (guarded by `GS_USE_WIN32_THREADS_AND_LOCKS`).
- Changed `+[NSThread setThreadPriority:]` method signature to return BOOL as on Apple platforms and added error handling as well as tests for thread priority getting/setting.

I’ve tested this with our own app on Windows which uses threading and locking quite extensively, and I’ve run all tests on MinGW-w64 and Windows MSVC, and no new tests are failing and the tests are also passing on CI (Ubuntu).

I appreciate any feedback, including naming / code style.

rfm (commented):
I've tried out the branch on a couple of gnu/linux systems and on msys/mingw64;
: it doesn't seem that anything is broken, so all looks fine.

You wrote:
I’ve renamed the existing types gs_cond_t and gs_mutex_t in GSConfig.h to gs_cond_ivar_t and gs_mutex_ivar_t, and added a 3rd type gs_recursive_mutex_ivar_t there as well. Let me know if a different naming would be preferable.

I was at first a little confused over the renaming (though it makes perfect sense after looking at the code/comments).  I don't have any strong feeling about the name, but since you asked, perhaps gs_mutex_public_t rather than gs_mutex_ivar_t might be a tiny bit clearer in letting people know that the only point of these types is to avoid exposing thread implementation details (having to include pthread.h) in the public headers.

triplef:
Thanks for giving this a try!

`gs_mutex_public_t` makes perfect sense – I’ll change that.

rfm:
Actually, I'm not so sure about the msys/mingw64 build all working properly, I need to check that further.

rfm:
Odd, on msys/mingw64 the NSBundle resource2.m test is hanging, but that seems to be the case on master as well as on wiin-locks

triplef:
Yeah I also found that tests under MinGW can be a bit unreliable at times.

rfm:
Well it doesn't seem to be a problem  related to merging these changes anyawya.

triplef:
Rebased to run against the new CI targets (also squashed the renaming of the public mutex types).

triplef:
@rfm I just wanted to check if you or @fredkiefer wanted to do any further review on this, or if this is good to merge? I’m happy to wait as I realize it’s a big change (and I think Fred might still be on vacation), but just wanted to clarify the status. Thanks for an update!

rfm:
> On 29 Jul 2021, at 08:16, Frederik Seiffert ***@***.***> wrote:
> 
> 
> @rfm I just wanted to check if you or @fredkiefer wanted to do any further review on this, or if this is good to merge? I’m happy to wait as I realize it’s a big change (and I think Fred might still be on vacation), but just wanted to clarify the status. Thanks for an update!

To be clear, visual review of the code changes looks fine and I've tried basic tests and operation on a few different systems to try to confirm that it doesn't break things.  As far as I'm concerned that means it's OK to merge.

triplef:
Great, thanks @rfm!

PR #192, opened 2021-07-16
Fix possible memory corruption in string handling

triplef:
The `TEMP_BUFFER()` macro was making some incorrect calculations. This is used extensively in `NSRegularExpression` and also in `GSUTextMutableString` (which however doesn’t seem to be used).

With `sizeof(unichar) == 2`, the `onstack` variable was initialized for 32 bytes (= 16 characters), while the `onheap` variable (used with more than 64 characters) was only half the size needed.

Note that the macro is always called with the string length and not the buffer size, so I renamed the variable from `size` to `length`.

Note that this increases the `onstack` variable to be 64 bytes, as noted in the macro’s comment. We could also keep this 32 bytes if preferred.

### DIFF
--- a/Source/GSICUString.h
+++ b/Source/GSICUString.h
@@ -80,13 +80,13 @@ static inline void free_string(unichar **buf)
  *
  * Buffers created in this way are exception safe when using native exceptions.
  */
-#define TEMP_BUFFER(name, size)\
+#define TEMP_BUFFER(name, length)\
   __attribute__((cleanup(free_string))) unichar *name ##_onheap = 0;\
-  unichar name ## _onstack[64 / sizeof(unichar)];\
+  unichar name ## _onstack[64];\
   unichar *name = name ## _onstack;\
-  if (size > 64)\
+  if (length > 64)\
     {\
-      name ## _onheap = malloc(size);\
+      name ## _onheap = malloc(length * sizeof(unichar));\
       name = name ## _onheap;\
     }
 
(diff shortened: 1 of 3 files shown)


Source/GSICUString.h
```diff
@@ -80,13 +80,13 @@ static inline void free_string(unichar **buf)
...
  * Buffers created in this way are exception safe when using native exceptions.
  */
-#define TEMP_BUFFER(name, size)\
+#define TEMP_BUFFER(name, length)\
   __attribute__((cleanup(free_string))) unichar *name ##_onheap = 0;\
-  unichar name ## _onstack[64 / sizeof(unichar)];\
+  unichar name ## _onstack[64];\
   unichar *name = name ## _onstack;\
-  if (size > 64)\
+  if (length * sizeof(unichar) > 64)\
```
fredkiefer, line 87: I don't understand why you multiply here. Now the code expects to have the number of unichars in length and only if that is greater than 64 a dynamic allocated buffer is needed.

triplef: I think `length * sizeof(unichar)` is the actual number of bytes we need for the buffer. Since the stack buffer is 64 bytes, we cannot use it if we need more than 64 bytes (i.e. if length is more than 32).

Or do I have a "Denkfehler" as we say in German? :)

fredkiefer: The size of the buffer is in unichar, that is the buffer may hold 64 unichar elements. Only when allocating the dynamic buffer we need to switch over to provide the size in bytes (or use calloc).

Ich denke es ist ein Denkfehler :-)

triplef: Ah yes, thank you for pointing out my Denkfehler! I didn’t consider the type of the "onstack" array.

rfm (approved):

PR #193, opened 2021-07-20
Extend GitHub Actions

triplef:
I’ve built out the GHA integration to include all targets we previously had with Travis:

- Ubuntu GCC
- Ubuntu Clang gnustep-1.9
- Ubuntu Clang gnustep-1.9 disable-mixedabi
- Ubuntu Clang gnustep-2.0
- Ubuntu Clang gnustep-2.0 disable-mixedabi
- Windows GCC MinGW-w64 i686
- Windows GCC MinGW-w64 x86_64

As before, the _tests_ are allowed to fail on Windows (because we have around ~50 failing tests under MinGW), however Windows _builds_ are not allowed to fail to ensure we don’t break them.

All builds use whatever version of GCC/Clang is installed as default. The Ubuntu runner [currently has Clang 10, 11, and 12 pre-installed](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#language-and-runtime), so we could easily run the builds with each of them if we want to. Let me know if this is of interest.

Instead of using `cat` to output config.log/tests.log in the build logs, these files are now always archived as artifacts and are available for download on the [test run page](https://github.com/gnustep/libs-base/actions/runs/1049193442) after all builds finish. I think this makes it easier to check the logs in case of errors, as previously the logs would contain both stdout and the log in one go.

I also took the liberty to convert the README file to Markdown in order to include a CI status badge, update the link for reporting issues to GitHub, and to make the GitHub project page generally more attractive with a formatted readme. You can see the result [here](https://github.com/gnustep/libs-base/tree/gha-ci). Please let me know if this is not desired and I’ll remove the commit.

I’m also planning to integrate builds for Android and Windows MSVC. Additionally there is a number of ways we could extend the CI further, let me know if any of these sound interesting/valuable:

- Build on armv6, armv7, aarch64, s390x, or ppc64le using [Run on Arch](https://github.com/marketplace/actions/run-on-architecture) action
- Build on Debian using [Run on Arch](https://github.com/marketplace/actions/run-on-architecture) action
- Build on FreeBSD, OpenBSD, or Solaris using [VM Actions](https://github.com/vmactions/freebsd-vm)


gcasa (approved):
This looks excellent to me.

rfm (approved):
Looks cool.

triplef:
It seems that the NSThread tests can sometimes hang on Windows as happened in these two recent test runs:

- https://github.com/gnustep/libs-base/runs/3116790752?check_suite_focus=true
- https://github.com/gnustep/libs-base/runs/3117765758?check_suite_focus=true

I’m going to rebase #190 and run the CI a couple of times to see if the native thread implementation might resolve this.

PR #194, opened 2021-07-29
Add support for using ICU DLL from Windows 10 when using MSVC toolchain

triplef:
As suggested by @davidchisnall in https://github.com/gnustep/tools-windows-msvc/issues/2, this PR adds support for using the [Windows-provided ICU libraries](https://docs.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-) when using the MSVC toolchain (I don’t think they can be used under MinGW). Using this requires Windows 10 version 1903 (May 2019 Update) or later.

The main change is in configure, where we now check for `icu.h` and `icu.dll` as a backup if the main check for ICU headers fails, plus using the merged `icu.h` that Windows provides instead of the individual `unicode/*.h` headers.

Also adds missing configure check for unicode/utext.h.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,20 @@
+2021-07-29 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Headers/GNUstepBase/config.h.in:
+	* Source/Additions/Unicode.m:
+	* Source/GSICUString.h:
+	* Source/NSCalendar.m:
+	* Source/NSDateFormatter.m:
+	* Source/NSLocale.m:
+	* Source/NSNumberFormatter.m:
+	* Source/NSPredicate.m:
+	* Source/NSRegularExpression.m:
+	* Source/NSString.m:
+	* Source/NSTimeZone.m:
+	* configure:
+	* configure.ac:
+	Add support for using ICU DLL from Windows 10.
+
 2021-07-29 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Source/NSTimeZone.m:
--- a/Headers/GNUstepBase/config.h.in
+++ b/Headers/GNUstepBase/config.h.in
@@ -344,6 +344,9 @@
 /* Define if you have this function */
 #undef HAVE_ICONV
 
+/* Define to 1 if you have the <icu.h> header file. */
+#undef HAVE_ICU_H
+
 /* Define to 1 if you have the `inet_aton' function. */
 #undef HAVE_INET_ATON
 
@@ -771,6 +774,9 @@
 
 /* Define to 1 if you have the <unicode/ustring.h> header file. */
 #undef HAVE_UNICODE_USTRING_H
+
+/* Define to 1 if you have the <unicode/utext.h> header file. */
+#undef HAVE_UNICODE_UTEXT_H
 
 /* Define to 1 if you have the <unistd.h> header file. */
 #undef HAVE_UNISTD_H
--- a/Source/Additions/Unicode.m
+++ b/Source/Additions/Unicode.m
@@ -56,6 +56,9 @@
 #if     defined(HAVE_UNICODE_UCNV_H)
 #include <unicode/ucnv.h>
 #endif
+#if     defined(HAVE_ICU_H)
+#include <icu.h>
+#endif
 
 
 typedef struct {unichar from; unsigned char to;} _ucc_;
--- a/Source/GSICUString.h
+++ b/Source/GSICUString.h
@@ -1,6 +1,16 @@
 #import <Foundation/NSString.h>
 #import <Foundation/NSException.h>
+
+#if defined(HAVE_UNICODE_UTEXT_H)
 #include <unicode/utext.h>
+#endif
+#if defined(HAVE_ICU_H)
+#include <icu.h>
+// icu.h in Windows 10 is missing a declaration of UTEXT_MAGIC
+#ifndef UTEXT_MAGIC
+#define UTEXT_MAGIC 0x345ad82c
+#endif
+#endif
 
 /*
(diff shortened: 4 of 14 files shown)


fredkiefer (approved):
Looks good to me. It might even work for other compilers on Windows.

fredkiefer:
The test results don't look so good, but tests on Windows have been failing before already. We really need to manually compare the specific test. (And of course fix the tests to properly work on Windows)

For a PR like this it would be great to have a CI build with MSVC as well.

triplef:
> Looks good to me. It might even work for other compilers on Windows.

Yeah I’m not entirely clear on how MinGW works wrt. system DLLs, but I was under the impression that all MinGW apps can only use MinGW-built DLLs also.

> The test results don't look so good, but tests on Windows have been failing before already. We really need to manually compare the specific test. (And of course fix the tests to properly work on Windows)

Hmm looks like all the tests are segfaulting on x86, but run fine on x64, but it seems to be related to the native locks since it started happening with those CI runs. Sorry I missed this and thanks for pointing me to it – I’d only been testing on x64 MinGW. I’m out this weekend but will take a look next week.

> For a PR like this it would be great to have a CI build with MSVC as well.

Yes absolutely, I hope to be able to set that up soon. For this I’ve been running CI builds for the MSVC toolchain in the tools-windows-msvc repo:
https://github.com/gnustep/tools-windows-msvc/actions/runs/1071901749

davidchisnall:
I believe MinGW can use any Windows DLL, but it shares only the C ABI with MSVC.  SEH and C++ exception interop doesn't work and the C++ ABIs are completely different (different name mangling and so on).  Their linkage model is somewhat different as well.

Clang with an msvc target triple attempts to be ABI compatible with the Visual Studio compiler.  That's what we're using here.  Visual Studio doesn't guarantee C++ ABI compatibility across releases though and so the ICU DLLs that Windows ships expose only the C interfaces, not the C++ ones.  Windows traditionally uses COM for things that require ABI stability.  WinObjC has some COM interop logic but I've not looked at it in detail.

triplef:
Thanks for the insightful info David! I’ll try using the icu.dll with MinGW.

Regarding the test failures under 32-bit MinGW, this seems to stem from my previous PR, specifically the use of native threading APIs instead of pthread. While this worked fine under 64-bit MinGW, for some reason it’s causing segfaults on exit with the following stack trace under 32-bit MinGW. Reverting back to pthread instead of of native threading and FLS/TLS APIs in NSThread.m fixes the issue.

```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x77434829 in ntdll!TpSetTimerEx () from C:\WINDOWS\SysWOW64\ntdll.dll
(gdb) bt
#0  0x77434829 in ntdll!TpSetTimerEx () from C:\WINDOWS\SysWOW64\ntdll.dll
#1  0x7745d749 in ntdll!LdrShutdownProcess () from C:\WINDOWS\SysWOW64\ntdll.dll
#2  0x00000001 in ?? ()
#3  0xa1cfb502 in ?? ()
#4  0x77525b40 in ?? ()
#5  0x7745d6b5 in ntdll!RtlExitUserProcess () from C:\WINDOWS\SysWOW64\ntdll.dll
#6  0x76374113 in KERNEL32!ExitProcess () from C:\WINDOWS\SysWOW64\kernel32.dll
#7  0x771f6127 in msvcrt!_exit () from C:\WINDOWS\SysWOW64\msvcrt.dll
#8  0x771f6687 in msvcrt!_initterm_e () from C:\WINDOWS\SysWOW64\msvcrt.dll
#9  0x771f66a1 in msvcrt!exit () from C:\WINDOWS\SysWOW64\msvcrt.dll
#10 0x00c61493 in __tmainCRTStartup () at C:/_/M/mingw-w64-crt-git/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:324
#11 0x7636fa29 in KERNEL32!BaseThreadInitThunk () from C:\WINDOWS\SysWOW64\kernel32.dll
#12 0x77467a7e in ntdll!RtlGetAppContainerNamedObjectPath () from C:\WINDOWS\SysWOW64\ntdll.dll
#13 0x77467a4e in ntdll!RtlGetAppContainerNamedObjectPath () from C:\WINDOWS\SysWOW64\ntdll.dll
#14 0x00000000 in ?? ()
```

Given that MinGW will always have pthreads, I’m leaning towards just always using that under MinGW (32 + 64 bit) and restricting use of native Win32 threading APIs to use of the MSVC toolchain, but I’m open to suggestions.

triplef:
I tried using the system ICU on MinGW, but it looks like the system header location containing icu.h is not part of the search path by default. However, given David’s comment above I’ve enabled the configure check for icu.h and the DLL for both MinGW and MSVC. @fredkiefer would be great if you could take another look over this.

I also checked the crash under 32-bit MinGW a bit more, and it seems to be triggered by the use of the FLS (Fiber Local Storage) APIs. If I change it to use pthread_get/setspecific() but keep the native threading APIs there’s no crash...

I’ll check this further, but I’d first like to merge this and then try to add MSVC builds to the CI (which will require using the tools-windows-msvc repo, which has some interdependencies with this change wrt. building ICU or not).

fredkiefer:
Looks great to me. If we cannot find the headers in MinGW there is not much we could do. Maybe we would also require a lib file to be able to use the dll. If these files ever show up in MinGW, we should automatically use them and until then the old code may work.

triplef:
Tests on 32 bit Windows are fixed in #198 by [annotating](https://github.com/gnustep/libs-base/pull/198/commits/78d46bd6bc6e64bcde0eee7238b48451aefd7d08) the FLS callbacks with `__stdcall`.

PR #197, opened 2021-08-03
Use pkgconfig to find libxslt

triplef:
Use pkgconfig instead of manual checks to find libxslt. Fixes finding library in cross-compilation environments and when library has further dependencies (e.g. when using MSVC toolchain).


fredkiefer (approved):
Looks good!

rfm (approved):
I worry just a little about removing the old code and relying solely on pkgconfig, but xslt has come with pkgconfig support for several years now, so I think it's safe enough.

PR #198, opened 2021-08-06
Add MSVC toolchain to CI

triplef:
Uses the [tools-windows-msvc](https://github.com/gnustep/tools-windows-msvc) scripts to build and install all necessary dependencies, and then builds sources and runs tests normally.

This was quite painful to figure out due to a couple things that in retrospect all make a lot of sense:

- GitHub Actions run on Windows 10 version 1709, which doesn’t yet have the system-provided icu.dll, so I had to re-enable installing ICU on such systems as part of the dependencies. In the process I also had to make the configure check for the system-provided ICU do a try-run, because the Windows SDK does provide icu.lib (so linking works), even if the system doesn’t have icu.dll (running doesn’t).
- Linking debug CRT libraries will cause tests to hang indefinitely, presumably because this will make the system show an alert when abort() is called as part of unhandled exceptions (which some failing tests will throw), so we need to use release builds (probably a good idea anyway). This required patching `gnustep-tests` to not build with `debug=yes`, as this is tied to linking debug libraries (possibly something that should be decoupled at some point).
- And general fun with forward and backward slashes in paths...

The tests now show that many tests segfault on 32-bit under both MinGW and MSVC (c/f https://github.com/gnustep/libs-base/pull/194#issuecomment-891646512). I’ll try to look into that next week.

I also removed the "disable-mixedabi" targets from the CI, as according to NEWS mixed ABI support was removed in v1.28.0.


triplef:
Looks like the `NSString/test02.m` test is hanging on MSVC for some reason, will need to look into it further.

I might also change Make to install DLLs into bin instead of libs under MSVC to match MinGW and simplify some things with the CI setup.

triplef:
Tests on 32 bit Windows are now working again after annotating the FLS callbacks with __stdcall. 👍

I also fixed a couple other issues, though much work remains to fix all the tests on Windows.

fredkiefer (approved):
Looks excellent to me. The minor test issues that remain should be fixed separately.

rfm (approved):
This all looks fine to me. Thanks.

triplef:
Thanks guys!

Issue #186, opened 2021-06-23
WSAStartup() should not be called from DllMain


triplef:
Running our Windows app using GNUstep Base through Application Verifier leads to the following warning:

```
VERIFIER STOP 000000000000E100: pid 0x3E40: Illegal networking API called from DllMain 

	00007FFEC14203B0 : Networking function being called from DllMain
	000001F4678A3FCC : Name of Dll making invalid call if not NULL
	0000000000000000 : Not used
	0000000000000000 : Not used
```

This seems to be caused by [WSAStartup() being called from DllMain in gnustep_base_socket_init()](https://github.com/gnustep/libs-base/blob/master/Source/libgnustep-base-entry.m#L52).

The [WSAStartup() documentation](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup#remarks) says:

> The **WSAStartup** function typically leads to protocol-specific helper DLLs being loaded. As a result, the **WSAStartup** function should not be called from the DllMain function in a application DLL. This can potentially cause deadlocks. For more information, please see the DLL Main Function.

Given that [gnustep_base_socket_init() is also called from `+[NSObject initialize]`](https://github.com/gnustep/libs-base/blob/master/Source/NSObject.m#L982), do we actually need to also call it from DllMain?

rfm:
> On 23 Jun 2021, at 09:49, Frederik Seiffert ***@***.***> wrote:
> 
> 
> Running our Windows app using GNUstep Base through Application Verifier leads to the following warning:
> 
> VERIFIER STOP 000000000000E100: pid 0x3E40: Illegal networking API called from DllMain 
> 
> 	00007FFEC14203B0 : Networking function being called from DllMain
> 	000001F4678A3FCC : Name of Dll making invalid call if not NULL
> 	0000000000000000 : Not used
> 	0000000000000000 : Not used
> 
> This seems to be caused by WSAStartup() being called from DllMain in gnustep_base_socket_init().
> 
> The WSAStartup() documentation says:
> 
> The WSAStartup function typically leads to protocol-specific helper DLLs being loaded. As a result, the WSAStartup function should not be called from the DllMain function in a application DLL. This can potentially cause deadlocks. For more information, please see the DLL Main Function.
> 
> Given that gnustep_base_socket_init() is also called from +[NSObject initialize], do we actually need to also call it from DllMain?

I am no expert on windows stuff, but that shounds reasonable;  I suggest we try it and see how well it works.

PR #200, opened 2021-08-11
Fix various warnings and issues

triplef:
Fixes various warnings that show up in the MSVC CI builds in order to be able to spot any warnings that are actually interesting. A lot of them should also affect other targets (not sure why they don’t show up there – I guess the default warning flags must differ between Clang on different platforms)?

Also reworks the Windows Sockets initialization (#186) and more correctly calls _CRT_INIT with error handling and calling it on both attach and detach as [outlined by MS](https://docs.microsoft.com/en-us/troubleshoot/cpp/use-c-run-time#section-2-using-the-crt-libraries-when-building-a-dll).

The remaining warnings for MSVC builds are:
- CRT deprecation warnings like – these should probably be fixed, so I won’t disable them:
   ```
   warning: 'sscanf' is deprecated: This function or variable may be unsafe. Consider using sscanf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [-Wdeprecated-declarations]
   warning: 'getenv' is deprecated: This function or variable may be unsafe. Consider using _dupenv_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [-Wdeprecated-declarations]
   warning: '_open' is deprecated: This function or variable may be unsafe. Consider using _sopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [-Wdeprecated-declarations]
   warning: '_wopen' is deprecated: This function or variable may be unsafe. Consider using _wsopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [-Wdeprecated-declarations]
   warning: 'inet_ntoa' is deprecated: Use inet_ntop() or InetNtop() instead or define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings [-Wdeprecated-declarations]
   warning: 'inet_addr' is deprecated: Use inet_pton() or InetPton() instead or define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings [-Wdeprecated-declarations]
   ```
- GSIMap.h warnings:
   ```
   warning: cast from 'unsigned long *' to 'struct GSPartMapEnumerator *' increases required alignment from 4 to 8 [-Wcast-align]
      enumerator.node = ((struct GSPartMapEnumerator*)(state->extra))->node; 
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   ```

The warnings in GSIMap.h sound somewhat concerning, and I found the following comment by @davidchisnall from ~10 years ago about this. It’s not clear to me from this whether 64-bit Windows is being handled correctly, and looking at the code I don’t think it is. Any input on this would be appreciated.
https://github.com/gnustep/libs-base/blob/25b25080c3413230033d673376091db2351438d1/Headers/GNUstepBase/GSIMap.h#L997-L1001


triplef:
List of changes:
- use `%lu` and `%ld` for `NSUInteger` / `NSInteger` with cast to `unsigned long` / `long` (as [recommended](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html#//apple_ref/doc/uid/TP40004265-SW5) by Apple)
- use `%lu` format for `DWORD` (aka `unsigned long`)
- use `%lld` and cast to `long long` for `SOCKET` types, which are `int` on Unix and `unsigned long long` on Windows
- use `%ls` instead of [non-standard](https://stackoverflow.com/a/60026359/1534401) `%S` (synonymous) for `const unichar *` and cast to add constant modifier
- use `z` modifier for `size_t` type
- fix some plain incorrect format calls
- fix some `-release` methods to use `oneway void` return type as defined in NSObject
- rework Windows Socket initialization to not be called from `DllMain()` and instead only from `-[NSObject initialize]` (fixes #186)
- clean up the DllMain initialization to handle CRT init errors and call `_CRT_INIT` also on DLL/thread detach, and remove uneeded .idata section after removing WSAStartup() import

triplef:
Changes in Windows tests with these changes:

### Windows MinGW

Unchanged except for some NSURL and NSURLConnection tests that are (still) sometimes failing randomly.

### Windows MSVC i686

- 8212 => 8321 passed tests
- otherwise unchanged

### Windows MSVC x86_64

- 8009 => 8640 Passed tests
- 49 => 47 Failed tests
- 34 Dashed hopes (unchanged)
- 14 => 5 Failed files
- 8 Skipped sets (unchanged)
- 3 Failed builds (unchanged)
- 1 => 0 Failed set

rfm (commented):
This looks good to me.

With regards to the query about the map enumerator, I suppose David's comment is a little unclear, but what I think he meant is that he thinks the code should work:  He's complaining that on windows (because perversely a long is 32bit on a 64bit system) the normal enumerator won't fit in the fast enumeration state., but he's also saying that what he has done works around that limitation (by using a new GSPartMapEnumerator type instead).

His point is that this should work, even on windows, because GSPartMapEnumerator is a structure of two pointers (two 64bit values, total 128 bits), and the 'extra' field of the fast enumeration state is an array of 5 longs (5*32 = 160 bits) so the value will fit.

However, it seems to me that the compiler warning is correct:
typedef struct
{
  unsigned long state;
  id            *itemsPtr;
  unsigned long *mutationsPtr;
  unsigned long extra[5];
} NSFastEnumerationState;

Looking at the fields, we see that extra would be aligned on a 4 byte boundary rather than an 8 byte boundary and this would be an issue on a platform where pointer access needs to be 8byte aligned.  I am not familiar with processor alignment requirements, but this seems a potential problem.  So on an architecture where long is 32bit and pointer is 64bits it seems to me that the enumerator should be stored at &extra[1] rather than at extra.

rfm:
If you change GSIMap.h to use
((struct GSPartMapEnumerator*)(&state->extra[1]))
does it fix the warning, and work?

rfm (approved):

rfm:
Actually, on further consideration the warning is probably bogus.
The structure is a long, then two pointers, then an array of longs.
If the architecture requires 8byte alignment for a pointer, then presumably the compiler should have inserted 4 bytes padding after the first long anyway, in which case the array of longs should begin on an 8 byte boundary.
In which case we just want to suppress the warning.

davidchisnall:
It's probably worth making my comment a bit less rude.

triplef:
I added [some basic tests for NSFastEnumeration](https://github.com/gnustep/libs-base/pull/200/commits/b5ffb22e4dd513ad7e7f9abed7f7801e44196e32) to see if there are any actual issues, although so far I couldn’t find any in my local testing. Could also be that my code doesn’t test the right things tho...

triplef:
I think @rfm you’re right here:

> If the architecture requires 8byte alignment for a pointer, then presumably the compiler should have inserted 4 bytes padding after the first long anyway, in which case the array of longs should begin on an 8 byte boundary.

I tested using `extra[1]` a bit more, and that does crash the NSFastEnumeration test, presumably because now we’re no longer memory-aligned. So I just changed the code to insert an extra cast to `uintptr_t` to silence the warning, and also updated the comment in the process.

rfm:
> On 11 Aug 2021, at 16:34, Frederik Seiffert ***@***.***> wrote:
> 
> 
> I think @rfm you’re right here:
> 
> If the architecture requires 8byte alignment for a pointer, then presumably the compiler should have inserted 4 bytes padding after the first long anyway, in which case the array of longs should begin on an 8 byte boundary.
> 
> I tested using extra[1] a bit more, and that does crash the NSFastEnumeration test, presumably because now we’re no longer memory-aligned. So I just changed the code to insert an extra cast to uintptr_t to silence the warning, and also updated the comment in the process.

Great, your change (including the GS_PART_MAP_ENUMERATOR macro), looks just like what I was playing with here :-)

triplef:
Perfect, thanks!

PR #201, opened 2021-08-17
New NSURL initializers (macOS 10.11 initFileURLWithPath:relativeToURL:) and cleanup

DerNuntius:
I noticed that the **initFileURLWithPath:relativeToURL:**, **initFileURLWithPath:isDirectory:relativeToURL:**, **fileURLWithPath:relativeToURL:** and **fileURLWithPath:isDirectory:relativeToURL:** initializers (and convenience initializers) are not present in gnustep-base. They were added in macOS 10.11.

Here is a full implementation of all initializers listed above with some cleanup in the initialization process (duplicate code removal).

My editor automatically formats the source as described in the _.clang-format_ file. Turns out that the source code did not follow the spec. Most of the changes are automatic clang-format adjustments.

### DIFF
--- a/Headers/Foundation/NSURL.h
+++ b/Headers/Foundation/NSURL.h
@@ -1,21 +1,21 @@
 /* NSURL.h - Class NSURL
    Copyright (C) 1999 Free Software Foundation, Inc.
-
+   
    Written by: 	Manuel Guesdon <mguesdon@sbuilders.com>
    Date:	Jan 1999
-
+   
    This file is part of the GNUstep Library.
-
+   
    This library is free software; you can redistribute it and/or
    modify it under the terms of the GNU Lesser General Public
    License as published by the Free Software Foundation; either
    version 2 of the License, or (at your option) any later version.
-
+   
    This library is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
    Lesser General Public License for more details.
-
+   
    You should have received a copy of the GNU Lesser General Public
    License along with this library; if not, write to the Free
    Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
@@ -24,16 +24,16 @@
 
 #ifndef __NSURL_h_GNUSTEP_BASE_INCLUDE
 #define __NSURL_h_GNUSTEP_BASE_INCLUDE
-#import <GNUstepBase/GSVersionMacros.h>
+#import	<GNUstepBase/GSVersionMacros.h>
 
-#import <Foundation/NSURLHandle.h>
+#import	<Foundation/NSURLHandle.h>
 #import <Foundation/NSRange.h>
 
-#if defined(__cplusplus)
+#if	defined(__cplusplus)
 extern "C" {
 #endif
 
-#if OS_API_VERSION(GS_API_MACOSX, GS_API_LATEST)
+#if	OS_API_VERSION(GS_API_MACOSX, GS_API_LATEST)
 
 @class NSError;
 @class NSNumber;
@@ -44,7 +44,7 @@ extern "C" {
 /**
  *  URL scheme constant for use with [NSURL-initWithScheme:host:path:].
  */
-GS_EXPORT NSString *const NSURLFileScheme;
+GS_EXPORT NSString* const NSURLFileScheme;
 
 /** URL Bookmark Resolution Options **/
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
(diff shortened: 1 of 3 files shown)


fredkiefer:
To me most of the formatting changes you did look wrong. They don't seem to follow the GNUstep formatting. Could you please redo your changes with automatic formatting disabled in your editor and only change the things you need to change?
What I like is the removal of trailing spaces. I do that in gui from time to time. But the way your editor formats method declarations or calls is rather strange.

DerNuntius:
> To me most of the formatting changes you did look wrong. They don't seem to follow the GNUstep formatting. Could you please redo your changes with automatic formatting disabled in your editor and only change the things you need to change?

Will do that!

> What I like is the removal of trailing spaces. I do that in gui from time to time. But the way your editor formats method declarations or calls is rather strange.

It's a clang-format vim extension. I even diffed it against the clang-format output and it was a 1:1 copy. Strange.

DerNuntius:
Aight. This fixes the formatting.

Headers/Foundation/NSURL.h
```diff
@@ -97,6 +97,19 @@ GS_EXPORT_CLASS
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
 + (instancetype) fileURLWithPath: (NSString*)aPath isDirectory: (BOOL)isDir;
 #endif
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_11, GS_API_LATEST)
++ (instancetype) fileURLWithPath: (NSString *)aPath
+	                 isDirectory: (BOOL)isDir
+	               relativeToURL: (NSURL *)baseURL;
+#endif
```
fredkiefer, line 105: As these were added for the same release you could use one #if for both methods. The same is true later on for the other two methods.

Source/NSURL.m
```diff
@@ -587,6 +587,21 @@ + (id) fileURLWithPath: (NSString*)aPath isDirectory: (BOOL)isDir
 					    isDirectory: isDir]);
 }
 
++ (id) fileURLWithPath: (NSString *)aPath
+	      isDirectory: (BOOL)isDir
+	    relativeToURL: (NSURL *)baseURL
+{
+  return AUTORELEASE([[NSURL alloc] initFileURLWithPath:aPath
```
fredkiefer, line 594: I really prefer to have a space after the colon. Not sure whether the GNUstep ObjC guidelines required this, but it would be in line with the rest of the file.

Source/NSURL.m
```diff
@@ -624,70 +639,66 @@ + (id) URLByResolvingAliasFileAtURL: (NSURL*)url
   return nil;
 }
 
-- (id) initFileURLWithPath: (NSString*)aPath
+- (id) initFileURLWithPath: (NSString *)aPath
 {
-  NSFileManager	*mgr = [NSFileManager defaultManager];
-  BOOL		flag = NO;
+  /* isDirectory flag will be overwritten if a directory exists. */
+  self = [self initFileURLWithPath:aPath isDirectory:NO relativeToURL:nil];
```
fredkiefer, line 645: It would be simpler just to return the result of the call. No need to reassign self.

That is:
return [self initFileURLWithPath: aPath isDirectory: NO relativeToURL: nil];

Source/NSURL.m
```diff
@@ -624,70 +639,66 @@ + (id) URLByResolvingAliasFileAtURL: (NSURL*)url
...
-	{
-	  aPath = [aPath stringByAppendingString: @"/"];
-	}
-    }
-  self = [self initWithScheme: NSURLFileScheme
-			 host: @""
-			 path: aPath];
+- (id) initFileURLWithPath: (NSString *)aPath isDirectory: (BOOL)isDir
+{
+  self = [self initFileURLWithPath:aPath isDirectory:isDir relativeToURL:nil];
```
fredkiefer, line 651: Same here

Source/NSURL.m
```diff
@@ -624,70 +639,66 @@ + (id) URLByResolvingAliasFileAtURL: (NSURL*)url
...
+	 raise:NSInvalidArgumentException
+	format:@"[%@ %@] nil string parameter", NSStringFromClass([self class]),
+	       NSStringFromSelector(_cmd)];
     }
   if ([aPath isAbsolutePath] == NO)
     {
-      aPath = [[mgr currentDirectoryPath]
-	stringByAppendingPathComponent: aPath];
+      if (baseURL)
+	{
```
fredkiefer, line 679: The indentation is wrong here. Two spaces for {.

Source/NSURL.m
```diff
@@ -624,70 +639,66 @@ + (id) URLByResolvingAliasFileAtURL: (NSURL*)url
...
-  if (isDir == YES && [aPath hasSuffix: @"/"] == NO)
+  if (isDir == YES && [aPath hasSuffix:@"/"] == NO)
     {
-      aPath = [aPath stringByAppendingString: @"/"];
+      aPath = [aPath stringByAppendingString:@"/"];
     }
-  self = [self initWithScheme: NSURLFileScheme
-			 host: @""
-			 path: aPath];
+  self = [self initWithScheme:NSURLFileScheme host:@"" path:aPath];
```
fredkiefer, line 701: Again, just returning the result would be simpler.

fredkiefer (approved):
There is still a bit wo weird formatting, but otherwise this patch looks OK to me. You still should wait for Richard's opinion before merging it.

rfm (approved):
I should probably have commented earlier, but Fred was doing such a good job ...

I agree about the formatting, but as it's a small change I'm happy to tweak things myself rather than insist on exactly matching style.

I kind of agree about returning the result of a method directly; it's certainly more elegant.
However, I have often found that it's very convenient for working with gdb to have a value assigned to a variable before returning it;  it can make it easier to set a breakpoint after the method but before the result is returned, so that the state of variables can be examined.  So I'm not really bothered which way that's done.

DerNuntius:
Thank you all for the feedback! I updated the code according to the suggestions.

DerNuntius:
I have one question for future code changes. Are you actually using clang-format or is it mostly manual formatting?

rfm:
Entirely manual formatting.

triplef:
So somehow the MSVC CI runs are failing with these errors, even though that code hasn’t been touched in this PR. Can anyone figure out why that might be?

```
NSTimeZone.m:1601:16: error: use of undeclared identifier 'tzname'
          if (NULL != tzname[0] && '\0' != *tzname[0] && 0 == daylight)
                      ^
NSTimeZone.m:1601:38: error: use of undeclared identifier 'tzname'
          if (NULL != tzname[0] && '\0' != *tzname[0] && 0 == daylight)
                                            ^
NSTimeZone.m:1601:56: error: use of undeclared identifier 'daylight'; did you mean '__daylight'?
          if (NULL != tzname[0] && '\0' != *tzname[0] && 0 == daylight)
                                                              ^~~~~~~~
                                                              __daylight
C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt\time.h:66:23: note: '__daylight' declared here
_ACRTIMP int* __cdecl __daylight(void);
                      ^
NSTimeZone.m:1602:56: error: use of undeclared identifier 'tzname'
            localZoneString = [NSString stringWithUTF8String: tzname[0]];
                                                              ^
```

Also, as a side note, I think GitHub users who are not part of GNUstep can’t merge PRs themselves, so someone from the team will have do do that for these cases.

triplef:
@DerNuntius maybe you can also squash your commits to keep the history a little cleaner?

rfm:
> On 18 Aug 2021, at 11:35, Frederik Seiffert ***@***.***> wrote:
> 
> 
> So somehow the MSVC CI runs are failing with these errors, even though that code hasn’t been touched in this PR. Can anyone figure out why that might be?
> 
> NSTimeZone.m:1601:16: error: use of undeclared identifier 'tzname'
>           if (NULL != tzname[0] && '\0' != *tzname[0] && 0 == daylight)
>                       ^
> NSTimeZone.m:1601:38: error: use of undeclared identifier 'tzname'
>           if (NULL != tzname[0] && '\0' != *tzname[0] && 0 == daylight)
>                                             ^
> NSTimeZone.m:1601:56: error: use of undeclared identifier 'daylight'; did you mean '__daylight'?
>           if (NULL != tzname[0] && '\0' != *tzname[0] && 0 == daylight)
>                                                               ^~~~~~~~
>                                                               __daylight
> C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt\time.h:66:23: note: '__daylight' declared here
> _ACRTIMP int* __cdecl __daylight(void);
>                       ^
> NSTimeZone.m:1602:56: error: use of undeclared identifier 'tzname'
>             localZoneString = [NSString stringWithUTF8String: tzname[0]];
>                                                               ^
> 
> Also, as a side note, I think GitHub users who are not part of GNUstep can’t merge PRs themselves, so someone from the team will have do do that for these cases.

I looked thisup online, and found https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/tzset?view=msvc-160
It basically seems to say that the functionality is deprecated (so maybe something is stopping deprecated stuff working in the MSVC build), but a microsoft specific version is available where the only obvious difference is an underscore prefix to the symbols.
I guess a solution might be on the lines of
#if     defined(_WIN32)
#  define	tzset	_tzset
#  define	tzname	_tzname
#  define	daylight	_daylight
#endif

triplef:
Thanks @rfm. You’re right, I think it was caused by an MSVC SDK update from 14.29.30037 (used in the last successful CI run) to 14.29.30133 (used now) on the GitHub Actions runners.

I’ll look into this separately, feel free to merge this despite the failure.

Issue #196, opened 2021-07-30
bus Error TimeZone - Linux


rmottola:
possibly similar to error gotten on Solaris, but on Linux

```
Program received signal SIGBUS, Bus error.
0xffff800100e2ab94 in -[GSTimeZone initWithName:data:] (self=0x0, _cmd=<optimized out>, name=0x0, 
    data=<optimized out>) at NSTimeZone.m:3067
3067	      n_trans = GSSwapBigI32ToHost(*(int32_t*)(void*)header->tzh_timecnt);
```

no bt available

rmottola:
Seems to work now, I can start GWorkspace fine on Linux/SPARC64 and that should be a complex enough example.

Issue #195, opened 2021-07-30
Bus error for timezone access


rmottola:
```Starting program: /opt/GNUstep/Local/Tools/Ink 
[Thread debugging using libthread_db enabled]
[New Thread 1 (LWP 1)]

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 1 (LWP 1)]
-[GSTimeZone initWithName:data:] (self=0x4b3fa8, _cmd=<optimized out>, name=0x114b00, data=<optimized out>)
    at NSTimeZone.m:3067
3067          n_trans = GSSwapBigI32ToHost(*(int32_t*)(void*)header->tzh_timecnt);
```
```(gdb) p  *(struct tzhead *)header
$4 = {tzh_magic = "TZif", tzh_version = "2", tzh_reserved = '\000' <repeats 14 times>, 
  tzh_ttisutcnt = "\000\000\000\b", tzh_ttisstdcnt = "\000\000\000\b", tzh_leapcnt = "\000\000\000", 
  tzh_timecnt = "\000\000\000\252", tzh_typecnt = "\000\000\000\b", tzh_charcnt = "\000\000\000\021"}
```
On Solaris 64bit UltraSPARC

rfm:
> On 30 Jul 2021, at 02:13, Riccardo ***@***.***> wrote:
> 
> 
> [Thread debugging using libthread_db enabled]
> [New Thread 1 (LWP 1)]
> 
> Program received signal SIGSEGV, Segmentation fault.
> [Switching to Thread 1 (LWP 1)]
> -[GSTimeZone initWithName:data:] (self=0x4b3fa8, _cmd=<optimized out>, name=0x114b00, data=<optimized out>)
>     at NSTimeZone.m:3067
> 3067          n_trans = GSSwapBigI32ToHost(*(int32_t*)(void*)header->tzh_timecnt);
> 
> $4 = {tzh_magic = "TZif", tzh_version = "2", tzh_reserved = '\000' <repeats 14 times>, 
>   tzh_ttisutcnt = "\000\000\000\b", tzh_ttisstdcnt = "\000\000\000\b", tzh_leapcnt = "\000\000\000", 
>   tzh_timecnt = "\000\000\000\252", tzh_typecnt = "\000\000\000\b", tzh_charcnt = "\000\000\000\021"}
> 
> On Solaris 64bit UltraSPARC

I'm guessing this is because we have a version 2 file.
[...]

PR #202, opened 2021-08-18
NSTimeZone rewrite

rfm:
This is the work from https://savannah.gnu.org/bugs/?60952 to replace the original NSTimeZone implementation of posix time zones with public domain code handling time transitions in the future (after seconds since 1970 is too being to fit in a 32bit integer) and to more correctly handle version 2+ timezone files generally (the original implementation pre-dates version 2).

I'm fairly confident of the integration of the new code, but am still a little worried about portability as I had to make a lot of changes for that.  There are obvious coding style issues and may be extraneous unused code (I removed a lot of that), but my concern is mostly about the possibility of breaking things by merging the branch, not so much about cosmetic issues that can be tidied up later.


rmottola:
Richard, thanks for asking. I am not at home, so I have a limited set of computers I can test, but will start with them. Only later I can test SPARC, for example.
Does this also maybe fix bug (https://github.com/gnustep/libs-base/issues/195) and (https://github.com/gnustep/libs-base/issues/196) or is that work still left to me?

To test is compilation, run of standard apps and tests enough?

rfm:
> On 18 Aug 2021, at 18:29, Riccardo ***@***.***> wrote:
> 
> 
> Richard, thanks for asking. I am not at home, so I have a limited set of computers I can test, but will start with them. Only later I can test SPARC, for example.
> Does this also maybe fix bug (#195) and (#196) or is that work still left to me?
> 
> To test is compilation, run of standard apps and tests enough?

I think so, yes.

No hurry for this: I'm going to be awayfrom my computer for a week, so won't be ble to do anything about this until the 27th at the earliest.

triplef:
@rfm would be great if you could merge/rebase master again here so we can make sure these changes still build under MSVC. Thanks!

rmottola:
Linux/PPC64BE compiles and passes NSTimeZone tests (although other fails, we should check)

rmottola:
NetBSD/amd64 fails to compile
``` Compiling file NSThread.m ...
NSThread.m:646:1: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'exitedThread'
 exitedThread(void *thread)
 ^~~~~~~~~~~~
In file included from NSThread.m:36:0:
NSThread.m: In function 'GSCurrentThread':
NSThread.m:704:50: error: 'exitedThread' undeclared (first use in this function); did you mean '_exitingThreads'?
       if (!GS_THREAD_KEY_INIT(thread_object_key, exitedThread))
                                                  ^
```

triplef:
That compile error around `exitedThread()` is interesting. I recently added the `GS_WINAPI` in that declaration but that should be defined empty on Unix: https://github.com/gnustep/libs-base/commit/35ef9423e00bffb6c1cc9f336d38c61ac3493208. Can you try removing `GS_WINAPI` there just to see?

rmottola:
> That compile error around `exitedThread()` is interesting. I recently added the `GS_WINAPI` in that declaration but that should be defined empty on Unix: [35ef942](https://github.com/gnustep/libs-base/commit/35ef9423e00bffb6c1cc9f336d38c61ac3493208). Can you try removing `GS_WINAPI` there just to see?

without GS_WINAPI it compiled on NetBSD... running tests right now

rmottola:
NetBSD 9.2 amd64 compiles without GS_WINAPI. NSTimeZone tests pass

```9509 Passed tests
  47 Dashed hopes
  15 Skipped sets
   1 Failed test
```

now next week I will start testing more odd architectures, more PPC, SPARC including Solaris

triplef:
> without GS_WINAPI it compiled on NetBSD... running tests right now

Huh that's strange. GS_WINAPI should be defined as nothing in GSConfig.h. Could you double check that that's the case for you as well?

Source/NSTimeZone.m
```diff
@@ -2905,349 +2905,195 @@ - (NSString*) timeZoneName
...
-  return type->isdst;
+  return type.isdst;
 }
 
 - (NSString*) name
 {
   return timeZoneName;
 }
 
+- (NSDate *) nextDaylightSavingTimeTransitionAfterDate: (NSDate *)aDate
```
fredkiefer, line 3065: Again, what is the purpose of a method just calling super?

Source/NSTimeZone.m
```diff
@@ -2905,349 +2905,195 @@ - (NSString*) timeZoneName
...
     }
   NS_ENDHANDLER
+
+  GS_MUTEX_LOCK(zone_mutex);
+  [zoneDictionary setObject: self forKey: timeZoneName];
+  GS_MUTEX_UNLOCK(zone_mutex);
   return self;
 }
 
 - (BOOL) isDaylightSavingTimeForDate: (NSDate*)aDate
```
fredkiefer, line 3053: Is it worthwhile to have this method and a few others? The super implementation would call timeZoneDetailForDate: and the behaviour would be the same, what we save here is the creation of the NSTimeZoneDetail object.

fredkiefer (commented):
Sorry for taking so long to review this code. I really struggled when trying to understand what goes on here. I did not find any real issue still I do not feel comfortable with the code. This should not block you from merging it (after a bit of clean up).
Did you check how the new format influences the memory use of NSTimeZone? Especially when abbreviationDictionary gets called.

rmottola:
> > without GS_WINAPI it compiled on NetBSD... running tests right now
> 
> Huh that's strange. GS_WINAPI should be defined as nothing in GSConfig.h. Could you double check that that's the case for you as well?

It is a non-issue, switching branches failed a proper reconfigure. Properly configured, GS_WINAPI is not an issue.

PR #203, opened 2021-08-18
Fix building for Windows MSVC with recent toolchain

triplef:
Fixes build errors on CI discussed here: https://github.com/gnustep/libs-base/pull/201#issuecomment-901030162

### DIFF
--- a/Source/common.h
+++ b/Source/common.h
@@ -85,7 +85,7 @@
 #  endif
 #endif
 
-/* Redefine some function names when using the MSVC ABI on Windows.
+/* Redefine some functions/variables when using the MSVC ABI on Windows.
  */
 #ifdef _MSC_VER
 #  include <io.h>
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
The change itself looks good to me but there seem to be a lot more deprecations on Windows that need to be taken care of. Just look through the logs created when compiling with MSVC. Looks like we need to add defines for sscanf ->sscanf_s and quite a few other functions.

rfm (approved):

triplef:
Yeah you’re right @fredkiefer, there are quite a few deprecations with the following suggested replacements:

- sscanf => sscanf_s
- getenv => _dupenv_s
- inet_ntoa => inet_ntop or InetNtop
- inet_addr => inet_pton or InetPton
- _wopen => _wsopen_s
- _open => _sopen_s

As far as I can see, all of them require different parameters though, including sscanf_s which requires 2 parameters for every `%s` format (variable + buffer size). While we currently don’t use sscanf to parse strings but only numbers, I’m not sure if redefining sscanf with sscanf_s would be the right solution here, as it would break if we ever use it for scanning for strings.

I’m guessing the correct solution for these deprecations is to have separate #ifdef _WIN32 implementations for all of these cases, unless you have a different suggestion?

PR #204, opened 2021-08-28
Implement NSURLSessionDownloadTask and NSURLSessionDownloadDelegate

DerNuntius:
The GNUstep NSURLSession API only implements NSURLSessionDataTask. This pull request adds the download functionality. 

An URLSessionDownloadTask will directly write the response data to a temporary file.  When completed, the delegate is sent URLSession:downloadTask:didFinishDownloadingToURL: and given an opportunity to move this file to a permanent location or to otherwise read the file.

But what about the resume functionality?

In Cocoa NSURLSessionDownloadTask creates an opaque data blob to resume a failed download task.
This method relies on an Apple-maintained XPC process to manage the bookmarking of partially downloaded data. Therefore, we'd need to create a similar bookmarking process. Any ideas on how to implement this?

Throughout the testing process I encountered a bug in the GNUstep NSURLSession implementation:
NSURLSessionTask does not retain the NSURLSession object. This results in a segmentation fault when ARC is used. I fixed it by retaining and releasing the object accordingly.

### DIFF
--- a/Headers/Foundation/NSURLSession.h
+++ b/Headers/Foundation/NSURLSession.h
@@ -24,6 +24,7 @@
 @class NSURLResponse;
 @class NSURLSessionConfiguration;
 @class NSURLSessionDataTask;
+@class NSURLSessionDownloadTask;
 
 
 /**
@@ -109,6 +110,12 @@ GS_EXPORT_CLASS
 
 /* Creates a data task to retrieve the contents of the given URL. */
 - (NSURLSessionDataTask*) dataTaskWithURL: (NSURL*)url;
+
+/* Creates a download task with the given request. */
+- (NSURLSessionDownloadTask *) downloadTaskWithRequest: (NSURLRequest *)request;
+
+/* Creates a download task to download the contents of the given URL. */
+- (NSURLSessionDownloadTask *) downloadTaskWithURL: (NSURL *)url;
 
 @end
 
@@ -427,6 +434,37 @@ didReceiveChallenge: (NSURLAuthenticationChallenge*)challenge
            dataTask: (NSURLSessionDataTask*)dataTask
  didReceiveResponse: (NSURLResponse*)response
   completionHandler: (void (^)(NSURLSessionResponseDisposition disposition))completionHandler;
+
+@end
+
+@protocol NSURLSessionDownloadDelegate <NSURLSessionTaskDelegate>
+
+/* Sent when a download task that has completed a download.  The delegate should 
+ * copy or move the file at the given location to a new location as it will be 
+ * removed when the delegate message returns. URLSession:task:didCompleteWithError: will
+ * still be called.
+ */
+- (void)         URLSession: (NSURLSession *)session
+               downloadTask: (NSURLSessionDownloadTask *)downloadTask
+  didFinishDownloadingToURL: (NSURL *)location;
+
+@optional
+/* Sent periodically to notify the delegate of download progress. */
+- (void)         URLSession: (NSURLSession *)session
+               downloadTask: (NSURLSessionDownloadTask *)downloadTask
+               didWriteData: (int64_t)bytesWritten
+          totalBytesWritten: (int64_t)totalBytesWritten
+  totalBytesExpectedToWrite: (int64_t)totalBytesExpectedToWrite;
+
+/* Sent when a download has been resumed. If a download failed with an
+ * error, the -userInfo dictionary of the error will contain an
+ * NSURLSessionDownloadTaskResumeData key, whose value is the resume
+ * data. 
+ */
+- (void)  URLSession: (NSURLSession *)session
+        downloadTask: (NSURLSessionDownloadTask *)downloadTask
+   didResumeAtOffset: (int64_t)fileOffset
+  expectedTotalBytes: (int64_t)expectedTotalBytes;
 
 @end
 
--- a/Source/GSNativeProtocol.m
+++ b/Source/GSNativeProtocol.m
@@ -9,6 +9,7 @@
 #import "Foundation/NSURL.h"
 #import "Foundation/NSURLError.h"
 #import "Foundation/NSURLSession.h"
+#import "Foundation/NSFileHandle.h"
 
 
 static BOOL isEasyHandlePaused(GSNativeProtocolInternalState state) 
@@ -97,6 +98,8 @@ static BOOL isEasyHandleAddedToMultiHandle(GSNativeProtocolInternalState state)
 
 - (void) setError: (NSError*)error;
 
+- (void) setCountOfBytesReceived: (int64_t)count;
+
 @end
 
 @implementation NSURLSessionTask (Internal)
@@ -136,6 +139,11 @@ static BOOL isEasyHandleAddedToMultiHandle(GSNativeProtocolInternalState state)
 - (void) setError: (NSError*)error
 {
   ASSIGN(_error, error);
+}
+
+- (void) setCountOfBytesReceived: (int64_t)count
+{
+  _countOfBytesReceived = count;
 }
 
 @end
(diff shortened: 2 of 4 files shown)


Headers/Foundation/NSURLSession.h
```diff
@@ -430,6 +437,37 @@ didReceiveChallenge: (NSURLAuthenticationChallenge*)challenge
...
+/* Sent when a download task that has completed a download.  The delegate should 
+ * copy or move the file at the given location to a new location as it will be 
+ * removed when the delegate message returns. URLSession:task:didCompleteWithError: will
+ * still be called.
+ */
+- (void)         URLSession: (NSURLSession *)session
+               downloadTask: (NSURLSessionDownloadTask *)downloadTask
+  didFinishDownloadingToURL: (NSURL *)location;
+
+@optional
```
fredkiefer, line 451: For optional parts we normally use something like this:

#if GS_PROTOCOLS_HAVE_OPTIONAL
@optional
#else
@end
@interface NSObject (NSURLSessionDownloadDelegate)
#endif // GS_PROTOCOLS_HAVE_OPTIONAL

I am not sure whether this is still needed for any compiler that we still support. GNUstep is very conservative, we try to keep stuff running even for very old systems.

Source/GSNativeProtocol.m
```diff
@@ -493,24 +502,24 @@ - (NSURLResponse*) validateHeaderCompleteTransferState: (GSTransferState*)ts
...
+  id<NSURLSessionDelegate>     delegate;
 
   task = [self task];
-
   NSAssert(nil != task, @"Cannot notify");
 
+  session = [task session];
+  NSAssert(nil != session, @"Missing session");
+
   delegate = [[task session] delegate];
```
fredkiefer, line 515: Now that you already have the session in a variable, why don't you use it here?

Source/GSNativeProtocol.m
```diff
@@ -493,24 +502,24 @@ - (NSURLResponse*) validateHeaderCompleteTransferState: (GSTransferState*)ts
...
+  NSURLSession                 *session;
+  NSURLSessionTask             *task;
+  id<NSURLSessionDelegate>     delegate;
 
   task = [self task];
-
   NSAssert(nil != task, @"Cannot notify");
 
+  session = [task session];
+  NSAssert(nil != session, @"Missing session");
```
fredkiefer, line 513: This is a change of semantic. Up to now the session was not required (although the code looks like it was). Now it is. Is this really intended?

Source/GSTransferState.m
```diff
@@ -277,6 +280,25 @@ - (void) setData: (NSData*)data
...
+      NSString *fileName;
+      NSURL *tempURL;
+      
+      randomUUID = [NSUUID UUID];
+      fileName = [[randomUUID UUIDString] stringByAppendingPathExtension: @"tmp"];
+      tempURL = [NSURL fileURLWithPath: NSTemporaryDirectory()];
+
+      _fileURL = [NSURL fileURLWithPath: fileName relativeToURL: tempURL];
+
+      RELEASE(randomUUID);
```
fredkiefer, line 297: I don't think that any of these three needs to be released.

Source/GSTransferState.m
```diff
@@ -287,6 +309,17 @@ - (void) setFileURL: (NSURL*)url
 
 - (NSFileHandle*) fileHandle
 {
+  /* Create temporary file and open a fileHandle for writing. */
+  if (!_fileHandle)
+    {
+      NSFileManager *file = [NSFileManager defaultManager];
```
fredkiefer, line 315: The name file looks strange here. Wouldn't fileManager or even fm be better?

Source/GSTransferState.m
```diff
@@ -287,6 +309,17 @@ - (void) setFileURL: (NSURL*)url
...
+      [file createFileAtPath: [[self fileURL] relativePath] contents: nil attributes: nil];
+
+      _fileHandle = [NSFileHandle fileHandleForWritingToURL: [self fileURL] error: NULL];
+
+      RELEASE(file);
```
fredkiefer, line 320: This does not need to be released, you use an already autoreleased NSFileManager.

Source/NSURLSession.m
```diff
@@ -632,7 +681,7 @@ - (void) URLProtocol: (NSURLProtocol *)protocol
 
 @implementation NSURLSessionTask
 {
-  NSURLSession                   *_session; /* not retained */
+  NSURLSession                   *_session;
```
fredkiefer, line 684: Are you sure that you do not create a retain cycle by having the task retain the session? I don't know the rest of the code here but would expect that there was a reason why the session was treated special in the old code.

fredkiefer (changes requested):
It would be great to have a test for the new code.
What I don't like about the already existing structure of the code is the dispatch based on isKindOfClass:. Would it be possible to find a more object oriented solution for this?

Source/NSURLSession.m
DerNuntius: The NSURLSession API is asynchronous. With ARC enabled the NSURLSession object gets released shortly after resume is called at the end of the method. I have to retain the session to keep up with the spec.

``` Objective-C
- (void) resultsForURL: (NSURL *) url 
{
  NSURLSessionConfiguration *defaultConfigObject;
  NSURLSessionDataTask      *dataTask;
  NSURLSession              *session;
 
  defaultConfigObject = [NSURLSessionConfiguration defaultSessionConfiguration];
  session = [NSURLSession sessionWithConfiguration: defaultConfigObject
                                          delegate: self
                                     delegateQueue: [NSOperationQueue mainQueue]];
  dataTask = [session dataTaskWithURL: url];
  
  [dataTask resume];
}
```

Source/GSNativeProtocol.m
DerNuntius: The original implementation declared the session variable inside the if block. I moved it outside to access it in the second if block.

Headers/Foundation/NSURLSession.h
DerNuntius: GCC 4.6 added the ´´´@optional´´´ directive.

DerNuntius:
What about the resume functionality? Any ideas on how to implement that?

Headers/Foundation/NSURLSession.h
DerNuntius: And NSURLSession uses blocks so it will only compile with clang. So this is not needed.

Source/GSNativeProtocol.m
fredkiefer: Yes, I know.  The difference is when there is no session in the old code there would be no delegate and for that reason the assert would not be triggered. Now when there is no session the assert gets triggered. The check in the old code never could be reached.

Source/NSURLSession.m
fredkiefer: I understand this issue, but this does not remove the problem you have created. Now the session retains its tasks and the task retains it session. Neither will ever be removed.
By the way, if you retain the session you also need to assign it in the copy method.

Headers/Foundation/NSURLSession.h
fredkiefer: These are good enough reasons to use @optional directly.

Source/NSURLSession.m
rfm: Logically you'd expect the session to own its tasks (ie the task would not retain its session).  However, searching around I find confirmation that the session does seem to be retained in the Apple implementation, and I suppose it makes sense for convenience if you want to keep a reference to the task but not the session it belongs to.
However, we *do* need to avoid a retain loop causing session and tasks to be leaked.
I guess the way to do that might be to have the task only retain the session once the -resume method is called (ie the operation is started), and release the session as the last thing done once the task has completed.  So if a task is never started (-resume is never called) the session would not be retained, and there would be no retain loop, and similarly when the task stops executing the session would no longer be retained and the retain loop would be broken.

fredkiefer (approved):

rfm (approved):
Looks good to me now.

PR #205, opened 2021-09-14
Merge current testplant_branch back into main repo testplant_branch for eventual merging into the master branch

gcasa:
I believe the approach here should be to merge this into our local testplant specific branch and then cherrypick each of the changes as we review them.   Does this make sense?


gcasa:
I am going to go through all of these changes and see what is relevant to us and what isn't.   We have most of the changes before the divergence of this branch, but not the ones after.

rfm:
It looks to me like most of the changes here are actually from gnustep-base master (ie where TestPlant have merged our code changes into their codebase).   So I think the options for comparison are:
a. look at all the individual commits separately (ignoring ones which are plainly merges from gnustep) or
b. merge the whole lot, then do a diff against master, and try to see what has been changed.
Either looks like a lot of work, but I guess examining individual commits might be easier.

gcasa:
> It looks to me like most of the changes here are actually from gnustep-base master (ie where TestPlant have merged our code changes into their codebase). So I think the options for comparison are:
> a. look at all the individual commits separately (ignoring ones which are plainly merges from gnustep) or
> b. merge the whole lot, then do a diff against master, and try to see what has been changed.
> Either looks like a lot of work, but I guess examining individual commits might be easier.

Unless there is an objection, I am going to go with option b.  We can then cherry-pick from the individual commits and create separate merge requests for each of them or groups of them.  I will merge sometime tomorrow if that works.   It just seems easier to go from our copy of the testplant branch to our master branch.

PR #209, opened 2021-10-07
All relevant testplant/eggplant changes since the divergence of the gnustep_testplant_branch PLEASE REVIEW

gcasa:
This is a manual merge of the testplant/keysight changes into base.  Much of this merge was done by hand since the changes could not be automatically merged due to refactoring or reorganization.  Please review and make comments as you wish.  We can split this merge up if needed.  I have tried to keep the changes in as they had made them so that they can get fair consideration.

I have left some comments on the code in...
https://github.com/gnustep/libs-base/pull/208

### DIFF
--- a/Headers/Foundation/NSObjCRuntime.h
+++ b/Headers/Foundation/NSObjCRuntime.h
@@ -55,7 +55,7 @@
 #endif
 #if defined(__MINGW64__)
 #undef PRIuPTR
-#define PRIuPTR "llu"
+#define PRIuPTR "lu"
 #undef PRIdPTR
 #define PRIdPTR "ld"
 #undef PRIxPTR
--- a/Headers/Foundation/NSObjCRuntime.h
+++ b/Headers/Foundation/NSObjCRuntime.h
@@ -49,6 +49,19 @@
 #define PRIuPTR "lu"
 #endif
 
+/* MINGW 32/64 go here... */
+#if defined(__MINGW32__)
+/* NONE FOR NOW... */
+#endif
+#if defined(__MINGW64__)
+#undef PRIuPTR
+#define PRIuPTR "llu"
+#undef PRIdPTR
+#define PRIdPTR "ld"
+#undef PRIxPTR
+#define PRIxPTR "lx"
+#endif
+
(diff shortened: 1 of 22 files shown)


gcasa:
@rfm I am going to create a branch which I believe contains the simplest fixes from this branch so that all that remains here are the "controversial" fixes.   Hopefully this make the task of determining what should and shouldn't go in a lot easier.

Source/GSRunLoopWatcher.m
```diff
@@ -41,7 +42,7 @@ - (id) initWithType: (RunLoopEventType)aType
 	       data: (void*)item
 {
   _invalidated = NO;
-  receiver = anObj;
+  receiver = RETAIN(anObj);
```
rfm, line 45: We should scrap all this change:  the receiver needs to be un-retained in order to prevent retain cycles which would cause memory leaks.  I have corrected the comments in trunk to say this.

Headers/Foundation/NSObjCRuntime.h
```diff
@@ -49,6 +49,19 @@
 #define PRIuPTR "lu"
 #endif
 
+/* MINGW 32/64 go here... */
+#if defined(__MINGW32__)
+/* NONE FOR NOW... */
+#endif
+#if defined(__MINGW64__)
+#undef PRIuPTR
+#define PRIuPTR "llu"
```
rfm, line 58: Is this right?  I would expect the format for a pointer-length unsigned integer to be lu rather than llu on 64bit.

Source/NSKeyValueObserving.m
```diff
@@ -1713,10 +1713,11 @@ - (void) willChangeValueForKey: (NSString*)aKey
...
   if (pathInfo != nil)
     {
+      RETAIN(pathInfo);
       if (pathInfo->recursion++ == 0)
         {
           id    old = [pathInfo->change objectForKey: NSKeyValueChangeNewKey];
-
+          
           if (old != nil)
             {
```
rfm, line 1722: All looks good to me

gcasa:
This work was moved to another PR, so this one can be closed.

PR #215, opened 2021-11-01
Changes from testplant_merge branch, minus the more complex changes.

gcasa:
Please review.

### DIFF
--- a/Source/NSDistributedLock.m
+++ b/Source/NSDistributedLock.m
@@ -263,11 +263,23 @@ static NSFileManager	*mgr = nil;
 	  attributes = [mgr fileAttributesAtPath: _lockPath
 				    traverseLink: YES];
 	  if (attributes == nil)
-	    {
-	      [NSException raise: NSGenericException
-		format: @"Unable to get attributes of lock file we made at %@",
-		_lockPath];
-	    }
+            {
+              /*
+               * We've seen cases where several instances of the same
+               * program were launched at once, where this exception
+               * got raised. It may be possible that one process removed
+               * the lock we've been using here, and with the right timing
+               * the exception could get raised, however we would only
+               * expect a failure in setting the lock:
+               */
+              /*
+               * [NSException raise: NSGenericException
+               *             format: @"Unable to get attributes of lock file we made"];
+               */
+              NSLog(@"Unable to get attributes of lock file we made");
+              return NO;
+            }
+          
 	  ASSIGN(_lockTime, [attributes fileModificationDate]);
 	  if (nil == _lockTime)
 	    {
--- a/Source/NSDistributedLock.m
+++ b/Source/NSDistributedLock.m
@@ -272,10 +272,8 @@ static NSFileManager	*mgr = nil;
                * the exception could get raised, however we would only
                * expect a failure in setting the lock:
                */
-              /*
-               * [NSException raise: NSGenericException
-               *             format: @"Unable to get attributes of lock file we made"];
-               */
+	       
+              [_localLock unlock];
               NSLog(@"Unable to get attributes of lock file we made");
               return NO;
             }
--- a/Source/NSDistributedNotificationCenter.m
+++ b/Source/NSDistributedNotificationCenter.m
@@ -736,8 +736,8 @@ static NSDistributedNotificationCenter	*netCenter = nil;
 	    }
 #if defined(__MINGW32__)
           NSTask *task = AUTORELEASE([NSTask new]);
-          [task setStandardError:[NSFileHandle fileHandleForWritingAtPath:@"NUL"]];
-          [task setStandardOutput:[NSFileHandle fileHandleForWritingAtPath:@"NUL"]];
+          [task setStandardError:[NSFileHandle fileHandleWithNullDevice]];
+          [task setStandardOutput:[NSFileHandle fileHandleWithNullDevice]];
           [task setLaunchPath:cmd];
           [task setArguments:args];
           [task launch];
--- a/Source/NSDistributedNotificationCenter.m
+++ b/Source/NSDistributedNotificationCenter.m
@@ -759,11 +759,8 @@ static NSDistributedNotificationCenter	*netCenter = nil;
                 }
               NS_HANDLER
                 {
-                  NSLog(@"%s:exception: %@", __PRETTY_FUNCTION__, localException);
+                  NSDebugMLLog(@"%@:exception: %@", NSStringFromSelector(_cmd), localException);
                   _remote = nil;
-                  // Re-raise??? Another exeption will be raised below...
-                  //[pool drain]; // Avoid autorelease pool leak...
-                  //[localException raise];
                 }
               NS_ENDHANDLER;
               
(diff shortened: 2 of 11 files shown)


gcasa:
@rfm @fredkiefer I really would like an approval or review on this at some point.

Source/NSDistributedLock.m
```diff
@@ -263,11 +263,23 @@ - (BOOL) tryLock
...
+               * the lock we've been using here, and with the right timing
+               * the exception could get raised, however we would only
+               * expect a failure in setting the lock:
+               */
+              /*
+               * [NSException raise: NSGenericException
+               *             format: @"Unable to get attributes of lock file we made"];
+               */
+              NSLog(@"Unable to get attributes of lock file we made");
+              return NO;
```
fredkiefer, line 280: This seems wrong to me at it would leave _localLock in the lock state. Also the above description is rather unlikely. We only end up in this branch, when we actually created the directory ourselves. I don't think another process would remove it in the meantime. Most likely there is a completely different issue here that this patch would just hide.

Source/NSDistributedNotificationCenter.m
```diff
@@ -734,18 +734,39 @@ - (void) _connect
 		@"--auto",
 		nil];
 	    }
-	  [NSTask launchedTaskWithLaunchPath: cmd arguments: args];
-
+#if defined(__MINGW32__)
+          NSTask *task = AUTORELEASE([NSTask new]);
+          [task setStandardError:[NSFileHandle fileHandleForWritingAtPath:@"NUL"]];
```
fredkiefer, line 739: Shouldn't we at least use fileHandleWithNullDevice here? And what about input?

Source/NSDistributedNotificationCenter.m
```diff
@@ -734,18 +734,39 @@ - (void) _connect
...
+              NS_DURING
+                {
+                  _remote = [NSConnection
+                              rootProxyForConnectionWithRegisteredName: service
+                                                                  host: host usingNameServer: ns];
+                  RETAIN(_remote);
+                }
+              NS_HANDLER
+                {
+                  NSLog(@"%s:exception: %@", __PRETTY_FUNCTION__, localException);
```
fredkiefer, line 762: Of course we need to remove the commented out code below. But I am also thinking whether __PRETTY_FUNCTION__ will result in a useful method name? Something like NSDebugMLLog would be more useful here.

Source/NSKeyValueObserving.m
```diff
@@ -1713,10 +1713,11 @@ - (void) willChangeValueForKey: (NSString*)aKey
   pathInfo = [info lockReturningPathInfoForKey: aKey];
   if (pathInfo != nil)
     {
+      RETAIN(pathInfo);
       if (pathInfo->recursion++ == 0)
         {
           id    old = [pathInfo->change objectForKey: NSKeyValueChangeNewKey];
-
+
```
fredkiefer, line 1720: This tab should not be included and I am also not sure about the formatting change below.

Source/NSKeyValueObserving.m
```diff
@@ -1713,10 +1713,11 @@ - (void) willChangeValueForKey: (NSString*)aKey
   pathInfo = [info lockReturningPathInfoForKey: aKey];
   if (pathInfo != nil)
     {
+      RETAIN(pathInfo);
```
fredkiefer, line 1716: This could be the actual solution for the problem I had in NSArrayController where I have this horrible workaround that causes a memory leak.
On the other hand I remember trying something similar at that time and running into issues with that solution. Could you please give this a thorough try before merging.

Source/NSSortDescriptor.m
```diff
@@ -113,16 +113,31 @@ - (NSComparisonResult) compareObject: (id) object1 toObject: (id) object2
   id comparedKey1 = [object1 valueForKeyPath: _key];
   id comparedKey2 = [object2 valueForKeyPath: _key];
 
-  if (_comparator == NULL)
-    {
-      result = (NSComparisonResult) [comparedKey1 performSelector: _selector
-                                                       withObject: comparedKey2];
+  // Cocoa seems to either capture the exception thrown in the nil parameter case or
```
fredkiefer, line 116: This looks like a hack to me. What is nil here and isn't there any better way to handle that case?
The solution to always return NSOrderedAscending regardless of the arguments might result in some sorting algorithms to get into an endless loop.

Source/NSURLResponse.m
```diff
@@ -138,6 +138,18 @@ - (void) _setHeaders: (id)headers
...
+          // If key == "Set-Cookie"/"Set-Cookie2" there could be multiple ones...
+          if ([n containsString: @"Set-Cookie"] && [self _valueForHTTPHeaderField: n])
+            {
+              NSMutableString *c = AUTORELEASE([[self _valueForHTTPHeaderField: n] mutableCopy]);
+              [c appendFormat: @", %@", v];
+              [self _setValue: c forHTTPHeaderField: n];
+            }
+          else // Otherwise just set the header field...
+            {
+              [self _setValue: v forHTTPHeaderField: n];
```
fredkiefer, line 151: Most likely this line is supposed to replace line 140, but this is still there.

Source/NSURLResponse.m
```diff
@@ -138,6 +138,18 @@ - (void) _setHeaders: (id)headers
 	      n = [NSString stringWithFormat: @"%@, %@", o, n];
 	    }
 	  [self _setValue: v forHTTPHeaderField: n];
+
+          // If key == "Set-Cookie"/"Set-Cookie2" there could be multiple ones...
```
fredkiefer, line 142: Here the indentation is wrong.
As for the logic here I am not sure what it does, or rather what it should do. First containsString: should most likely be hasPrefix:. The existing code above also seems to try to merge values for the same key. But there we seem to mix values and keys. Maybe somebody needs to write a few test cases for this code to get the desired behaviour better defined?

fredkiefer:
> @rfm @fredkiefer I really would like an approval or review on this at some point.

I was away for work the whole week and could only look at this today.

Source/NSKeyValueObserving.m
gcasa: Will do.  I will add a test for this.

Source/NSDistributedNotificationCenter.m
gcasa: I made the appropriate changes.  Done.

gcasa:
> > @rfm @fredkiefer I really would like an approval or review on this at some point.
> 
> I was away for work the whole week and could only look at this today.

No worries.

Source/NSDistributedLock.m
gcasa: Made a change here to unlock _localLock.  Aren't we leaving it in a lock state if we throw an exception as well?

Source/NSDistributedNotificationCenter.m
gcasa: I changed it to use the method suggested.  Also, I am not sure input is needed as it is launching gdnc here.   One other question.   Will it matter that the handle is opened for writing if we just use the fileHandleWithNullDevice method?

Source/NSKeyValueObserving.m
gcasa: Fixed

gcasa:
Each of these have been manually merged.

PR #216, opened 2021-11-02
Fix `NSGetSizeAndAlignment` usage

madsmtm:
The argument order is `(*sizep, *alignp)`, but it was often incorrectly used the other way around.

This is probably the source of several bugs, but the one that hit me was that an `NSValue` containing `NSRange` didn't return the proper value in `-getValue`. I've added a test for that case, correct me if it's wrong, I'm not familiar with your testing setup.


triplef:
Good find! 🙏

PR #217, opened 2021-11-11
Annotate NSBundle localized string function with NS_FORMAT_ARGUMENT

triplef:
Fixes warnings like the following when using a localized string as a format argument:
```c
[NSString stringWithFormat:NSLocalizedString(@"localized string")]

warning: format string is not a string literal (potentially insecure) [-Wformat-security]
```


rfm (approved):

rfm:
Thanks ... merged

PR #218, opened 2021-11-11
Add required cast in case nil is not defined as id

triplef:
On Apple platforms, `nil` is defined as `NULL`, whereas libobjc2 defines it as `(id)((void*)0)`.

I’ve created https://github.com/gnustep/libobjc2/pull/212 to have libobjc2 match Apple’s implementation, but this results in errors when building libs-base because the GSUNION for GSIArrayItem doesn’t contain an entry for `void*`. This change adds the required cast in libs-base, which doesn’t hurt if it’s already present in the runtime.

### DIFF
--- a/Source/NSKeyedUnarchiver.m
+++ b/Source/NSKeyedUnarchiver.m
@@ -853,7 +853,7 @@ static NSMapTable	*globalClassMap = 0;
 	  // Add markers for unencoded objects.
 	  for (i = 1; i < count; i++)
 	    {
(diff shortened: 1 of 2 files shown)


triplef:
I re-ran the CI on these changes after the libobjc2 changes were merged:
https://github.com/gnustep/libs-base/actions/runs/1450319513

triplef:
FYI since https://github.com/gnustep/libobjc2/pull/212 has been merged in libobjc2, building libs-base now fails without these changes:
https://github.com/gnustep/libs-base/runs/4193956396

PR #219, opened 2021-11-16
improve iconv configure for headers

rmottola:
also add include parameter for checking iconv library, or libc and gnlibiconv might get mixed

### DIFF
--- a/configure
+++ b/configure
@@ -815,6 +815,7 @@ with_ffi_include
 with_ffi_library
 enable_iconv
 with_libiconv_library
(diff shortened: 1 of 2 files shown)


rmottola:
@rfm proven to work and fixes musl issues.
Of course, as discussed it is best if we can still support some BOM workarounds not to require gnulibiconv, yet configure needed fixing.
Regenerated with 2.69 to avoid issues.

rfm (approved):
Looks good, as discussed.

PR #220, opened 2021-11-17
Do not call handler blocks if they are nil

levinli303:
Calling NSProgress.cancel should not crash when no cancel handler is set.

### DIFF
--- a/Headers/GNUstepBase/GSBlocks.h
+++ b/Headers/GNUstepBase/GSBlocks.h
@@ -48,12 +48,12 @@ typedef retTy(^name)()
 /**
  * Calls a block.  Works irrespective of whether the compiler supports blocks.
  */
-#define CALL_BLOCK(block, args, ...) block(args, ## __VA_ARGS__)
+#define CALL_NON_NULL_BLOCK(block, args, ...) block(args, ## __VA_ARGS__)
 
 /**
  * Calls a block without arguments.
  */
-#define CALL_BLOCK_NO_ARGS(block) block()
+#define CALL_NON_NULL_BLOCK_NO_ARGS(block) block()
 #else
 
 /* Fall-back versions for when the compiler doesn't have native blocks support.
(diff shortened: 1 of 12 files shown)


rfm (commented):
The idea certainly seems a good one, but wouldn't it be better to modify GSBlocks.h so that the CALL_BLOCK and CALL_BLOCK_NO_ARGS macros performed the check to see if the block is NULL?

levinli303:
> The idea certainly seems a good one, but wouldn't it be better to modify GSBlocks.h so that the CALL_BLOCK and CALL_BLOCK_NO_ARGS macros performed the check to see if the block is NULL?

yeah I was thinking the same, if its acceptable, then I'd change this PR to modify the macros instead.

rfm:
Yes, I think in the case where we have 'real' blocks, the compiler/runtime will check for a null block, but in the case where we are emulating blocks, we need the macro to add the code to do that check.

levinli303:
@rfm on cocoa, calling a nil block would also crash.

rfm:
Thanks ... good to know.  We should check for null in all versions of the macro then.

levinli303:
@rfm then let's go ahead with this PR as is?

rfm:
I expressed myself poorly.
I was agreeing with your earlier suggestion 
> I'd change this PR to modify the macros instead.

levinli303:
@rfm ok, I've renamed the original macro that does not check nullability to CALL_NON_NULL_BLOCK, and use the original name for the case where nullability is checked. cannot use the checked one for cases where a return value is needed

Headers/GNUstepBase/GSBlocks.h
```diff
@@ -100,13 +100,23 @@ typedef retTy(^name)()
...
+#define CALL_NON_NULL_BLOCK(block, args...) block->invoke(block, args)
+#define CALL_NON_NULL_BLOCK_NO_ARGS(block) block->invoke(block)
 #define BLOCK_SCOPE
 #endif /* GCC_VERSION >= 3000 */
 
 #endif /* __has_feature(blocks) */
 
+#define CALL_BLOCK(block, args...) do { \
+  if (NULL != block) \
+    CALL_NON_NULL_BLOCK(block, args); \
```
fredkiefer, line 112: You just need a "return", plus an else case that will return "nil". This is not perfect, but better than this.

levinli303: do you suggest that I change it to `#define CALL_BLOCK(block, args...) ((NULL != block) ? CALL_NON_NULL_BLOCK(block, args) : nil);` instead ?

rfm: That looks good to me; clear and concise.

levinli303: updated

levinli303: @rfm  can this be merged?

PR #222, opened 2021-11-30
Add the NSDebugEnabled boolean

iamleeg:


rfm (commented):
While I don't really object to this change, it consists of declaring a boolean variable having no effect, which seems so trivial as to be practically pointless.
Is it really the case that NSDebugEnabled is supposed to do nothing?
Or should setting it to YES have an effect on the debugging of core libraries themselves?  Perhaps setting it to YES should turn on default debug (like supplying --GNU-Debug=dflt on the command line)?

fredkiefer:
I fully agree with Richard here. Maybe we also need a way to set that variable from the defaults or an environment variable?

iamleeg:
My short term goal in creating this pull request was to [build MPWTest](https://github.com/mpw/MPWTest/blob/master/testlogger_main.m#L123). But from what I can tell from [the documentation](https://developer.apple.com/documentation/foundation/nsdebugenabled), it's not supposed to do anything other than be available for user code to test.

rfm:
I can't find any reference to this variable actually *doing* anything either, so I don't really like it.
But in the interest of OSX compatibility, it seems reasonable to merge it anyway.

ivucica:
> I can't find any reference to this variable actually _doing_ anything either, so I don't really like it.
> But in the interest of OSX compatibility, it seems reasonable to merge it anyway.

@rfm I went and checked a bit, since I happen to be on a Mac right now.

Setting the environment variable `NSDebugEnabled` to value `YES` will result in the variable being prepopulated with `1` at runtime. I was unable to pass the argument that would trigger this behavior (i.e. `-NSDebugEnabled=YES` did nothing).

image


```objective-c
- (id)initWithCoder:(NSCoder *)coder {
    self = [super initWithCoder:coder];
    if (!self)
    return nil;
    
    NSLog(@"%d", NSDebugEnabled);
// ...
```

results in `0` when environment variable is unset, versus `1` when it is set.

I had to explicitly import `#import <Foundation/NSDebug.h>` for some reason -- importing just `Foundation.h` was insufficient:

> Declaration of 'NSDebugEnabled' must be imported from module 'Foundation.NSDebug' before it is required

Presumably because this is unsupported API. I'll paste some of the relevant (copyrighted) header, in hopes this is acceptable:

```objective-c
/**************************************************************************
WARNING: Unsupported API.

This module contains material that is only partially supported -- if
at all.  Do not depend on the existance of any of these symbols in your
code in future releases of this software.  Certainly, do not depend on
the symbols in this header in production code.  The format of any data
produced by the functions in this header may also change at any time.

However, it should be noted that the features (but not necessarily the
exact implementation) in this file have been found to be generally useful,
and in some cases invaluable, and are not likely to go away anytime soon.
**************************************************************************/
```

There was no additional output in the Xcode output console. I did not pay attention to Console.app.

@iamleeg If you want to fully complete this, I'd probably just add a call to `getenv()` (or equivalent Foundation call) somewhere early in the startup. I did not validate whether this affects console-only applications (i.e. I did not validate where it is likely to be populated).

PR #223, opened 2021-12-14
Default to UTF-8 when reading localized string files

triplef:
Background: We usually use Xcode to edit our Localizable.strings files for all platforms, which defaults to writing them in UTF-8. Currently when launching our app we get a warning about "Localisation file ... not in portable encoding" for every strings file being loaded. Unfortunately we cannot fix this by adding a BOM marker, as Xcode will reliably delete it on every change.

I’m not sure what "the original standard" in the comment in NSBundle.m refers to exactly, but it seems to me that defaulting to ASCII might no longer be the best choice (the [original commit](https://github.com/gnustep/libs-base/commit/9526fb065f59463b2856c03dbe612fffb2990ad7) that introduced this logic is 13 years old). With this change we now always use `[NSString defaultCStringEncoding]` when reading strings files, unless the files has a BOM marker.

I’m not sure what the default encoding works out to be on different platforms, but since both ISO Latin 1 (the hardcoded fallback in Unicode.m) and UTF-8 can be seen as extensions of ASCII I am hoping that this change would not cause any issues with existing strings files.

The other commit here just fixes some code paths when using the Windows-bundled ICU installation.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2021-12-14 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSBundle.m:
+	Default to UTF-8 when reading localized string files without BOM.
+
 2021-12-07 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Source/GSTLS.m: (-disconnect:) try once if we are closing the TCP
--- a/Source/Additions/Unicode.m
+++ b/Source/Additions/Unicode.m
@@ -3025,7 +3025,7 @@ GSPrivateICUCStringEncoding()
   if (icuEnc == GSUndefinedEncoding)
     {
       const char        *encoding = 0;
-#if HAVE_UNICODE_UCNV_H
+#if defined(HAVE_UNICODE_UCNV_H) || defined(HAVE_ICU_H)
       const char *defaultName;
       UErrorCode err = U_ZERO_ERROR;
 
--- a/Source/NSBundle.m
+++ b/Source/NSBundle.m
@@ -2639,10 +2639,10 @@ IF_NO_GC(
           bytes = [tableData bytes];
           length = [tableData length];
           /*
-           * A localisation file can be ...
-           * UTF16 with a leading BOM,
-           * UTF8 with a leading BOM,
-           * or ASCII (the original standard) with \U escapes.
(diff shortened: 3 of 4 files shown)


rfm (changes requested):
I think allowing the system default encoding is a non-starter for strings files because, certainly for any that are part of the gnustep project, but similarly perhaps for files people produce for other projects using gnustep, the resulting strings files would become non-portable:  the strings files need to be used by people with different system string encodings.
I think the way to go is to permit utf-8 (with or without a leading BOM), but continue to support the \U escapes that the original format uses to represent non-ascii characters.  I believe we already do that in places, but I guess the code in NSBundle is a bit stricter (probably got missed when utf-8 support was added elsewhere).

triplef:
Thanks for your feedback, that’s a good point about portability.

Just to make sure I’m understanding your suggestion correctly, is this what we should do?

1. Try reading using `NSASCIIStringEncoding`
2. If that fails, try reading using `NSUTF8StringEncoding`
3. If that fails, try reading using `[NSString defaultCStringEncoding]` and output warn log

rfm:
If people really want to use editors which won't save text as utf-8, we already provide the cvtenc tool to convert between encodings (and to add or remove \U escapes).

triplef:
Thanks, is this what you had in mind: https://github.com/gnustep/libs-base/pull/223/commits/2e83b3d4830a309e8eadc72d4a316efeea397ec3 ?

What’s not clear to me is whether reading as UTF-8 would fail for an ASCII file using \U escapes. Or does reading as UTF-8 also support \U escapes, and we can just remove the ASCII case as it’s handled as part of the UTF-8 path?

triplef:
Sorry was missing a semicolon, here’s the corrected commit:
https://github.com/gnustep/libs-base/pull/223/commits/e817f76a10dff5aa0f15ea50eac38852fb28d1a1

rfm:
Yes, that's what I was thinking of.
Since UTF8 is a superset of ASCII, any ASCII string will be read in correctly.
Then, the -propertyListFromStringsFileFormat method will deal with any \U escapes (irrespective of the character encoding scheme used when writing to  the file).

triplef:
Thanks I think I got it now then.

rfm (approved):

PR #225, opened 2021-12-21
Fix handling of concurrent NSOperations

triplef:
- Concurrent operations never called the completion block.
- In the "isFinished", the observer code in both NSOperation and NSOperationQueue assumed that isFinished was YES when the KVO notification was triggered, but this doesn’t have to be the case.
- The "isFinished" observer for a dependency was removed both when the notification triggered and when the dependency was removed. This assumes an implementation detail of our current KVO implementation, which doesn’t throw an exception if an observer is not actually registered (Apple platforms do). (We previously fixed the same issue for the operation itself in https://github.com/gnustep/libs-base/commit/ae10f58dc456a954d06ecc57376c8e77e9b243ae).

I also added tests for the first two issues (which [failed](https://github.com/gnustep/libs-base/runs/4593591523?check_suite_focus=true) without the changes).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2021-12-21 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSOperation.m: fix handling of concurrent NSOperations
+	if isFinished KVO is triggered without the operation being finished,
+	and call completion block for concurrent operations. Also fixes
+	removing dependency observers more than once.
+
 2021-12-14 Frederik Seiffert <frederik@algoriddim.com>
 
 	* Source/NSBundle.m:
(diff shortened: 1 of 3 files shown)


rfm (approved):
Generally I approve of this change.  It looks god and well commented.
My reservations are minor:
1. I have covid19 and am definitely a least a little cognitively impaired, and this is quite complicated code, so I may have missed something.
2. In the -beginOperation method it's not clear to me whether it's correct to unconditionally call the will/did change value methods, or if they should be called only when the value of the excuting/finished ivar actually changes?

triplef:
Thanks Richard, I hope you're doing ok and wish you a quick and full recovery!

Regarding your second point, I'm pretty sure there's nothing in KVO that says that you may only trigger change notifications if there was actually a value change. Consider any setter that automatically triggers KVO, but happens to be called with the current value - you would still expect this to trigger a KVO notification, right? Otherwise this would add a lot of extra burden to KVO-compliant code. Instead it's up to the receiver of the notification to figure out whether to act on it e.g. by comparing the old and new value that can be part of the change notification.

PR #226, opened 2021-12-25
Support more DOS/Windows string encodings

levinli303:
These NSDOS/NSWindows encodings are not present in Apple's Foundation's public headers but are supported on Apple OSes through `+[NSString availableStringEncodings]`

The list is from https://github.com/youknowone/FoundationExtension/blob/master/FoundationExtension/NSStringEncoding.h , I only added the ones with NSDOS/NSWindows prefix.


levinli303:
Tested with NSDOSRussianStringEncoding on Windows, see 82 in screenshot.

Screen Shot 2021-12-25 at 4 44 50 PM

rfm (approved):
Thanks, that seems fine to me.

Headers/Foundation/NSString.h
```diff
@@ -208,10 +221,42 @@ typedef enum _NSStringEncoding
...
+  NSDOSJapaneseStringEncoding, // code page 932, also for Windows
+  NSDOSChineseSimplifStringEncoding, // code page 936, also for Windows
+  NSDOSKoreanStringEncoding, // code page 949, also for Windows; Unified Hangul Code
+  NSDOSChineseTradStringEncoding, // code page 950, also for Windows
+  // NSWindowsLatin1StringEncoding, // code page 1252, NSWindowsCP1252StringEncoding
+  // NSWindowsLatin2StringEncoding, // code page 1250, Central Europe, NSWindowsCP1250StringEncoding
+  // NSWindowsCyrillicStringEncoding, // code page 1251, Slavic Cyrillic. NSWindowsCP1251StringEncoding
+  // NSWindowsGreekStringEncoding, // code page 1253, NSWindowsCP1253StringEncoding
+  // NSWindowsLatin5StringEncoding, // code page 1254, Turkish, NSWindowsCP1254StringEncoding
+  NSWindowsHebrewStringEncoding, // code page 1255
```
fredkiefer, line 255: If we leave out some definitions aren't the following a bit off?  Should we explicitly specify the number here?

levinli303: do you suggest that we should set number explicitly for all cases? I'm in favor of that.

for these cases that are commented out, i think i could change it to `NSWindowsLatin1StringEncoding = NSWindowsCP1252StringEncoding`

fredkiefer:
Apart from the possible wrong numbers this looks good to me. Overall we should rethink the code here. It looks to complicate for what it is doing. Looks like we are defining the same stuff over and over again.

levinli303:
@fredkiefer I hope we could just use the numbers from Apple, but the code in NSString.m allocates memory of the encoding table based on the value here, so before fixing that, we cannot just use Apple's values that are greater than 0x80000000.

So as of now, in GNUstep's `+[NSString availableStringEncoding]` these will not be included even though they exists in NSString.h:

```
#if OS_API_VERSION(MAC_OS_X_VERSION_10_4,GS_API_LATEST) 
   ,
  NSUTF16BigEndianStringEncoding = 0x90000100,
  NSUTF16LittleEndianStringEncoding = 0x94000100,
  NSUTF32StringEncoding = 0x8c000100,
  NSUTF32BigEndianStringEncoding = 0x98000100,
  NSUTF32LittleEndianStringEncoding = 0x9c000100
#endif
```

fredkiefer (approved):
I didn't find the time to test this code, but the new version looks excellent to me.

levinli303:
I did some test on my WinUI app that uses it.

Using GNUstep Foundation
Screen Shot 2021-12-26 at 6 21 43 PM

Using Apple Foundation on Mac
Screen Shot 2021-12-26 at 6 22 07 PM

The right to left encodings seems to have trouble on Windows but it should be related to WinUI itself.

triplef:
Thanks for your contribution!

Unrelated to this PR, I'd love to hear about your setup using GNUstep with WinUI, e.g. what's your setup to integrate these and are you developing ObjC code in Visual Studio? We tried using WinUI in our Qt app on Windows (which is 90% ObjC using GNUstep), but so far we haven't been able to get that working (though not due to GNUstep but Qt).

levinli303:
@triplef I'm only porting the non-UI part of the app from Mac to Windows so it is relatively easy.

I am using CMake/Ninja/Clang to build ObjC code into a dll, so there is not much fancy stuff going on. The DLL exposes C++ interface (bridged ObjC functions) which then will be consumed in a C++/WinRT component (which is easier to use in a WinUI app). So only C++ code gets compiled in Visual Studio with MSVC. I tried to set up compiling ObjC code directly in Visual Studio but gave up because it is tricky to get it right (for example, a C++.WinRT component has to use MSVC as compiler, a WinUI app project cannot depend on a DLL project with Clang as the compiler).

In Visual Studio I just have the commands to copy these GNUstep DLL/Resources to the output directory.

triplef:
Thanks @levinli303, that’s interesting! Do you happen have any info/link/documentation on why WinUI won’t work with a DLL built using Clang? I thought Clang was producing fully MSVC-compatible binaries at this point.

Without this limitation you could probably use Objective-C++ to integrate your C++/WinRT and Objective-C code without any need for C++ wrappers. We’ve been doing this in our project to quite some extend and it makes it really seamless to build a Windows app using Objective-C.

I’d also love to figure out how to build ObjC code in Visual Studio at some point. In Qt we configure a custom compiler spec using Clang to use for ObjC code, but I’m not sure if VS has a similar feature. It may be sufficient to set the following environment variable somehow in VS to tell Clang to strip the `/TP` flag that interferes with building ObjC code:
`CCC_OVERRIDE_OPTIONS=x-TC x-TP x/TC x/TP`

levinli303:
@triplef WinUI app can use DLL built with Clang, it is only Visual Studio does not allow a WinUI project depend on a Clang DLL project.

levinli303:
So after some digging and hacks tonight, I was actually able to get an ObjC app running all in Visual Studio...

It turns out to be not super complicated. Compiling, linking... all done in Visual Studio without any other tool.

I can document my approach if it is worthwhile. Screen Shot 2021-12-29 at 12 44 19 AM

triplef:
That’s very cool and would definitely help people to get started with this. Could you maybe open a pull request on the [tools-windows-msvc project](https://github.com/gnustep/tools-windows-msvc) to document how to set up a VS project and possibly add an example project?

levinli303:
ok I'll make a PR when I have time.

PR #227, opened 2021-12-26
Complete string encoding localization

levinli303:
Results are from API calls to Apple's `+[NSString localizedNameOfStringEncoding:]`. also changed all files to UTF8 and replaced the \u strings with actual values.

Esperanto is not updated because Apple does not have localize it in that language.


fredkiefer:
The current standard in GNUstep is to provide a UTF8 BOM if the sting files are in UTF8. This is hard to tell in the files you submitted but could you please ensure to set these bytes?

levinli303:
@fredkiefer yes a leading BOM is added to each file
Screen Shot 2021-12-27 at 9 10 40 AM

levinli303:
@fredkiefer is this good to go? the test failures look irrelevant

fredkiefer:
I really would like to understand why the tests are failing. But I am unable to download the actual test.log file. Could you please run the test locally and see whether they fail for you and if so, post the snipped of the test.log file?

levinli303:
@fredkiefer the test failure is 

```
base/NSCalendarDate/test01.m:
82
Failed test:     (2021-12-26 14:38:38.387 +0000) test01.m:74 ... -addTimeInterval: works on a NSCalendarDate parsed with no timezone
```

fredkiefer:
Sorry, I was not clear. This part is already visible when looking at the failed tests in the pipeline. In her background we also log the actual output of all tests and this gets stored in a separate file that also should be uploaded, but I am unable to find that. I think previously we had a "cat" of that file in the script to have it in the automatically stored files. Locally you should be able to see why this test is failing and report that. That way we could tell whether it is related to your changes.
But looking at the actual test code
` date1 = [NSCalendarDate dateWithString: @"Nov 29 06 01:25:38" 
                          calendarFormat: @"%b %d %y %H:%M:%S"
				  locale: locale]; 
  PASS([date1 timeIntervalSinceReferenceDate] + 1 == [[date1 addTimeInterval:1]
  						timeIntervalSinceReferenceDate],
       "-addTimeInterval: works on a NSCalendarDate parsed with no timezone");`
I don't expect to see to much in the log file. We need to output both values to see how they might differ.

Another way forward would be that I merge this PR and inspect and fix the bug later on my machine.

levinli303:
@fredkiefer since I do not have a  build setup with Ubuntu, so probably let's merge this PR for you to test locally.

fredkiefer:
I just started all tests again. Maybe this time we get a different result.

levinli303:
hmm it is still failing, then this PR might be related. let me try to breakdown the changes

levinli303:
@fredkiefer it seems somehow GSTimeZone->sp contains garbage value which caused then lead to the test failure in this PR. I am not sure why it is happening but adding a memset fixed it.

fredkiefer:
Thank you for solving this. And sorry that this unrelated problem blocked you pull request.
The timezone code is rather new, we really need to inspect that code more closely to remove any further issues there.

levinli303:
@fredkiefer no problem, just very weird behavior this would fail 😹

PR #228, opened 2021-12-30
Add support for Simplified Chinese

levinli303:


fredkiefer (approved):

rfm:
While the fix for the timezone bug was great (thanks for spending all the time to track that down and solve it), the main encoding patch broke a whole lot of software because it removed/renamed several of the string encoding constants.
I have tried to fix all that by reinstating the original GNUstep extensions and define any corresponding new constants to be the same as the existing ones.  That lets existing code compile again, but since there's no existing software or testcases for the new constants it's possible I broke something there without noticing;  please could you take a look.

levinli303:
@rfm  I think it is better to list the old encodings as alias to the new encoding so both can be used, what do you think?
```
NSISOCyrillicStringEncoding = NSISOLatinCyrillicStringEncoding;
```

rfm:
> On 4 Jan 2022, at 11:33, Levin Li ***@***.***> wrote:
> 
> 
> @rfm I think it is better to list the old encodings as alias to the new encoding so both can be used, what do you think?
> 
> NSISOCyrillicStringEncoding = NSISOLatinCyrillicStringEncoding;


I did create aliases (rather than deleting the new symbols) in the NSStringEncoding enumeration, so new code should be able to use both interchangably.

eg.
  NSISOLatinCyrillicStringEncoding        = NSISOCyrillicStringEncoding,

levinli303:
ah I missed that. we would also need to update the .strings resources. since we are changing to this too {NSISOCyrillicStringEncoding,
     "NSISOCyrillicStringEncoding","ISO-8859-5",0,1,0},

while the keys in the .strings file are still NSISOLatinCyrillicStringEncoding

rfm:
> On 4 Jan 2022, at 12:07, Levin Li ***@***.***> wrote:
> 
> 
> ah I missed that. we would also need to update the .strings resources. since we are changing to this too {NSISOCyrillicStringEncoding,
> "NSISOCyrillicStringEncoding","ISO-8859-5",0,1,0},
> 
> while the keys in the .strings file are still NSISOLatinCyrillicStringEncoding
> 

Thanks for sporting that:  I have committed a change to restore the .strings files,  though I'm not sure if that was the only option;  we could have changed the canonical names of the encodings (the ascii strings naming each encoding) if they are only used internally by the base library, and I think they are.  However, I think reverting to the original names is the safer option in case I'm mistaken.

PR #230, opened 2022-01-08
Definition of NS_INLINE

gcasa:
Simple definition of NS_INLINE


gcasa:
I think this one is simple, unless either of you have any objections.

fredkiefer (approved):
This is a very simplistic definition of this macro. As a starting point it will do.

gcasa:
> This is a very simplistic definition of this macro. As a starting point it will do.

My thoughts exactly.

PR #232, opened 2022-01-18
Various improvements (mainly) for Windows

triplef:
This comprise a couple of improvements mainly targeting Windows:

- Fix NSTimeZone lookup on Windows MSVC, by converting the Windows time zone names to IANA names used in the NSTimeZone resources ([more info](https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255)).
- Fix `NSTemporaryDirectory()` returning a path with backslashes on Windows, even though we use forward slashes for all internal path handling (with -fileSystemRepresentation converting forward to backward slashes to pass paths to system APIs).
- Non-functional: use `GSNativeChar` in various places handling paths instead of #ifdefs and `unichar` (which is defined as `uint16_t` like `GSNativeChar`).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2022-01-15 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSPathUtilities.m:
+	Fix NSTemporaryDirectory() returning path with backslashes on
+	Windows.
+
 2022-01-13 Frederik Seiffert <frederik@algoriddim.com>
 
 	* Source/NSTimeZone.m:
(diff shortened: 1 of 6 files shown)


triplef:
I’m happy to remove the commit for the last change (replacing `unichar` with `GSNativeChar`), as it might be debatable which makes more sense here. I had originally made these changes when changing `GSNativeChar` to `wchar_t` on Windows, in which context it made a bit more sense.

triplef:
@rfm would be great to get your feedback on this when you have a chance. Thanks!

rfm (approved):
Sorry to take so long; I thought I'd already reviewed these changes.
All three sets of changes look good to me.
I agree that the GSNativeChar change is somewhat a matter of taste, but I think using that makes for more readable code, removing some #ifdefs, so I'm in favour.

triplef:
Thank you, and no worries!

PR #239, opened 2022-02-11
Epf 17843 win32 proxy

DavidHVernon:
Added code to support .pac file config in http proxy for windoze.


gcasa:
@DavidHVernon Please make the corrections suggested by @rfm

PR #241, opened 2022-02-18
Fix some checks when using ICU DLL from Windows 10

triplef:
[`icu.h` from Windows 10](https://docs.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-) doesn’t define `U_ICU_VERSION_MAJOR_NUM`, but is a recent ICU version providing various functionality we need.


triplef:
Looks like this also fixes 5 tests on Windows MSVC. 🎉

triplef:
Thanks, yes exactly, `HAVE_ICU_H` is only set when using the Windows-provided ICU.

PR #242, opened 2022-03-03
Add necessary methods to NSIndexPath to support NSCollectionView

gcasa:
Add needed item, section information in NSIndexPath class

### DIFF
--- a/Headers/Foundation/NSIndexPath.h
+++ b/Headers/Foundation/NSIndexPath.h
@@ -50,9 +50,6 @@ GS_EXPORT_CLASS
   NSUInteger	_hash;
   NSUInteger	_length;
   NSUInteger	*_indexes;
-  
-  NSInteger     _item;
-  NSInteger     _section;
 #endif
 #if     GS_NONFRAGILE
 #else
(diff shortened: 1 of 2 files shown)


fredkiefer:
I just did a small test and it looks like the whole approach of this change is wrong.

`NSIndexPath *is = [NSIndexPath indexPathForItem: 2 inSection: 3];
    //NSIndexPath *is = [NSIndexPath indexPathForRow: 2 inSection: 3];
    NSUInteger x[] = {3, 2};
    NSIndexPath *is2 = [[NSIndexPath alloc] initWithIndexes: x length: 2];
    PASS([is isEqual: is2], "Indexpath is not equal");`

The new methods seem to be shortcuts to build up the same structure that has been used before. No need to add new instance variables.

fredkiefer (changes requested):
Please remove the new instance variables and reimplement the methods based on already existing functionality.

rfm:
The 'section' means an index into the zero'th part of the path.
- (NSUInteger) section
{
  return [self indexAtPosition: 0];
}

The 'item' means an index into the second path of the path.
- (NSUInteger) item
{
  return [self indexAtPosition: 1];
}

So [NSIndexPath indexPathForItem: 2 inSection: 3] is equivalent to [[NSIndexPath indexPathWithIndex: 3] indexPathByAddingIndex: 2];

fredkiefer (approved):

rfm (commented):
Looks fine to me

gcasa:
> This looks good. And while you are at it, could you also add the new (13.0) `row` method?

Didn't see this... I will do this as a quick fix in just a few.

PR #245, opened 2022-03-21
Fix time zone lookup on Windows MSVC when using non-English locale

triplef:
Fixes the conversion of Windows time zone names to IANA names introduced in #232 on systems not using an English locale, and adds a warning when [`ucal_getTimeZoneIDForWindowsID`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ucal_8h.html#ac345a8561828554c092663e541b1c7bd) is unable to map the time zone (in which case it’s documented to return success but a zero length string).

### DIFF
--- a/Source/NSTimeZone.m
+++ b/Source/NSTimeZone.m
@@ -1438,9 +1438,19 @@ static NSMapTable	*absolutes = 0;
        */
       {
         TIME_ZONE_INFORMATION tz;
-        DWORD dst = GetTimeZoneInformation(&tz);
+        DWORD dst;
         wchar_t *tzName;
-  
+
+#if defined(_MSC_VER) && defined(UCAL_H)
+        // Get time zone name for US locale as expected by ICU method below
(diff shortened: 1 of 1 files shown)


rfm (commented):
Reading around this, the changes all look fine.  However, I wonder if making them conditional on _MSC_VER is necessary;  would they not work whether that define is present or not?

triplef:
I didn’t try this but I thought maybe MinGW behaves more like Unixes and already return IANA time zone names in the first place. At least all the NSTimeZone tests are passing on MinGW so I didn’t want to change anything on that platform. It would probably be worth checking that though.

hmelder:
On MinGW the standard IANA format is used. tzdata is a MinGW core dependency, as seen in the screenshot.

tzdata: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-tzdata/PKGBUILD

image

triplef:
Thanks for checking @hmelder!

PR #249, opened 2022-04-04
Add 10.9 methods for getEra:..., getHour:... etc

gcasa:
This PR adds 10.9 methods for:

```
#if OS_API_VERSION(MAC_OS_X_VERSION_10_9, GS_API_LATEST)
- (void) getEra: (NSInteger *)eraValuePointer
           year: (NSInteger *)yearValuePointer
          month: (NSInteger *)monthValuePointer
            day: (NSInteger *)dayValuePointer
       fromDate: (NSDate *)date;

- (void) getHour: (NSInteger *)hourValuePointer
          minute: (NSInteger *)minuteValuePointer
          second: (NSInteger *)secondValuePointer
      nanosecond: (NSInteger *)nanosecondValuePointer
        fromDate: (NSDate *)date;

- (void) getEra: (NSInteger *)eraValuePointer 
yearForWeekOfYear: (NSInteger *)yearValuePointer 
     weekOfYear: (NSInteger *)weekValuePointer 
        weekday: (NSInteger *)weekdayValuePointer 
       fromDate: (NSDate *)date;
#endif
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2022-04-06 Gregory John Casamento <greg.casamento@gmail.com>
+
+	* Headers/Foundation/NSCalendar.h: Add documentation.
+	* Source/NSCalendar.m: Implement getEra:... and getHour:...
+        methods based on Keysight/TestPlant changes
+	* Tests/base/NSCalendar/basic.m: Minor cleanup.
+	* Tests/base/NSCalendar/component-diff.m: Add tests for
+        methods implemented above.
+
 2022-03-07 Gregory John Casamento <greg.casamento@gmail.com>
 
         * Headers/Foundation/NSIndexPath.h: Add declarations for new
--- a/Headers/Foundation/NSCalendar.h
+++ b/Headers/Foundation/NSCalendar.h
@@ -329,6 +329,27 @@ GS_EXPORT_CLASS
             interval: (NSTimeInterval *)tip
              forDate: (NSDate *)date;
 #endif
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_9, GS_API_LATEST)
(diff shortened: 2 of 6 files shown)


rfm (commented):
1. In an ideal world we'd have an implementation for use when ICU is not available, but realistically that's not worth the effort.  Normally, when a method is unimplemented it should call the -notImplemented: method to raise an exception.
2. The documentation of the new methods is missing
3. New methods call for new regression test cases
Apart from that, all looks good to me.

gcasa:
> 1. In an ideal world we'd have an implementation for use when ICU is not available, but realistically that's not worth the effort.  Normally, when a method is unimplemented it should call the -notImplemented: method to raise an exception.
> 
> 2. The documentation of the new methods is missing
> 
> 3. New methods call for new regression test cases
> 
> Apart from that, all looks good to me.

I will add tests and documentation.  Once I do that I will go ahead and merge.   Thanks.

gcasa:
@rfm please take a quick look at this before merge.

gcasa:
Since you said earlier that everything looks okay, and all tests pass including the ones I have added... I am going to proceed with the merge. :) Feel free to yell at me later, GC.  :)

triplef:
Thanks for implementing these @gcasa!

I guess it’s only missing the call to `-notImplemented:` to generate an exception if ICU is missing, as mentioned by @rfm above.

gcasa:
> Thanks for implementing these @gcasa!
> 
> I guess it’s only missing the call to `-notImplemented:` to generate an exception if ICU is missing, as mentioned by @rfm above.

When I wrote these, I followed the pattern of what was done for the other methods already implemented.   The other methods also require ICU, but do nothing if it is not present.  I am wondering if writing something that doesn't use ICU would be better.

PR #251, opened 2022-06-08
Add new methods for NSTask

gcasa:
Add implementation for 10.13 methods to NSTask.


rfm (commented):
It looks to me like the point of the new API (other than a trivial callback on task completion) is to get detaqiled error information if/when a task fails to launch.
So until that feature is actually implemented, the new methods should call -notImplemented: to raise an exception.
It looks to me like, to actually implement this, requires the -launchAndReturnError: method to contain all the code currently in the -launch method (and new code to create the NSError object when the task fails to launch), and the -launch method to be reimplemented as a simple  wrapper round the new method.

gcasa (commented):
My thoughts exactly.  I was working towards this.

gcasa:
Assuming that this is okay to merge...

PR #252, opened 2022-08-03
Fix NSBundle unit test on Windows MSVC

hmelder:
In addition to the fix in tools-make, this PR fixes the NSBundle unit test.

The Framework set, which imports a framework located in Resources/TestFramework.framework, is not supported on Windows (MSVC and MinGW) as Windows lacks support for run-time search paths (rpath). Another set imports the gnustep-base bundle, i.e. framework loading is still tested.


rfm (approved):
Looks fine, thanks.

PR #254, opened 2022-08-04
GSTLS, NSData, NSString, and NSProgress Unit Test Fixes on Windows MSVC

hmelder:
- Fix GSTLS unit test
- NSData unit test
   - Check for nil in NSData
- NSString unit test
- Fix NSProgress unit test
   - Replace Objective-C string with cstring in PASS_EQUAL macros
   - Fix UserInfo check by retrieving UTF8String


hmelder:
I'll update the ChangeLog and then this branch is ready :)

PR #255, opened 2022-08-04
NSArchiver Unit Test Fixes (Size of BOOL) on Windows MSVC

hmelder:


rfm (approved):
Thanks for doing that.

PR #256, opened 2022-08-04
GSTLS, NSData, NSString Unit Test Fixes on Windows MSVC

hmelder:
- Fix GSTLS unit test
- NSData unit test
   - Check for nil in NSData
- NSString unit test
- Fix NSProgress unit test
   - Replace Objective-C string with cstring in PASS_EQUAL macros
   - Fix UserInfo check by retrieving UTF8String

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,9 +4,6 @@
 	Disable test completly if gnutls is not present.
 	* Tests/base/NSData/additions.m:
 	Check if dataWithContentsOfFile: returns an instance.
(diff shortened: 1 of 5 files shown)


rfm (commented):
Looks good, but I think I already fixed the faulty use of ObjC strings where C strings should be in response to someone else's bug report.

hmelder:
> Looks good, but I think I already fixed the faulty use of ObjC strings where C strings should be in response to someone else's bug report.

Aight. I'll rebase the branch and remove the NSProgress fix :)

PR #257, opened 2022-08-05
Fix NSTask unit tests on Windows MSVC

hmelder:


Tests/base/NSTask/notify.m
```diff
@@ -32,7 +32,19 @@ - (void) taskDidTerminate: (NSNotification *)notification
...
   BOOL          earlyTermination = NO;
+  
+  /* Windows MSVC adds the '.exe' suffix to executables
+   */
+#if defined(_MSC_VER)
+  testecho = @"testecho.exe";
+  testsleep = @"testsleep.exe";
+#else
+  testecho = @"testecho";
+  testsleep = @"testsleep.exe";
```
triplef, line 46: This should not have the exe suffix.

hmelder: You are right

rfm (approved):
Thanks.

PR #258, opened 2022-08-07
Windows to IANA time zone conversion: remove unsupported daylight conversion

hmelder:
There are still two dashed hopes in the NSTimeZone/use.m unit test, because GMT-3 is returned instead of GMT-03:00. Some conversion would be required to match the spec.


Source/NSTimeZone.m
```diff
@@ -1452,10 +1452,12 @@ + (NSTimeZone*) systemTimeZone
...
-        if (dst == TIME_ZONE_ID_DAYLIGHT)
-          tzName = tz.DaylightName;
-        else
-          tzName = tz.StandardName;
+        // Only tz.StandardName time zone conversions are supported, as
+        // the Zone-Tzid table lacks all daylight time conversions:
+        // e.g. 'W. Europe Daylight Time' <-> 'Europe/Berlin' is not listed.
+        //
+        // See: https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/zone_tzid.html
+        tzName = tz.StandardName;
```
triplef, line 1460: I think this should be conditional for MSVC only. As you mentioned in another PR, MinGW uses IANA timezone names already, and all tests are passing, so I don’t think we should change anything for MinGW:

PR #260, opened 2022-08-10
NSXML unit test and NSURL file URI fixes

hmelder:
The sloppy parser still has some issues on Windows platforms.


rfm (approved):
That looks good

PR #262, opened 2022-08-11
Mark notANumber test case as hopeful

hmelder:
It is not guaranteed by ICU that UNUM_NAN_SYMBOL is "NaN". On Windows, NaN has a negative signum and returns "-NaN".


rfm (approved):
That looks good.

I'm honestly not sure what the correct/best solution is in the long term ... it seems to me that the symbol would ideally be consistent (having the alternatives NaN and -NaN seems stupid) across all systems, but might reasonably vary between locales  for different languages.

Making the test a hope is a good short term fix.

PR #264, opened 2022-08-12
Replace make check with gnustep-tests in CI to avoid edge-cases

hmelder:
Running 'make check' on Windows environments has some quirks:

- Tests randomly fail or crash
- Time zones are broken
- Configuration errors:  general.m 25 ... it looks like GNUstep-base is not yet installed

Running 'gnustep-tests' directly is more reliable and works as expected.


triplef:
Looks like this is hanging on CI.

hmelder:
> Looks like this is hanging on CI.

It was stuck on NSException.

hmelder:
It is weird that the change only fails on MSVC.

PR #265, opened 2022-08-16
win32: Support overlapped (asynchronous) I/O on standard streams in GSFileHandle

hmelder:


hmelder:
It seems like win32/GSFileHandle uses tabs instead of spaces for indentation.

hmelder:
Should I retab it or use tabs for indentation? @rfm

rfm:
I think it's better to standardise on spaces.

hmelder:
> I think it's better to standardise on spaces.

Should I convert the complete source file to spaces?

rfm:
> Should I convert the complete source file to spaces?

I had a look at the file and can see that it's all correctly indented.   It's not indented with tabs.  It uses tabs and spaces with tabstops at the conventional 8 space intervals, so it doesn't really need any changes to work in decent code editors (with fixed width fonts).
However, using spaces exclusively will do no harm.

triplef:
> It's not indented with tabs. It uses tabs and spaces with tabstops at the conventional 8 space intervals

FWIW I’m always struggling to configure my editors to deal with this combination of tabs and spaces, which I’ve also never seen in any other project. Would be great if going forward we could use spaces exclusively to keep things simple.

rfm:
I don't see a problem with using all spaces.

hmelder:
Adjusted the indentation @rfm

rfm (approved):
Looks good, thanks.

Issue #266, opened 2022-08-18
GSSocketStream win32 Assertion


hmelder:
## Brief Introduction
The GSInetSocketStream/GSInet6SocketStream classes abstract platform-dependent networking APIs away and expose them via the standard NSStream interface. Each socket stream has an input (GSSocketInputStream) and an output stream (GSSocketOutputStream).

After opening a new socket, the socket is recorded and monitored. On POSIX compliant systems the socket file descriptor is used as the loop id [GSSocketStream.m:1715](https://github.com/gnustep/libs-base/blob/560b1d19fb8062ed5a870f34e92584b708df4ef4/Source/GSSocketStream.m#L1715). On Windows, the loop id is an event object which is set up to receive socket events. The event object is associated with the socket via the [WSAEventSelect](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaeventselect) function.

## Winsock2 Events
Windows only permits a single event object to be associated with a socket at any time. GSInetSocketInputStream coordinates two Windows event objects via the _dispatch method. If one stream is closed, the _dispatch method [GSSocketStream.m:2041](https://github.com/gnustep/libs-base/blob/560b1d19fb8062ed5a870f34e92584b708df4ef4/Source/GSSocketStream.m#L2041) automatically switches to the second (sibling) event object [GSSocketStream.m:1926](https://github.com/gnustep/libs-base/blob/560b1d19fb8062ed5a870f34e92584b708df4ef4/Source/GSSocketStream.m#L1926).

The _dispatch method discovers changes for the selected socket. On Windows, the [WSAEnumNetWorkEvents](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaenumnetworkevents) function is used.

## First Bug
We use the [keepalive](https://github.com/gnustep/libs-base/blob/master/Tests/base/NSURL/Helpers/keepalive.m) test as an example:

Starting the server:
```sh
$ ./obj/keepalive.exe --GNU-Debug=NSStream -FileName C:/tools-windows-msvc/src/gnustep-base/Tests/base/NSURL/Chunked.dat -FileHdrs YES -Count 1 -Port 12345
```

Log (GSInetServerStream is initialised and added to the default Runloop):
```
[...]

PR #268, opened 2022-08-23
Skip all NSURL* test cases using GSInetServerStream on Windows

hmelder:


hmelder:
Btw I think we should deprecate GSServerStream. It is not using anywhere in our code base and only exposed as a category in GNUstep additions. The implementation is too complicated for what it does.

hmelder:
Is it ok to merge @rfm ?

rfm:
I'm not sure of the rationale here.
While GSServerStream is (so far as i can see) only used in NSSetServices in the base library and in a few testcases, I wonder if it's really not useful / too complicated?  I think it would be fine to deprecate and then replace it in those places (and remove it from the additions api)  if it's easy.
But, if it's simple to replace, why not do that rather than removing the tests?

hmelder:
>> But, if it's simple to replace, why not do that rather than removing the tests?
The unit tests implement multiple test server based on GSServerStream. Rewriting all of them would be a major undertaking. I think we should temporarily disable them.

triplef:
#266 has some more info about the issues with GSServerStream on Windows.

For now it would be great if we can mark the tests using it as hopeful on Windows (as this PR does). This will allow us to disable `allow-test-failures` on the Windows CI targets, so that possible future changes breaking Windows builds will actually fail those builds.

rfm:
Marking the tests as hopeful seems fine to me.

hmelder:
> Marking the tests as hopeful seems fine to me.

This was my original plan, but I ran into some issues back then. I can look into this and update the PR.

hmelder:
Forgot the semicolons. Wrote to much golang :)

hmelder:
Done! We have no test failures in Windows MinGW GCC x86_64 and Windows MSVC x86_64. There is still an unrelated test failure on 32-bit machines. Ready to check @rfm

triplef:
Awesome, great job @hmelder! 🎉

I think this should be removed then as well (for both 64-bit Windows targets):
https://github.com/gnustep/libs-base/blob/996a5b8a841c15ac6592cdb654ed2655d6c5ee3d/.github/workflows/main.yml#L67

hmelder:
@rfm ready to review

RoyalStewart90:
Hi @hmelder, very sorry to be this late to the party, earlier this year I was poking around the IPv6 functionality and made some changes to my company's branch of Master. In your experience, what was the issue with GSInetServerStream? I had encountered a few issues of formatting IPv6 address with brackets related to GSStreams. The biggest issue I ran into was memory allocation of IP addresses. I'm hoping at some point we can get this issue fixed and re-enable more connection tests.

hmelder:
@RoyalStewart90 see #266

I think it was an issue with the asynchronous event handling on Windows. I would have to go into the issue as this was about a year ago :)

PR #269, opened 2022-08-23
Rewrite autoconf libcurl detection

hmelder:
- Bump minimum supported version to 7.66.0 and remove old checks for quirks
- Use pkg-config instead of curl-config, as curl-config is not installed when targeting Windows.
- Remove gnutls warning, as no loss of functionality is observed.


rfm (changes requested):
I think there are multiple problems with this patch breaking things.
The use of pkg-config is a last resort because it is not be available on some systems.
If we don't have curl-config to tell us about curl, we could try pkg-config as a fallback.

configure.ac
```diff
@@ -3615,71 +3615,31 @@ AC_SUBST(HAVE_LIBDISPATCH_RUNLOOP)
...
-  curl_maj=`echo $curl_ver | sed -e "s/^\(.*\)\.\(.*\)\.\(.*\)$/\1/"`
-  curl_min=`echo $curl_ver | sed -e "s/^\(.*\)\.\(.*\)\.\(.*\)$/\2/"`
-  if test $curl_maj -lt 7 -o \( $curl_maj -eq 7 -a $curl_min -lt 49 \); then
-    AC_MSG_RESULT([FAILED (version too old to use])
-  else
-    AC_MSG_RESULT(yes ... version $curl_ver)
-    AC_CHECK_HEADERS(curl/curl.h, curl_ok=yes, curl_ok=no)
+
+if test -n "$PKG_CONFIG"; then
+  if pkg-config --exists libcurl; then
```
fredkiefer, line 3624: Shouldn't you be using $PKG_CONFIG here as well (and in line 3631)?

hmelder: You are right

PR #270, opened 2022-08-23
Detect CRLF in GSMime tests

hmelder:
This test fails when running on msys2, but succeeds when executed from cmd.exe. It passes in CI.

### DIFF
--- a/Tests/base/GSMime/general.m
+++ b/Tests/base/GSMime/general.m
@@ -118,7 +118,7 @@ int main()
   [parser parse: data];
   doc = [parser mimeDocument];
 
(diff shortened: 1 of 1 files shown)


hmelder:
The msys2 runtime automatically converts CRLF to LF. I think that matters because gnustep-tests is a POSIX shell script running on the msys2 bash version, but I don't understand how the CI environment is different from my local one.

https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/#runtime

rfm (approved):
Being tolerant of both cases here makes sense to me.

PR #271, opened 2022-08-23
Allow CRLF and LF in NSXMLParser test

hmelder:


rfm (approved):
Yes, thanks.

Tests/base/NSXMLParser/parse.m
```diff
@@ -257,13 +259,12 @@ - (void) reset
...
-#if defined(_WIN32)
       /* Replace CRLF from result files with LF
        */
-      expect = [expect stringByReplacingOccurrencesOfString: @"\r\n" withString: @"\n"];
-#endif
+      convertWin = [expect stringByReplacingOccurrencesOfString: @"\r\n" withString: @"\n"];
+      desc = [handler description];
 
-      if (NO == [[handler description] isEqual: expect]) 
+      if (NO == [desc isEqual: expect] && [desc isEqual: convertWin] == NO)
```
fredkiefer, line 267: It would be nicer to have NO always on the same side, that is to the left of the equal signs.

hmelder: Alright.

PR #279, opened 2022-10-16
fix missing links in generated documentation

onflapp:
The offline documentation is missing links to the manual and coding standards. 
This is due to errors in the texi files which prevent the make file to generate HTML versions and a broken link.


rfm (approved):
Thanks ... looks good.

PR #280, opened 2022-10-31
ICU specifies a NULL is passed for default locale

sbeard-eh:
From https://unicode-org.github.io/icu/userguide/locale/#locales-and-services:

"Using the ICU C functions, NULL can be passed for a locale parameter to specify the default locale."

On my systems, passing an empty string either results in the systemLocale being set to "en_US_POSIX" (when using ICU ver50.0), or an empty string (when using ICU ver67.0).

Passing 'nil' results in the system locale being read correctly. On Centos/7 and AlmaLinux9.0, this matches the value set for 'LANG' in the /etc/locale.conf file.

Thanks!

### DIFF
--- a/Source/NSLocale.m
+++ b/Source/NSLocale.m
@@ -536,7 +536,7 @@ static NSRecursiveLock *classLock = nil;
   if (nil == systemLocale)
     {
 #if	GS_USE_ICU == 1
-      systemLocale = [[NSLocale alloc] initWithLocaleIdentifier: @""];
+      systemLocale = [[NSLocale alloc] initWithLocaleIdentifier: nil];
 #endif
     }
 


rfm:
Looking at the NSLocale documentation for the SystemLocale method:

systemLocale
A locale representing the generic root values with little localization.
i
Return Value
The generic locale that contains fixed “backstop” settings that provide values for otherwise undefined keys.

Discussion
Use the system locale when you don’t want any localizations. If you want localizations that match the user's region settings, use the locale given by the [currentLocale]

Now, the current behavior (passing an empty string) uses the posix locale, which is the correct one for the documentation behavior of the method.

sbeard-eh:
Unfortunately, the isseu that I am seeing is that passing an empty string with ICU version 67.0 returns an empty locale rather than the posix locale.

rfm:
It seems at some point ICU stopped interpreting an empty string as the C/POSIX locale.
Looking at their current documentation at https://unicode-org.github.io/icu/userguide/locale/#locales-and-services it seems they now use/require en_US_POSIX rather than an empty string, so we need to check the ICU version and use one or the other.

sbeard-eh:
Ok I will close this and have a look at what needs to be done to implement the change you have suggested.

PR #281, opened 2022-11-01
Fix for locale handling changes in ICU 64

sbeard-eh:
Hi,

Since ICU 64, the locale handling has changed such that passing an empty string to -initWithLocaleIdentifier: (and consequently to uloc_canonicalize) no longer returns the en_US_POSIX locale as it did in previous versions. Passing "en_US_POSIX" instead maintains consistency with how + (id) systemLocale worked previously.

See https://icu.unicode.org/download/64#h.plg55ia6o3du

Thanks!


rfm:
Merged ... thanks.

PR #291, opened 2023-04-06
`[NSData initWithBase64EncodedString]`: Fix decoding of an empty string

qmfrederik:
Don't call `NSZoneRealloc` with a length of 0, but free the zone and return an empty `NSData` buffer instead.

This pattern was observed when unarchiving Xib files.

### DIFF
--- a/Source/NSData.m
+++ b/Source/NSData.m
@@ -804,6 +804,13 @@ failure:
 	}
     }
   length = dst - result;
+
+  if (length == 0)
+    {
+      NSZoneFree(zone, result);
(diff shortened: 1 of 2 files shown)


rfm (commented):
Good spot.
Event better to have a testcase :-)

But the return statement leaks the receiver and returns an instance potentially of the wrong class.

I think instead of
  return [[NSData alloc] init];
it should be
    return [self initWithBytesNoCopy: 0 length: 0 freeWhenDone: YES];

qmfrederik:
Thanks @rfm , I've updated my PR.

rfm:
Thanks ... merged.

PR #293, opened 2023-04-26
Fix -[NSFileManager contentsOfDirectoryAtURL:...]

triplef:
Was returning URLs relative to current working directory instead of given directory.


triplef:
Looks like httpbin.org is having some network issues and some tests are failing due to that.

rfm (approved):
Looks good to me.  Thanks.

PR #297, opened 2023-05-04
NSOrderedSet: Fix keyed encoding to match Apple platforms

triplef:
We found that the code marked as follows does not match the actual behavior on macOS:

> HACK ... MacOS-X seems to code differently if the coder is an actual instance of NSKeyedArchiver

This seems to have been copy-pasted from the NSSet implementation, but does not apply to NSOrderedSet, making NSOrderedSet instances encoded with GNUstep incompatible with Apple platforms.


rfm (approved):
Thanks, we definitely want to match Apple's archive formats as perfectly as we can.

PR #298, opened 2023-05-05
Add stub methods for secure unarchiving

triplef:
Adds NSKeyedUnarchiver methods that were added in macOS 10.13 and 11.0 for secure unarchiving as stub methods forwarding to `-unarchiveObjectWithData:` (i.e. no actual secure unarchiving is done).


Headers/GNUstepBase/GSVersionMacros.h
```diff
@@ -66,6 +66,11 @@
 #define MAC_OS_X_VERSION_10_12	1120
 #define MAC_OS_X_VERSION_10_13	1130
 #define MAC_OS_X_VERSION_10_14	1140
+#define MAC_OS_X_VERSION_10_15	101500
```
triplef, line 69: Ahh I just realized that the GS defines are just using 4 digits, whereas on Apple platforms they are using 6 starting with 10.10 (i.e. `101000`). Not sure what’s best to do here.
- I guess editing old defines might cause issues although it might also be better to be in sync with Apple platforms?
- Alternatively we could keep 4 digits up to 10.16 and use 6 digits starting with 11.0?

rfm (commented):
I think it's a mistake that we are only using 4 digits;  looking at the autogsdoc code for working with versions (which appears to be the only place in the base library where numeric values are used explicitly) I find that it is expecting the six digit values.
So I think a fixup to use six makes sense.

triplef:
Sounds good, I’ve updated the defines to match Apple headers.

rfm (approved):
Thanks ... looks good.

PR #309, opened 2023-07-26
Fix NSData initWithContentsOfURL: caching data of file URLs

triplef:
Initializing NSData with a file URL was using `-[NSURL resourceDataUsingCache:]` to load the file, which as a side effect caused the file data to be cached in memory. This is probably not desired behavior and caused very large memory growths in our application, and doesn’t match the behavior on Apple platforms.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,10 @@
 2023-07-26 Frederik Seiffert <frederik@algoriddim.com>
 
+	* Source/NSData.m:
+	Fix NSData initWithContentsOfURL: caching data of file URLs.
+
+2023-07-26 Frederik Seiffert <frederik@algoriddim.com>
+
 	* Source/NSURL.m:
 	* Tests/base/NSURL/basic.m:
 	Fix NSURL path on Windows for UNC paths.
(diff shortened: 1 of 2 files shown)


triplef:
Thinking about this some more I'm wondering if we should generally disable caching of file URLs in the NSURL resource loading system?

That being said I think this change still makes sense to reduce overhead when loading file URLs.

rfm (approved):
That looks OK.  Generally it makes sense to implement Apple's behavior.
With regards to the caching of file URLs more generally, I think you are probably right that we shouldn't do it. Generally a modern O/S will cache disk reads, so re-reading the same file is likely to be quite quick, and if a specialised app really needs to cache information in its own memory to avoid system call overheads, it's probably going to want to control that cache itself (I know that when optimising I want to control my own cache behaviors).

triplef:
Thanks. Do you know where the NSURL caching behavior is controlled / what we’d need to change to disable file caching?

rfm:
It's in Source/NSURLHandle.m ... the implementation of [GSFileURLHandle +cachedHandleForURL:] appears (at a glance) to cache things forever.
That's obviously a bad thing to do, and was probably intended to be changed but forgotten about.

PR #310, opened 2023-08-10
Change NSNumberFormatter minimum/maximum properties type

triplef:
Not sure why they were defined as NSDecimalNumber, but in current Apple headers they are [defined as NSNumber](https://developer.apple.com/documentation/foundation/nsnumberformatter/1417228-minimum).

As far as I can tell these properties are not currently used in any implementations.


rfm (approved):
The change looks fine to me.

PR #313, opened 2023-08-18
Revert "Initial commit of updated IPv6"

rfm:
Reverts gnustep/libs-base#307
Oops ...I was wrong to do that.

### DIFF
--- a/Source/Additions/NSURL+GNUstepBase.m
+++ b/Source/Additions/NSURL+GNUstepBase.m
@@ -87,7 +87,7 @@
   if ([host length] > 0)
     {
       [urlString appendString:
-	[host stringByAddingPercentEncodingWithAllowedCharacters: [NSCharacterSet URLHostAllowedCharacterSet]]];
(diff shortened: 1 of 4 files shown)


rfm:
This change introduces a *lot* of compiler warnings, and I wonder if it actually does anything useful;  the original code already has struct socket_any which is a union containing sockaddr_in6 (the IPV6 socket address) and is therefore both correctly aligned and correctly sized.

gcasa:
@RoyalStewart90 Can you address the questions @rfm has raised?

rfm:
I reverted that and added the bit that appears to be functional in Source/win32/NSStream.m
When we can't create IPV4 streams, try to create IPV6:

 if (!ins)
    {
#if     defined(PF_INET6)
      ins = AUTORELEASE([[GSInet6InputStream alloc]
        initToAddr: address port: port]);
      outs = AUTORELEASE([[GSInet6OutputStream alloc]
        initToAddr: address port: port]);
#endif
    }

PR #314, opened 2023-08-19
Try to address issue of treating pointer function values as bitfield

fredkiefer:
This does not fix all the issues that were reported. At least it tries to sort out the bitfield access to the memory and personality options on pointer functions.

### DIFF
--- a/Source/NSConcreteHashTable.m
+++ b/Source/NSConcreteHashTable.m
@@ -77,7 +77,7 @@ typedef GSIMapNode_t *GSIMapNode;
 #define	GSI_MAP_TABLE_S	instanceSize
   
 #define IS_WEAK(M) \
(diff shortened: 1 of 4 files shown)


rfm (commented):
This all looks very reasonable.
A couple of pedantic suggestions though:
The definition of IS_WEAK should really use || (logical or) rather than | (bitwise)
The memoryType() and personalityType() functions should really be of the form
return ((options & mask) == flag) ? YES : NO;

I don't think those change the way the code behaves, but it would make it clearer than we are working with booleans.

Also, I don't use any of this pointer function code myself ... so I wonder if we should be adding testcases to exercise the various possibilities (and make sure we don't break anything).

fredkiefer:
I addressed you point and yes, having a few tests for this functionality would surely help.

rfm (commented):
Looks fine to me.

Issue #316, opened 2023-08-30
NSLock throws an exception when trying to unlock when not previously locked...


gcasa:
Issue: NSLock throws an exception when trying to unlock when not previously locked...

Steps:
1) Create an NSLock / NSConditionLock
2) Try to unlock the instance without previously locking it.

Result: An exception is thrown, as illustrated by the attached image
Expected: This should either fail silently or log a warning...
![NSConditionLock_error](https://github.com/gnustep/libs-base/assets/27943/987cebcb-ff55-451f-9bcb-9a1eb11fae13)

gcasa:
The test is here... https://github.com/gcasa/NSConditionLock_test

hmelder:
What is the behaviour on macOS? I would consider unlocking before locking a programmer error.

gcasa:
> What is the behaviour on macOS? I would consider unlocking before locking a programmer error.

The behavior on macOS is to simply ignore it.  I think out behavior is better but we should allow it to be configurable so a default was added. 

If merged a pr related to this with code from @rfm so I am closing this.

PR #317, opened 2023-08-30
Issue a warning instead of throwing an exception

gcasa:
This is a basic solution to issue #316,  it is likely not the right one, but I am hoping this PR provides the right forum for discussing what is correct.   It is worth noting that macOS does not throw an exception when an unlock fails.

### DIFF
--- a/Source/NSLock.m
+++ b/Source/NSLock.m
@@ -63,7 +63,6 @@ static Class    untracedLockClass = Nil;
 static Class    untracedRecursiveLockClass = Nil;
 
 static BOOL     traceLocks = NO;
-static BOOL     compatible = NO;
 
 @implementation NSObject (GSTraceLocks)
 
@@ -250,10 +249,15 @@ static BOOL     compatible = NO;
 {\
   if (0 != GS_MUTEX_UNLOCK(_mutex))\
     {\
-      if (compatible)\
+      if (GSPrivateDefaultsFlag(GSMacOSXCompatible))\
+	{\
+          NSLog(@"Failed to unlock mutex %@ at %@",\
+	    self, [NSThread callStackSymbols]);\
+	}\
+      else \
 	{\
           [NSException raise: NSLockException\
-  	        format: @"failed to unlock mutex"];\
+		      format: @"failed to unlock mutex %@", self];\
 	}\
     }\
   CHK(Drop) \
@@ -349,7 +353,6 @@ NSString *NSLockException = @"NSLockException";
       untracedConditionLockClass = [GSUntracedConditionLock class];
       untracedLockClass = [GSUntracedLock class];
       untracedRecursiveLockClass = [GSUntracedRecursiveLock class];
-      compatible = (strcmp(getenv("NSLOCK_COMPATIBLE"), "YES") == 0) ? YES : NO;
     }
 }
 
--- a/Source/NSLock.m
+++ b/Source/NSLock.m
@@ -42,6 +42,7 @@
 #import "Foundation/NSLock.h"
 #import "Foundation/NSException.h"
 #import "Foundation/NSThread.h"
+#import "Foundation/NSUserDefaults.h"
 
 #define class_createInstance(C,E) NSAllocateObject(C,E,NSDefaultMallocZone())
 
@@ -61,6 +62,7 @@ static Class    untracedLockClass = Nil;
 static Class    untracedRecursiveLockClass = Nil;
 
 static BOOL     traceLocks = NO;
+static BOOL     compatible = NO;
 
 @implementation NSObject (GSTraceLocks)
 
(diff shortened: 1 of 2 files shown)


rfm:
From the point of a developer, the existing code is much better than the Apple-like version.  This is dealing with a (potentially critical) bug in the application locking logic, and currently we get the program halted with a log of (assuming stack traces are enabled) exactly where the problem occurred.  Throwing all that away in favour of a minimal warning log only makes sense in  production code where the developer can't do anything about the problem, and you can hope that continuing might mean the app works (rather than failing in a worse way than just crashing).

So, there are a variety of options:
1. Make the alternatives conditional at runtime using the GSMacOSXCompatible setting, which is provided to turn on OSX behaviors in case where we think they are poor.
2. Make the alternatives conditional at compile time, with the minimal warning if NDEBUG is defined
3. If the aim is to have the program continue after the error, print the stack trace and the name of the lock as part of the warning.

triplef:
I also strongly agree that this shouldn’t just be a log by default.

Would NSAssert() make sense? That way users can disable these assertions in production code using `NS_BLOCK_ASSERTIONS`.

That being said, always throwing an exception seems the right thing to do to me, because you will always want to fix your locking logic. @gcasa what’s the underlying issue you are trying to solve here?

gcasa:
The issue is that I am workin on some production code where there is an attempt to unlock an NSLock which is either unlocked or has never been locked.  I have tested this on macOS and there it fails silently.

I have always maintained that we should always recover gracefully whenever possible since that makes it easier for the end users.  I like the suggestion of using the  GSMacOSXCompatible setting.

rfm:
> we should always recover gracefully whenever possible
That's what the exception lets us do:  the higher level code can catch the exception, decide whether graceful recovery is possible, and act accordingly.  This is why the current behavior is better than the Apple behavior.

Consider this:
If the code attempts to unlock a lock which is not held, there is a good chance that it's because we failed to lock a region where locking was expected.
If locking was expected, its because we needed to prevent in-memory data structures from being modified by another thread.
So finding the lock not held means theres definitely a chance that in-memory data structures have been corrupted.
If we just log a warning, we are likely to use corrupt data and either
a. have the application behave wrongly or crash or
b. write the corrupt data to disk, so that the problem spreads to other systems

Admittedly I'm biased by the fact that I write server processes which are expected to run 24*7 handling millions of transactions a day (and which are monitored so that crashed processes are automatically restarted and  hung/crashed processes raise alarms with sysadmins), so it's particularly important that the software doesn't misbehave and corrupt data.

gcasa:
I concede what you're saying.  It is better to throw the exception because it forces the developer to clean up the locking logic.  I think the idea of an option to allow less "strict" behavior is a good idea.  

From a devs standpoint the current behavior is desirable as it gives the opportunity to fix the problem.  I am, in fact, dealing with that situation now, but there is also a product to release. It's always a balance.

rfm:
Sure.  I'd recommend using GSMacOSXCompatible (default is off) to change it to a warning *and* improving the warning to (at least) report the lock description (which will include its name if it has one set).

gcasa:
I have added the default suggested in just unlock.  I am not sure if it should be applied elsewhere.

gcasa:
Apparently, any attempt to access NSUserDefaults from within NSLock fails with a segfault.   I believe this is because NSUserDefaults is not initialized at the point where NSLock is called.   I will try to use an environment variable instead.

gcasa:
Same with NSProcessInfo... ;/

gcasa:
I get a segfault also if I try to do getenv.   I don't know what else to try at this point. :/ @rfm  Any suggestions?

gcasa:
I will commit what I have tried so you can see.

rfm:
> On 11 Sep 2023, at 16:00, Gregory Casamento ***@***.***> wrote:
> 
> 
> I will commit what I have tried so you can see.

I accidentally (lack of git familiarity) put fixes and new test cases in origin/NSLock_fix_issue316 rather than NSLock_fix_issue316

Your basic problem is trying to do stuff in the +initialise method of NSLock ... a test is best done lazily (ie later at the point when it's needed) because there's nothing to be gained by trying to cache information for an event which only ever happens in response to a bug.

gcasa:
> 
> 
> > On 11 Sep 2023, at 16:00, Gregory Casamento ***@***.***> wrote:
> > 
> > 
> > I will commit what I have tried so you can see.
> 
> I accidentally (lack of git familiarity) put fixes and new test cases in origin/NSLock_fix_issue316 rather than NSLock_fix_issue316
> 
> Your basic problem is trying to do stuff in the +initialise method of NSLock ... a test is best done lazily (ie later at the point when it's needed) because there's nothing to be gained by trying to cache information for an event which only ever happens in response to a bug.

I actually tried it in init and in the actual unlock method and got a segfault in both places as well.   I will try a couple of other things.

gcasa:
Cherry-picked the two changes @rfm made on his alternative branch.

gcasa:
@rfm let me know if we can merge since this is, essentially, your change.  GC

gcasa:
@rfm Reran the test case that failed and it passes.  I am merging this based on your previous comment.

PR #320, opened 2023-09-09
INSTALL: Add libavahi and zlib as requirements

hmelder:
This is only a small patch that lifts libxslt, which is used in NSXMLDocument for xslt parsing, from OPTIONAL to RECOMMENDED, as a gnustep-base installation without libxslt would introduce inconsistency. The method objectByApplyingXSLT:arguments:error: (Available since Mac OS X 10.4) would just return nil if gnustep-base was compiled without libxslt support.

Additionally, libavahi for mDNS discovery, used in NSNetService (https://developer.apple.com/documentation/foundation/nsnetservice), and zlib, used in GSFileHandle for file compression should be listed and marked as RECOMMENDED.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -18,7 +18,9 @@ higher.
    * libxml2 (RECOMMENDED)
    * libcurl (RECOMMENDED)
    * libdispatch (RECOMMENDED)
-   * libxslt (OPTIONAL)
+   * libavahi (RECOMMENDED for NSNetServices)
+   * libxslt (RECOMMENDED)
+   * zlib (RECOMMENDED)
    * iconv (OPTIONAL, not needed if you have glibc)
    * openssl (OPTIONAL, not needed if you have gnutls)
 


INSTALL
```diff
@@ -18,7 +18,9 @@ higher.
    * libxml2 (RECOMMENDED)
    * libcurl (RECOMMENDED)
    * libdispatch (RECOMMENDED)
-   * libxslt (OPTIONAL)
+   * libavahi (RECOMMENDED for NSNetServices)
+   * libxslt (RECOMMENDED)
```
znek, line 22: Why was this changed from `OPTIONAL` to `RECOMMENDED`?

hmelder: GNUstep extensions are using zlib for compression support, and OPTIONAL
was used in a context where there are different options (i.e. gnutls and openssl, or iconv and glibc).

Having an incomplete installation is just wrong imo.

hmelder: NSXMLDocument is using libxslt for methods like objectByApplyingXSLT:arguments:error: and return nil if gnustep-base
was compiled without libxslt (See https://github.com/gnustep/libs-base/blob/5f92c87d784753c996e14e4c75af7464014b6833/Source/NSXMLDocument.m#L429C8-L429C28).

Having an incomplete installation is just wrong imo.

OPTIONAL is used in a context where there are different options (i.e. gnutls and openssl, or iconv and glibc).

znek: I see, this is public API and available since macOS 10.4.

hmelder: I think we generally need to improve toolchain consistency by enforcing certain libraries during the configuration phase.

rfm (approved):
Quite right.  I believe xslt was optional back before Apple added XML support (it was used only for the GNUstep specific XML extensions, so considered optional) and INSTALL just never got updated.

Issue #322, opened 2023-09-13
removeObserver:forKeyPath:context: not implemented


gcasa:
Calling this method throws an exception.

gcasa:
https://github.com/gnustep/libs-base/pull/323

gcasa:
Closing as pr was merged

PR #323, opened 2023-09-13
Add removeObserver:forKeyPath:context: to implementation

gcasa:
Addresses issue #322.

### DIFF
--- a/Headers/Foundation/NSKeyValueObserving.h
+++ b/Headers/Foundation/NSKeyValueObserving.h
@@ -138,6 +138,10 @@ GS_EXPORT NSString *const NSKeyValueChangeNotificationIsPriorKey;
 - (void) removeObserver: (NSObject*)anObserver
 	     forKeyPath: (NSString*)aPath;
 
+- (void) removeObserver: (NSObject*)anObserver
+	     forKeyPath: (NSString*)aPath
+                context: (void*)context;
+
 @end
 
 @interface NSArray (NSKeyValueObserverRegistration)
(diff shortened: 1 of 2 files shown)


rfm (commented):
The interface looks fine.

I think we have two options for the implementation:
1. raise an exception to say that the method is not yet implemented.
2. implement the method to do what it's supposed to.

Implementing it as a synonym for -removeObjecter:forKeyPath: is IMO wrong since that would mean the app silently misbehaves.

gcasa:
> The interface looks fine.
> 
> I think we have two options for the implementation:
> 
> 1. raise an exception to say that the method is not yet implemented.
> 2. implement the method to do what it's supposed to.
> 
> Implementing it as a synonym for -removeObjecter:forKeyPath: is IMO wrong since that would mean the app silently misbehaves.

I am not sure what the proper implementation is.  I don't know what should be done with the "context" parameter.

rfm:
I'm not familaiar with it either, but it looks from the documentation as if -removeObjecter:forKeyPath: should remove all observers, but -removeObjecter:forKeyPath:context: should only remove the observer which was added using the specified context (an opaque pointer).

gcasa:
Since the context pointer is opaque I believe this is something that is left to the user to implement.   I am going to do a bit more research on this.  For now, I have made it so that the method throws an exception.

triplef:
The `context` pointer is in fact not usable in GNUstep‘s current KVO implementation. I’ve opened #324 with some more details.

rfm (approved):
That looks plausible to me.  If only we had testcases :-(

gcasa:
> That looks plausible to me.  If only we had testcases :-(

I will add some as a separate PR. I will develop them on macOS.

PR #326, opened 2023-09-18
Fix of various warnings

rmottola:


rfm (approved):
Looks good to me ... I don't know if OSX raises exceptions on truncate failure like the new code, but it seems the right thing to do.

gcasa:
The documentation can be read that way so that's how @rmottola and I implemented it.

PR #330, opened 2023-10-01
Add removeObserver:fromObjectsAtIndexes:forKeyPath:context: method

gcasa:
Add new methods and some tests for KVO.

### DIFF
--- a/Source/NSKeyValueObserving.m
+++ b/Source/NSKeyValueObserving.m
@@ -1631,13 +1631,6 @@ cifframe_callback(ffi_cif *cif, void *retp, void **args, void *user)
   [kvoLock unlock];
 }
 
-- (void) removeObserver: (NSObject*)anObserver
-   fromObjectsAtIndexes: (NSIndexSet *)indexes
-             forKeyPath: (NSString*)aPath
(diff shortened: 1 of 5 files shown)


Source/NSKeyValueObserving.m
```diff
@@ -1325,6 +1325,13 @@ - (void) removeObserver: (NSObject*)anObserver
   [iLock unlock];  
 }
 
+- (void) removeObserver: (NSObject*)anObserver
```
fredkiefer, line 1328: Why do we have this method here? We don't define the method without context neither and we also don't have the corresponding add methods.

gcasa: It doesn't belong there...  I discovered the method should only exist where I have it implemented now, as a category on NSArray.

gcasa:
@fredkiefer @rfm The tests are passing now. One of the url tests seems to fail intermittently.

gcasa:
@rfm @fredkiefer any added feedback?

rfm (approved):
This looks fine to me.

gcasa:
> This looks fine to me.

Thanks. :)

PR #331, opened 2023-10-07
Win64: Only use `__builtin_setjmp`/`__builtin_longjmp` when compiling with GCC

qmfrederik:
Using `__builtin_setjmp` when compiling with Clang on MSYS2 will result in compiler warnings and runtime crashes.  It seems to work on GCC, though.

Compiler warning on Clang:

```
warning: incompatible pointer types passing 'jmp_buf' (aka 'struct _SETJMP_FLOAT128[16]') to parameter of type 'void **' [-Wincompatible-pointer-types]
                NS_DURING
                ^~~~~~~~~
note: expanded from macro 'NS_DURING'
                    if (!setjmp(NSLocalHandler.jumpState)) {
                                ^~~~~~~~~~~~~~~~~~~~~~~~
note: expanded from macro 'setjmp'
#define setjmp(X)       __builtin_setjmp(X)
                                         ^
1 warning generated.
```

### DIFF
--- a/Headers/Foundation/NSException.h
+++ b/Headers/Foundation/NSException.h
@@ -54,17 +54,6 @@
 #include <setjmp.h>
 #include <stdarg.h>
 
-#if	defined(__WIN64__)
-/* This hack is to deal with the fact that currently (June 2016) the
- * implementation of longjmp in mingw-w64  sometimes crashes in msvcrt.dll
- * but the builtin version provided by gcc seems to work.
- */
-#undef	setjmp
-#define	setjmp(X)	__builtin_setjmp(X)
-#undef	longjmp
-#define	longjmp(X,Y)	__builtin_longjmp(X,Y)
-#endif
-
 #if	defined(__cplusplus)
 extern "C" {
 #endif


qmfrederik:
Actually, I think the best approach may be to remove this alltogether? `setjmp` and `__builtin_setjmp` have slightly different semantics. Using `__builtin_setjmp` as a replacement for `setjmp` apparently breaks interoperability between GNUstep compiled using gcc and other software compiled with clang.

rfm:
Yes, I see this is a hack specifically for the state of mingw64 as it was in 2016 ...  so I agree that if current mingw64 is working without the hack we should simply remove it.

qmfrederik:
@rfm I compiled libs-base on the UCRT64 environment of MSYS2, using gcc as the compiler, and using the  'traditional'/setjmp-based exceptions (e.g. without `-fexceptions` and `fojbc-exceptions`).  All tests (`make check`) passed.  I ran the test suite of one of our applications, which passed as well.

That seems to be an indication that it's OK to remove this workaround, 7 years after it was introduced.

PR #333, opened 2023-10-08
Add missing includes for 

qmfrederik:
`NSObject.m` and `NSProxy.m` both reference `objc_msg_lookup`, but don't include `objc/message.h`.  This may cause an undefined reference to `objc_msg_lookup` in some configurations.

### DIFF
--- a/Source/NSObject.m
+++ b/Source/NSObject.m
@@ -35,6 +35,7 @@
 
 #import "common.h"
 #include <objc/Protocol.h>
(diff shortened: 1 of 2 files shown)


rfm (approved):
I slightly worry that on some systems objc/message.h might not exist.
However, I think one dile already includes it without any check to see if it exists, so this change shouldn't break anything that's not already broken.

PR #334, opened 2023-10-09
Fix compiler warning when compling NSFontAssetRequest.m in libs-gui

qmfrederik:
Cast the result type of CALL_NON_NULL_BLOCK to void to satisfy the ? operator.

If not, compiling on clang on msys2/ucrt64 fails with this error:

```
NSFontAssetRequest.m:48:3: error: incompatible operand types ('int' and 'id')
    48 |   CALL_BLOCK(completionHandler, error);
        |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/tools/msys64/ucrt64/include/GNUstepBase/GSBlocks.h:110:53: note: expanded from macro 'CALL_BLOCK'
    110 | #define CALL_BLOCK(block, args...) ((NULL != block) ? CALL_NON_NULL_BLOCK(block, args) : nil)
        |                                                     ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~   ~~~
1 error generated.
```

### DIFF
--- a/Headers/GNUstepBase/GSBlocks.h
+++ b/Headers/GNUstepBase/GSBlocks.h
@@ -107,8 +107,8 @@ typedef retTy(^name)()
 
 #endif /* __has_feature(blocks) */
 
-#define CALL_BLOCK(block, args...) ((NULL != block) ? CALL_NON_NULL_BLOCK(block, args) : nil)
-#define CALL_BLOCK_NO_ARGS(block) ((NULL != block) ? CALL_NON_NULL_BLOCK_NO_ARGS(block) : nil)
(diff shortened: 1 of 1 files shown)


rfm:
I don't think that can be correct, because the result of the call can be assigned (which is impossible for void).
Of course this is designed to be used where we don't have real blocks, and to support a subset of the possible return types.
So presumably we want to cast to 'id' rather than 'void' (or perhaps cast both sides to 'void*')?

rfm (approved):
Thanks for the update.

rfm:
Oops, failed to notice that the compiler objects.  I wonder why?

rfm:
In th absence of feedback, I went ahead with that last proposal.

PR #336, opened 2023-10-10
Update... fix respondsToSelector:

gcasa:
Correct interpretation of GSMacOSXCompatible in respondsToSelector:

### DIFF
--- a/Source/NSObject.m
+++ b/Source/NSObject.m
@@ -2104,7 +2104,7 @@ static id gs_weak_load(id obj)
 
   if (aSelector == 0)
     {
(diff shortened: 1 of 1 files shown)


gcasa:
I am assuming the original intent here was to make our handling of this more strict.   I, honestly, have my questions about how this is being handled as, in general, a nil selector or a nil object are usually handled as a NO-OP by Objective-C.

gcasa:
@rfm let me know if it looks okay to merge.

rfm (approved):
Yes, that seems fine.

PR #337, opened 2023-10-11
NSDistributedNotificationCenter: Throw a descriptive exception when gdnc could not be found

qmfrederik:
`NSTask launchPathForTool: @"gdnc"];` can return `nil` when the tool is not found.  In this case, `[NSTask launchedTaskWithLaunchPath]` would raise a `NSInvalidArgumentException` with error message "NSTask - no launch path set" which is not very descriptive.


qmfrederik:
This one bit me a couple of times, and it took me some time to figure out what the `NSTask - no launch path set` error message in my log files was referring to.  Hopefully generating a better error message will save some folks some time in the future ;-)

rfm (approved):
Looks good.  I thought I'd already said so, but I guess I must have failed to hit the submit button.

PR #344, opened 2023-11-11
NSObject: Change return type of setVersion: from id to void

hmelder:
I have check libs-base and found no code that expects a return value from setVersion:

As the GNUstep implementation diverges from Foundation, we treat this as a bug.


rfm (approved):
Looks fine.  Thanks.

hmelder:
CI error is unrelated error:

Building base/NSNumber/test01.m
make  test01
Making all for test_tool test01...
 Compiling file test01.m ...
test01.m:28:38: error: use of undeclared identifier 'M_PI'
      n = [NSNumber numberWithFloat: M_PI];

PR #346, opened 2023-11-14
Support deleting NSMutableURLRequest headers

triplef:
Calling `setValue:forHTTPHeaderField:` with a nil value deletes the header on Apple platforms.


rfm (commented):
That looks good, but please could you also update the documentation (NSURLRequst.h) to note that setting a nil value removes the header.

triplef:
Yes good point, I updated the documentation.

PR #353, opened 2023-11-26
Update IPv6 implementation from keysight

gcasa:


rfm (changes requested):
Almost all this change is wrong and needs to be removed; certainly everything messing with the sockaddr layout (whichseems to simply make the code messy by changing it to use an inappropriate type and then adding casts back to the correct type) and the non-functional testcase.

I think that leaves the one line change in Source/Additions/NSURL+GNUstepBase.m which looks correct.

Did I miss anything else?

gcasa:
> Almost all this change is wrong and needs to be removed; certainly everything messing with the sockaddr layout (whichseems to simply make the code messy by changing it to use an inappropriate type and then adding casts back to the correct type) and the non-functional testcase.
> 

It was my understanding that sockaddr_storage was large enough to work with anything that uses sockaddr.  

The test-case is a placeholder for now.  The intent is to make it test reading and writing from a stream. 

> I think that leaves the one line change in Source/Additions/NSURL+GNUstepBase.m which looks correct.

Ok. 

> Did I miss anything else?

No, I don't think so. Honestly I am not sure the changes are actually needed.  These are based on the changes done by royal @ keysight.  

Does the existing code in base (without this change) handle IPv6 properly?  I have been asking them about failure modes with the existing code on master.

What I put here is only a starting point based on the work done by them.  I'm trying to figure out if it is needed at all.  If it's not I would rather not incorporate it.

rfm:
> On 26 Nov 2023, at 09:31, Gregory Casamento ***@***.***> wrote:
> 
> 
> No, I don't think so. Honestly I am not sure the changes are actually needed. These are based on the changes done by royal @ keysight.

I think the guy at keysight simply got the sockaddr stuff wrong.  Possibly they misdiagnosed a problem which was fixed by some other change elsewhere.

> Does the existing code in base (without this change) handle IPv6 properly?

As far as I know, yes.  Certainly the simple testcase I added to the test suite works (transferring a file over a loopback network connection).

gcasa:
> 
> 
> > On 26 Nov 2023, at 09:31, Gregory Casamento ***@***.***> wrote:
> > 
> > 
> > No, I don't think so. Honestly I am not sure the changes are actually needed. These are based on the changes done by royal @ keysight.
> 
> I think the guy at keysight simply got the sockaddr stuff wrong.  Possibly they misdiagnosed a problem which was fixed by some other change elsewhere.
> 
> > Does the existing code in base (without this change) handle IPv6 properly?
> 
> As far as I know, yes.  Certainly the simple testcase I added to the test suite works (transferring a file over a loopback network connection).
> 
> 
> 

As I remember the concern was over memory management.  I believe they were seeing a crash after a connection was closed.  The change was meant to address that issue I believe but I am not sure the current set of changes actually does that.

rfm:
There's nothing in the changes to deal with overwriting memory after a connection is closed: that would be a sequencing error (writing to memory after it's freed) rather than anything to do with the size of the buffer (which their patch doesn't actually seem to change anyway ... the union was already big enough for all possible socket addresses).

PR #354, opened 2023-11-29
Simplified version of IPv6 changes, this is investigational and a WIP

gcasa:
I need this on a branch at the moment to test with some functionality at keysight.   Adding it here as a WIP,  this is NOT ready for review.

### DIFF
--- a/Source/GSSocketStream.m
+++ b/Source/GSSocketStream.m
@@ -1849,7 +1849,7 @@ setNonBlocking(SOCKET fd)
           [GSTLSHandler tryInput: self output: _sibling];
         }
 
-      result = connect([self _sock], &_address.s,
+      result = connect([self _sock], (struct sockaddr *)&_address.s,
         GSPrivateSockaddrLength(&_address.s));
       if (socketError(result))
         {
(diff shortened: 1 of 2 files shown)


rfm (commented):
All the changes in GSSocketStream.m do is add explicit casts to say that variables are of the type that the compiler already knows they are.
That's not actually harmful because it is a no-op, but it's bad style to add unnecessary casts, and makes the code a little harder to read.

The change in win32/NSStream.m is basically already in master (so would probably conflict with the existing code).  The difference is that, in master it's protected by checking for PF_INET6 being defined rather than AF_INET6, but I think on any system where one is defined both are defined.
If the code in master was wrong on this point (ie PF_INET6 was undefined) inet6 streams could not work at all on windows, but the automated regression tests are passing on windows systems so we know they do.

gcasa:
> All the changes in GSSocketStream.m do is add explicit casts to say that variables are of the type that the compiler already knows they are. That's not actually harmful because it is a no-op, but it's bad style to add unnecessary casts, and makes the code a little harder to read.
> 
> The change in win32/NSStream.m is basically already in master (so would probably conflict with the existing code). The difference is that, in master it's protected by checking for PF_INET6 being defined rather than AF_INET6, but I think on any system where one is defined both are defined. If the code in master was wrong on this point (ie PF_INET6 was undefined) inet6 streams could not work at all on windows, but the automated regression tests are passing on windows systems so we know they do.

I already have a strong suspicion that the issue is with the application and not with GS code.   I should know more over the next few days.

gcasa:
Summarily closing this issue as it is invalid and outdated.

PR #358, opened 2023-12-09
Correctly handle KVC interaction with NSNull

hmelder:
This PR addresses the handling of undefined keys in NSNull by implementing the valueForUndefinedKey: method. By doing so, we do not end up invoking the default handler which raises an exception.

Included in this PR are several unit tests. These tests are designed to verify the correct functioning of various KVC methods when invoked on [NSNull null].


Tests/base/NSPredicate/nsnull.m
```diff
@@ -0,0 +1,34 @@
...
+#import "Foundation/NSArray.h"
+#import "Foundation/NSNull.h"
+
+int main(void) {
+    NSAutoreleasePool *arp = [NSAutoreleasePool new];
+	NSArray *array, *filtered;
+	NSPredicate *predicate;
+
+    // Basic filtering with NSPredicate
+    array = @[@{@"key": @"value1"}, @{@"key": @"value2"}, [NSNull null]];
```
fredkiefer, line 13: This notation will break with gcc based compilation.

Tests/base/KVC/nsnull.m
```diff
@@ -0,0 +1,31 @@
...
+    id result = [nullObject valueForKey:@"undefinedKey"];
+    PASS(result == nullObject, "NSNull returns itself for undefined keys.");
+
+    // Attempting to set a value for an undefined key
+    PASS_EXCEPTION([nullObject setValue:@"value" forKey:@"undefinedKey"],
+	          NSUndefinedKeyException,
+              "Setting an undefined key on NSNull should not crash.");
+
+    // Accessing an undefined key path
+    id result = [nullObject valueForKeyPath:@"some.path"];
```
fredkiefer, line 21: Here gcc complains that you are redefining `result` and I think this complain is correct.

hmelder: You are right. I tested something and forgot to change it back :)

Tests/base/NSPredicate/nsnull.m
hmelder: Fixed

hmelder: Seems like this is also not allowed:

```sh
nsnull.m: In function ‘main’:
nsnull.m:21:45: error: array subscript is not an integer
   21 |     PASS(filtered.count == 1 && [filtered[0][@"key"] isEqualToString:@"value2"],
   ```

fredkiefer (approved):
Looks good to me now. Please rebase and merge.

PR #362, opened 2024-01-08
Fix memory handling of NSNotificationCenter observer object

triplef:
Fixes issue discussed in https://github.com/gnustep/libs-base/commit/db90ab6bb9e65c6076308234d4e1e21586cbde17.

### DIFF
--- a/Source/NSNotificationCenter.m
+++ b/Source/NSNotificationCenter.m
@@ -869,7 +869,7 @@ static NSNotificationCenter *default_center = nil;
 	       name: name 
 	     object: object];
 
-  return AUTORELEASE(observer);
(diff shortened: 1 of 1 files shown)


rfm (commented):
I think that's OK, but I have just written a more efficient version I was about to commit.
Almost the same code, but caching the class in a static variable

static Class    GSNotificationObserverClass = Nil;
+ (void) initialize
{
  if ([GSNotificationObserver class] == self)
    {
      GSNotificationObserverClass = self;
    }
}
and making the check be

  if (object_getClass(observer) == GSNotificationObserverClass)
    {
      RELEASE(observer);
    }

triplef:
Ah nice, thank you! I’ll just wait for your commit then.

PR #363, opened 2024-01-12
MinGW: Add dllimport/dllexport attributes when compiling with clang

qmfrederik:
The GCC compiler will automatically export all symbols on MinGW, but this is not the case for clang.  Add the `__declspec` attributes when compiling with clang on MinGW.


rfm (approved):
I see no problem with this, though perhaps it might be worth adjusting the comments  too.

qmfrederik:
Makes sense, updated the comments.  By the way, the reason for this PR is that clang now has better support for Objective C on MinGW:

- https://github.com/llvm/llvm-project/pull/77797
- https://github.com/llvm/llvm-project/pull/77385 
- https://github.com/llvm/llvm-project/pull/77255

qmfrederik:
@rfm Anything else you need from me for this PR?

PR #369, opened 2024-02-01
Initialize Android process with current locale and system time zone

triplef:
Fixes NSLocale and NSTimeZone on Android.

Also adds another public initialization method to allow users to take control over initialization arguments.


rfm (approved):
Looks good, thanks.

PR #373, opened 2024-02-07
./configure thread-safe +initialize check: include unistd everywhere except msvc

jcaselman-keysight:
The config check for thread-safe +initialize was failing to build on msys2/mingw64, and it seemed reasonable to force it to use pthread.h since that's what it did [a long time ago](https://github.com/gnustep/libs-base/blob/gnustep_testplant_branch/config/config.initialize.m#L5), before msvc support was introduced

### DIFF
--- a/config/config.initialize.m
+++ b/config/config.initialize.m
@@ -4,6 +4,10 @@
 #include "objc-common.g"
 #include <stdio.h>
 
+#if !defined(_MSC_VER)
+#include <unistd.h>
+#endif
+
 #if defined(_WIN32)
 
 #include <process.h>


trunkmaster:
The problem may relate to wider number of platforms than just msys2/mingw64. At least on Fedora 39 this configure test also fails. Compiler failed to find sleep() function. After manual addition of `#include <unistd.h>` in config/config.initialize.m test succeeds.

rfm:
The test normally passes for msys2/mingw64 so the problem must be specific to something else ... as such the patch to prefer pthread must be wrong (the reason to prefer pthread can't be that the system is msys2/mingw64).
Please could you find out what is causing the normal test using windows native threading to fail.

The bit about using unistd.h when using pthreads seems reasonable though.

jcaselman-keysight:
I was originally building the 1.29.0 tag which did not include https://github.com/gnustep/libs-base/commit/fe863c981d61e50e3ea9cbe7f2b23135c1f2faf1 which led me astray a bit

The only issue I see on the tip is `undeclared function usleep` which is fixed by including unistd, so I have attempted now to include unistd everywhere except msvc @rfm

jcaselman-keysight:
@rfm I see the CI failures but they don't appear to me to be related to my changes?

qmfrederik:
@rfm 

> The test normally passes for msys2/mingw64 so the problem must be specific to something else

The test passes on msys/mingw64 when using gcc-objc, but fails on msys/mingw64 when using libobjc2.  I assume the GCC objc implementation includes unistd.h whereas libobjc2's headers do not.

I'll put up a PR which addes CI for msys/mingw64 + libobjc2, but given this information, would you mind re-considering this PR?

Also,

>  the CI failures but they don't appear to me to be related to my changes?

I think the CI failures have been introduced by d6bb6deff602ec23cf8bc35ba7c53a1fd8275f30

qmfrederik:
... and here's the relevant config.log output:

```
configure:8741: checking for thread-safe +initialize in runtime
configure:8750: clang -o conftest.exe -g -O2  -I/d/a/libs-base/libs-base/build/include -I/d/a/libs-base/libs-base/build/include -I/d/a/libs-base/libs-base/build/include  -x objective-c -fuse-ld=lld -lstdc++ -lgcc_s -L/d/a/libs-base/libs-base/build/lib -L/d/a/libs-base/libs-base/build/lib -L/d/a/libs-base/libs-base/build/lib conftest.c  -Wl,--enable-auto-import -fuse-ld=lld -lstdc++ -lgcc_s -fexceptions -fobjc-runtime=gnustep-2.0 -fblocks -L/home/runneradmin/GNUstep/Library/Libraries -L/d/a/libs-base/libs-base/build/lib -lobjc -lws2_32 -ladvapi32 -lcomctl32 -luser32 -lcomdlg32 -lmpr -lnetapi32 -lm -I.     >&5
In file included from conftest.c:108:
In file included from ././config/config.initialize.m:4:
././config/objc-common.g:54:3: warning: assignment to Objective-C's isa is deprecated in favor of object_setClass() [-Wdeprecated-objc-isa-usage]
   54 |   obj->isa = self;
      |   ^  ~~~~~~~     
      |   object_setClass( , )
././config/objc-common.g:47:5: note: instance variable is declared here
   47 |  id isa;
      |     ^
In file included from conftest.c:108:
././config/config.initialize.m:92:4: error: call to undeclared function 'usleep'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   92 |           mySleep(1);
      |           ^
././config/config.initialize.m:13:20: note: expanded from macro 'mySleep'
   13 | #define mySleep(X)      usleep(1000*(X))
      |                         ^
././config/config.initialize.m:92:4: note: did you mean '_sleep'?
././config/config.initialize.m:13:20: note: expanded from macro 'mySleep'
   13 | #define mySleep(X)      usleep(1000*(X))
      |                         ^
D:/a/_temp/msys64/mingw64/include/stdlib.h:689:24: note: '_sleep' declared here
  689 |   _CRTIMP void __cdecl _sleep(unsigned long _Duration) __MINGW_ATTRIB_DEPRECATED;
      |                        ^
In file included from conftest.c:108:
././config/config.initialize.m:105:4: error: call to undeclared function 'usleep'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  105 |           mySleep(1);
      |           ^
././config/config.initialize.m:13:20: note: expanded from macro 'mySleep'
   13 | #define mySleep(X)      usleep(1000*(X))
      |                         ^
1 warning and 2 errors generated.
```

rfm:
I thought this change had been merged.

jcaselman-keysight:
@rfm this has not been merged

There are two issues building `config.initialize.m` in `1.29.0` under msys2/mingw64:
* one that's already fixed on `master` by https://github.com/gnustep/libs-base/commit/fe863c981d61e50e3ea9cbe7f2b23135c1f2faf1
and
* one that's still happening on tip of `master` and fixed by this PR

rfm:
I couldn't re-open this, so I added the change manually.

PR #380, opened 2024-03-10
Implement +[NSDate now]

hmelder:
See https://developer.apple.com/documentation/foundation/nsdate/3180113-now?language=objc


rfm (approved):
Looks fine to me.  Thanks.

PR #383, opened 2024-03-27
Tests/base/NSURL/basic.m: Use https://gnustep.github.io instead of http://example.com

qmfrederik:
http://example.com/silly-file-name is currently returning 500 instead of 404, breaking tests.  Use https://gnustep.github.io instead.

### DIFF
--- a/Tests/base/NSURL/basic.m
+++ b/Tests/base/NSURL/basic.m
@@ -56,22 +56,22 @@ int main()
   url = [NSURL URLWithString:@"this isn't a URL"];
   PASS(url == nil, "URL with 'this isn't a URL' returns nil");
 
-  url = [NSURL URLWithString: @"http://example.com/silly-file-name"];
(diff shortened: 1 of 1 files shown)


ivucica:
In an ideal world, these tests

1. should not be making an off-local-machine request in the first place (i.e. tests should spin up a local HTTP server, configure resolvers correctly if trying to validate resolver behavior as well, and point at the appropriate isolated hostname+path)
2. alternatively, should still not be making a request to GitHub hosted infrastructure, but ours

However, it's better than example.org, and I'd prefer not to make tests actually depend on our current infra.

I'll go ahead and merge this myself since tests failing for such a silly reason is a reason to fix them in itself.

ivucica:
n.b. MSVC Clang x64 still fails with the same error; strange:

```
base/NSURL/basic.m:
Failed test:     (2024-03-27 13:25:47.843 +0000) basic.m:63 ... Status of load is 404 for gnustep.github.io/silly-file-name
```

PR #384, opened 2024-03-30
NSNetService and NSNetServiceBrowser: Throw runtime exceptions

qmfrederik:
Always compile `NSNetService` and `NSNetServiceBrowser`, but have their `intialize` method throw an exception if libavahi and mDNS are unavailable.

This:
- Results in a clear error message if code which was compiled agains a copy of GNUstep with avahi/mDNS support runs on a copy of GNUstep which doesn't have avahi/mDNS support.
- Avoids code which uses `NSNetService` or `NSNetServiceBrowser` to successfully compile (because the classes are defined in the header) but then failing to link (because there is no implementation for these classes).

### DIFF
--- a/Source/GNUmakefile
+++ b/Source/GNUmakefile
@@ -270,6 +270,7 @@ NSMeasurement.m \
 NSMetadata.m \
 NSMetadataAttributes.m \
 NSMethodSignature.m \
+NSNetServices.m \
 NSNotification.m \
 NSNotificationCenter.m \
 NSNotificationQueue.m \
@@ -391,8 +392,7 @@ endif
 
 ifeq ($(GNUSTEP_BASE_HAVE_MDNS), 1)
   BASE_MFILES += \
-	GSMDNSNetServices.m \
-	NSNetServices.m
+	GSMDNSNetServices.m
 endif
 
 ifeq ($(GNUSTEP_BASE_HAVE_AVAHI), 1)
@@ -400,8 +400,7 @@ ifeq ($(GNUSTEP_BASE_HAVE_AVAHI), 1)
 	GSAvahiNetService.m \
 	GSAvahiNetServiceBrowser.m \
 	GSAvahiClient.m \
-	GSAvahiRunLoopIntegration.m \
-	NSNetServices.m
+	GSAvahiRunLoopIntegration.m
 endif
 
 ifeq ($(WITH_FFI),libffi)
--- a/Source/NSNetServices.m
+++ b/Source/NSNetServices.m
@@ -31,6 +31,7 @@
 #import "Foundation/NSArray.h"
 #import "Foundation/NSData.h"
 #import "Foundation/NSDictionary.h"
+#import "Foundation/NSException.h"
 #import "Foundation/NSHost.h"
 #import "Foundation/NSStream.h"
 #import "Foundation/NSString.h"
@@ -48,7 +49,7 @@ static Class concreteBrowserClass;
       abstractServiceClass = self;
 #     if GS_USE_AVAHI==1 
         concreteServiceClass = [GSAvahiNetService class];
-#     else
+#     elif GS_USE_MDNS==1
         concreteServiceClass = [GSMDNSNetService class];
 #     endif
   }
@@ -58,7 +59,14 @@ static Class concreteBrowserClass;
 {
   if (self == abstractServiceClass)
     {
-      return [concreteServiceClass allocWithZone: zone];
+      if (concreteServiceClass != nil)
+      {
+        return [concreteServiceClass allocWithZone: zone];
+      }
+      else
+      {
+        return nil;
+      }
     }
   return [super allocWithZone: zone];
 }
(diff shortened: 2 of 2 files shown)


qmfrederik:
The test failures are unrelated, should be fixed via #383

gcasa:
@rfm, @qmfrederik I agree with the building of NSNetServices no matter what, that makes perfect sense, the part I am uncertain about is throwing an exception.  This may cause code that might otherwise function to fail unless modified to catch the exception.   @rfm, please use your discretion to decide what you believe should happen here.

gcasa:
@rfm, @qmfrederik another option here may be to introduce a default/setting that will allow setting this behavior.   It could be off (e.g. simply log) by default, but throw an exception if needed and could be set in an application's defaults.

rfm (changes requested):
I think Greg likes the idea of everything looking fine (it being able to build/link Apple code with gnustep as easily as possible), even if the code than silently fails to work when used ... the blame is attached to the developer who failed to check the return value from the method to get an instance of a class (which returned nil).

I do not agree with that approach, preferring the opposite: I think code that can't work should fail at compile time so that the developer knows about it and can fix it.  If I was going to improve on the existing code I'd perhaps want to make it clearer to the developer, at the point when they try to compile, exactly *why* the feature is not available and how to fix it.

This patch seems to move to somewhere between these two approaches, with the advantages of neither, so I'm not sure what the motivation is.  Can you explain?

qmfrederik:
So there's two dimensions to this:

### Handing an initialization failure

[The Apple documentation](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Initialization/Initialization.html#//apple_ref/doc/uid/TP40010810-CH6-SW57) states, in the "Issues with Initializers" section:

> Sometimes an `init...` method cannot perform the initialization requested. For example, an `initFromFile:` method expects to initialize an object from the contents of a file, the path to which is passed as a parameter. But if no file exists at that location, the object cannot be initialized. A similar problem happens if an `initWithArray:` initializer is passed an `NSDictionary` object. When an `init...` method cannot initialize an object, it should:
> - Release the newly allocated object (in memory-managed code)
> - Return `nil`
>
> Returning nil from an initializer indicates that the requested object cannot be created.

The [`init documentation`](https://developer.apple.com/documentation/objectivec/nsobject/1418641-init) documentation makes a similar statement:

> If the new object can’t be initialized, the method should return `nil`.

So in that view, I think @gcasa is right, and "blame is attached to the developer who failed to check the return value from the method to get an instance of a class (which returned nil)."

In that spirit, I've updated the PR to have `[NSNetService initWithDomain]` and `[NSNetServiceBrowser init]` return `nil`.

### Compile time versus runtime

libavahi/mDNS are optional dependencies of GNUstep, so it's possible to compile GNUstep without a working implementation of `NSNetService` and `NSNetServiceBrowser`.

In the current situation, you get a linker error when you compile code that uses `NSNetService(Browser)` because the classes are defined in the headers but don't exist in gnustep-base.  You would get a runtime exception when you run code which has been built using a copy of GNUstep which has avahi/mDNS support using a copy of GNUstep which doens't have that optional dependency.

It's the second scenario I'm most concerned with -- `NSNetService(Browser)` is part of the GNUstep API.  So I believe GNUstep should always include an implementation of these interfaces.  The conventional way to signal that a class is non-functional is by having its `init` method return `nil`, which the proposed change does.

qmfrederik:
The other option (perhaps in combination with the patch as currently proposed) is to not install the headers for `NSNetService` and `NSNetServiceBrowser` if the classes are not implemented.

However, if we go down that path, what would Objective-C code which accounts for both scenarios (with and without bonjour support) look like?  If you have the runtime check, you could do `if ([[NSNetService] alloc] init] == nil)`, but what would you do if the headers are missing?

rfm (approved):
OK.  I would ideally like a compile time warning, but I don't see a good way to do that.  So I think this is good to commit.

gcasa:
> I think Greg likes the idea of everything looking fine (it being able to build/link Apple code with gnustep as easily as possible), even if the code than silently fails to work when used ... the blame is attached to the developer who failed to check the return value from the method to get an instance of a class (which returned nil).
> 
> 
> 
> I do not agree with that approach, preferring the opposite: I think code that can't work should fail at compile time so that the developer knows about it and can fix it.  If I was going to improve on the existing code I'd perhaps want to make it clearer to the developer, at the point when they try to compile, exactly *why* the feature is not available and how to fix it.
> 
> 
> 
> This patch seems to move to somewhere between these two approaches, with the advantages of neither, so I'm not sure what the motivation is.  Can you explain?

I try to take things on a more case by case basis.

My feeling is that many users are not necessarily developers.  Throwing a runtime exception might cause a crash in an application that is not using NSNetService for any core functionality.  For instance, Eggplant uses it, but can launch and function without it and the application remains fully usable.

So here it was my feeling that, in some cases, there may be applications that will use NSNetService in a similar way so introducing an exception would be a disruptive change in behavior.

That being said, I agree with @rfm 's decision.

qmfrederik:
Thanks both.  I'm not a contributor to this repo, so I'd rely on someone else to go ahead and merge this.

gcasa:
> Thanks both.  I'm not a contributor to this repo, so I'd rely on someone else to go ahead and merge this.

Given @rfm 's approval I will proceed to merge.  Thanks @qmfrederik

PR #386, opened 2024-04-17
Refactor +[NSTimeZone timeZoneArray] to use NSString instead of raw FDs

hmelder:
We cannot use raw fopen, fscanf, etc. to access Assets on Android.
A higher-level API like +[NSString stringWithContentsOfFile:encoding:error:] should be used instead.


rfm (approved):
Looks great, thanks.

PR #387, opened 2024-04-18
Add support for -[NSThread setName:] on win32 and fix inconsistency in API behaviour

hmelder:
Implements support for -[NSThread setName:] on win32. I am using `GetThreadDescription` and `SetThreadDescription` from `processthreadsapi.h`. This API was introduced in Windows 10 Redstone 1. Here is an overview of setting a thread name in native code: https://learn.microsoft.com/en-us/visualstudio/debugger/how-to-set-a-thread-name-in-native-code?view=vs-2022.

The test currently only checks correct behaviour on Windows, as I do not know how to reconfigure `configure.ac`. `autoreconf -vi` results in an erroneous `configure` script on Debian. @rfm how do you reconfigure autoconf in gnustep-base?

```objc
#import <pthread.h>

int main() {
				@autoreleasepool {
								pthread_setname_np("Hello World");
								NSLog(@"Current Thread: %@", [[NSThread currentThread] name]);
								[[NSThread currentThread] setName:@"Test"];
								NSLog(@"Current Thread: %@", [[NSThread currentThread] name]);
								pthread_setname_np("Hello World");
								NSLog(@"Current Thread: %@", [[NSThread currentThread] name]);
				}
				return 0;
}
```

Output on macOS:
```sh
hugo@w192-m-v4 ~ % ./a.out
hugo@w192-m-v4 ~ % ./a.out
2024-04-18 23:40:31.432 a.out[24036:1426738] Current Thread: Hello World
2024-04-18 23:40:31.432 a.out[24036:1426738] Current Thread: Test
2024-04-18 23:40:31.432 a.out[24036:1426738] Current Thread: Test
```

Foundation's NSThreads implementation seems to retrieve the current thread's name in the initialiser. -[NSThread name] returns nil, until a name is set with -[NSThread setName:].

### DIFF
--- a/Tests/base/NSThread/name.m
+++ b/Tests/base/NSThread/name.m
@@ -1,0 +1,44 @@
+#import "ObjectTesting.h"
+#include <winerror.h>
+#import <Foundation/NSThread.h>
+
+#if defined(_WIN32) && (NTDDI_VERSION >= NTDDI_WIN10_RS1)
+#include <processthreadsapi.h>
+#else
+#include <pthread.h>
+#endif
+
+
+int main(void)
+{
+  NSAutoreleasePool *arp = [NSAutoreleasePool new];
+// SetThreadDescription() was added in Windows 10 1607 (Redstone 1)
+#if defined(_WIN32) && (NTDDI_VERSION >= NTDDI_WIN10_RS1)
+    PWSTR nativeThreadName = NULL;
+    HANDLE current = GetCurrentThread();
+    HRESULT hr = SetThreadDescription(current, L"Test");
+    PASS(SUCCEEDED(hr), "SetThreadDescription was successful");
+
+    NSString *name = [[[NSThread alloc] init] name];
+    PASS(name != nil, "-[NSThread name] returns a valid string");
+    NSLog(@"Name: %@", name);
+    PASS([name isEqualToString: @"Test"], "Thread name is correct");
+    [name release];
+
+    [[NSThread currentThread] setName: @"Test2"];
+    name = [[NSThread currentThread] name];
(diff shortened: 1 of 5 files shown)


rfm:
I had exactly the same problem running autoconf on debian earlier this week. From the warnings/errors I think our configure.ac uses obsolete features that need to be changed to work with latest autoconf (presumably autoconf release notes / documentation would help with that). The quick workaround I used was simply to regenerate configure from.configure.ac using an older version of autoconf (I used my old CentOSt 8 VM). I can't do that for you now though, as I'm away from my computer for a week.

hmelder:
> I had exactly the same problem running autoconf on debian earlier this week

But it is just `autoreconf -vi`, right? Autoreconf generated a new `configure` but when running it, it just said `cannot find required auxiliary files:` (without a list of files).

Tests/base/NSThread/name.m
```diff
@@ -0,0 +1,44 @@
...
+    PWSTR nativeThreadName = NULL;
+    HANDLE current = GetCurrentThread();
+    HRESULT hr = SetThreadDescription(current, L"Test");
+    PASS(SUCCEEDED(hr), "SetThreadDescription was successful");
+
+    NSString *name = [[[NSThread alloc] init] name];
+    PASS(name != nil, "-[NSThread name] returns a valid string");
+    NSLog(@"Name: %@", name);
+    PASS([name isEqualToString: @"Test"], "Thread name is correct");
+    [name release];
```
hmelder, line 26: I have to release NSThread and not name here

hmelder:
@rfm ready for review!

rfm (approved):
Builds OK on a number of systems.

PR #389, opened 2024-04-20
Use AC_SEARCH_LIBS instead of AC_CHECK_LIB for sframe

qmfrederik:
Commit 4515e4a28b63ccd3e715cb21736e95242d671843 introduced a `AC_CHECK_LIB(sframe, sframe_decode)` in `configure.ac`. 

This causes libs-base to take a hard dependency on `libsframe`; `AC_CHECK_LIB` will prepend `lsframe` to LIBS even if the library is not found.  This results in a build failure on systems that don't have libsframe installed (and on which libsframe isn't available).  For example, libsframe is not available on Enterprise Linux (e.g. Rocky Linux).

This PR slightly modifies the original commit by using `AC_SEARCH_LIBS` making the library optional.

### DIFF
--- a/Headers/GNUstepBase/config.h.in
+++ b/Headers/GNUstepBase/config.h.in
@@ -412,6 +412,9 @@
 /* Define to 1 if you have the `z' library (-lz). */
 #undef HAVE_LIBZ
 
+/* Define to 1 if you have the `zstd' library (-lzstd). */
+#undef HAVE_LIBZSTD
+
 /* Define to 1 if you have the <limits.h> header file. */
 #undef HAVE_LIMITS_H
 
--- a/configure
+++ b/configure
@@ -9505,13 +9505,12 @@ _ACEOF
 
 fi
 
-{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for sframe_decode in -lsframe" >&5
-$as_echo_n "checking for sframe_decode in -lsframe... " >&6; }
-if ${ac_cv_lib_sframe_sframe_decode+:} false; then :
+{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for library containing sframe" >&5
+$as_echo_n "checking for library containing sframe... " >&6; }
+if ${ac_cv_search_sframe+:} false; then :
   $as_echo_n "(cached) " >&6
 else
-  ac_check_lib_save_LIBS=$LIBS
-LIBS="-lsframe  $LIBS"
+  ac_func_search_save_LIBS=$LIBS
 cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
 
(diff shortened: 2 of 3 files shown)


qmfrederik:
@rfm The commit message stated "libsframe needed on some systems" but it didn't mention which systems, so I couldn't verify this PR on those systems.

qmfrederik:
FYI, the build error on Rocky Linux 9 is:

```
./configure
[...]
checking for sframe_decode in -lsframe... yes
[...]
checking ffi.h usability... yes
checking ffi.h presence... yes
checking for ffi.h... yes
checking for forwarding callback in runtime... yes
checking FFI library usage... configure: error: The ffi library (libffi) does not appear to be working.  Perhaps it's missing or you need a more recent version.  Version 3.0.9 or later should work, and you can find a link to it n the list of packages for download at http://www.gnustep.org/resources/sources.html
```

and in `config.log` we have:

```
configure:9508: checking for sframe_decode in -lsframe
configure:9533: clang -o conftest -g -O2  -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include  -x objective-c -fuse-ld=lld -L/usr/local/lib64 -lstdc++ -L/usr/local/lib64 -L/usr/local/lib64 -L/usr/local/lib64 -L/usr/local/lib64 conftest.c -lsframe  -lz -lrt -ldl  -lpthread >&5
configure:9533: $? = 0
configure:9542: result: yes
[...]
configure:11499: checking FFI library usage
configure:11520: clang -o conftest -g -O2  -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include  -x objective-c  -fuse-ld=lld -L/usr/local/lib64 -lstdc++ -L/usr/local/lib64 -L/usr/local/lib64 -L/usr/local/lib64 -L/usr/local/lib64 conftest.c -L/usr/lib64/../lib64 -lffi  -lsframe -lz -lrt -ldl  -lpthread -lz >&5
configure:11520: $? = 0
configure:11520: ./conftest
./conftest: error while loading shared libraries: libsframe.so.0: cannot open shared object file: No such file or directory
configure:11520: $? = 127
configure: program exited with status 127
```

qmfrederik:
@rfm OK - digging a bit deeper, looks like what happened is that a recent binutils version introduced `libsframe` and made it a
dependency of `libbfd`. This probably caused the build failure you're seeing.

I tested these changes on Ubuntu 24.04 (which has the libsframe as a separate library) and that works.

Would you mind having a look at this PR?  The build is broken as is on Rocky Linux.

rfm:
I'm away this week, but will look at it when I get home.  I was under the impression that, if libsframe was not found autoconf would simply produce a configuration that didn't try to log no with it (so nothing could be broken by testing for it), but I guess I misunderstood something.  I'm sure it can be fixed quite easily.

PR #390, opened 2024-04-20
Add a CI job for Rocky Linux 9

qmfrederik:
4515e4a28b63ccd3e715cb21736e95242d671843 inadvertently broke building libs-base on Enterprise Linux (e.g. Rocky Linux).  #389 fixes the build, and this PR adds a CI job for Rocky Linux 9, hopefully helping us catch these errors when a PR for such changes gets raised.

It:
- Runs the Ubuntu 22.04 and Rocky Linux 9 jobs in a container
- Accounts for the lib/lib64 difference between Ubuntu & Rocky Linux

#389 and #388 are required to make this build succeed.

### DIFF
--- a/.github/scripts/dependencies.sh
+++ b/.github/scripts/dependencies.sh
@@ -14,7 +14,7 @@ install_gnustep_make() {
     if [ -n "$RUNTIME_VERSION" ]; then
       MAKE_OPTS="$MAKE_OPTS --with-runtime-abi=$RUNTIME_VERSION"
     fi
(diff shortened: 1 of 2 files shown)


qmfrederik:
@rfm This adds a Rocky CI leg, which could have helped catch the regression in 6df6b1cb97aa40a3886114ed59dae6c65b1a37f5.  Would you agree with this?  I can add another leg for the OS you were targeting to make sure `./configure` works fine on that platform, if that helps.

rfm:
I have no problem with adding more (reasonably popular/current) target operating systems.

PR #391, opened 2024-04-20
Remove Ubuntu x64 Clang gnustep-1.9 build leg

qmfrederik:
This build has been broken for a while, and I'm going to assume most users are on the 2.0 ABI by now.


rfm (approved):
That seems reasonable to me.

PR #392, opened 2024-04-20
Use actions/upload-artifact@v4

qmfrederik:
actions/upload-artifact@v3 is scheduled for deprecation on November 30, 2024, so upgrade to v4.

Also hardcode the paths to `source/config.log` and `source/Tests/tests.log`, for some reason test logs weren't always captured and this fixes that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -162,14 +162,13 @@ jobs:
           make check
 
       - name: Upload logs
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: always()
         with:
           name: Logs - ${{ matrix.name }}
           path: |
-            ${{ env.SRC_PATH }}/config.log
-            ${{ env.SRC_PATH }}/Tests/tests.log
-
+            source/config.log
+            source/Tests/tests.log
 
   ########### Windows ###########
   windows:
@@ -386,10 +385,10 @@ jobs:
           make check
 
       - name: Upload logs
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: always()
         with:
           name: Logs - ${{ matrix.name }}
           path: |
-            ${{ env.SRC_PATH }}/config.log
-            ${{ env.SRC_PATH }}/Tests/tests.log
+            source/config.log
+            source/Tests/tests.log


ivucica (changes requested):
Is the envvar change related to the upload-artifact version upgrade?

qmfrederik:
So, w.r.t. the env var: there's some _really_ weird behavior in GitHub when using upload-artifacts with jobs in containers.

I'd like for the Linux CI job to run in a container, so we can have CI for Rocky Linux, too (see #390, a recent change to master broke building on Rocky and this would have caught that).

The upload-artifacts action has always had some issues when used with a job which runs in a container (e.g. https://github.com/actions/upload-artifact/issues/13) as it runs on the host.

In our scenario, it looks like it gets confused when translating container paths to host paths if the artifacts are specified using absolute paths.  Regardless whether I'm using:

```
      - name: Upload logs
        uses: actions/upload-artifact@v4
        if: always()
        with:
          name: Logs - ${{ matrix.name }}
          path: |
            ${{ github.workspace }}/source/config.log
            ${{ github.workspace }}/source/Tests/tests.log
```

or

```
      - name: Upload logs
        uses: actions/upload-artifact@v3
        if: always()
        with:
          name: Logs - ${{ matrix.name }}
          path: |
            ${{ env.SRC_PATH }}/config.log
            ${{ env.SRC_PATH }}/Tests/tests.log
```

the path for the _second_ entry in the list gets converted to an invalid path, and the tests.log artifact is not uploaded.  Here's a copy of the debug output of the upload-artifact step in the job (the full log is here: https://github.com/qmfrederik/libs-base/actions/runs/8780812326/job/24091501960#step:9:82)

```
##[debug]Search path '/__w/libs-base/libs-base/source/config.log'
##[debug]Search path '/home/runner/work/libs-base/libs-base/source/Tests/tests.log'
```

When using a relative path, this issue does not appear.

So there's a couple of things at play:

1. Upgrading to the latest version of the upload-artifact action
2. A desire to get the upload-artifact action to work in a job which runs in a container
3. What appears to be a weird bug in GitHub, which happens when using full paths for artifacts; here, the workaround is to use relative paths for artifacts

qmfrederik:
hi @ivucica , anything else you'd like me to comment on / change in this PR?

ivucica:
Apologies @qmfrederik, I am not regularly following any activity on GitHub on any project, so this got caught in that as well.

Having played with GH actions recently, I'd have approved this, as the CWD is already supposed to be the current 'workspace'.

ivucica:
I can't find a way to mark this as LGTM after-the-fact, after the merge has already happened.

PR #396, opened 2024-04-26
NotificationCenter removeObserver

williameveretteggplant:
This modifies the NotificationCenter class to do the following:
  - add an ivar to the class for an array to hold observers that need to be released when they are removed
  - adds an observer to that array if the observer is of class GSNotificationObserver
  - upon removing an observer, check if it is in that array, if so remove it from the array and release it.
 
This replaces the previous implementation which checked the class of the observer as it was being removed, which would cause a crash if the observer had been deallocated.

### DIFF
--- a/Headers/Foundation/NSNotification.h
+++ b/Headers/Foundation/NSNotification.h
@@ -42,6 +42,7 @@ extern "C" {
 @class NSDictionary;
 @class NSLock;
 @class NSOperationQueue;
+@class NSMutableArray;
 
 typedef NSString* NSNotificationName;
 
--- a/Headers/Foundation/NSNotification.h
+++ b/Headers/Foundation/NSNotification.h
@@ -74,7 +74,8 @@ GS_EXPORT_CLASS
 {
 #if	GS_EXPOSE(NSNotificationCenter)
 @private
-  void	*_table;
+  void	         *_table;
+  NSMutableArray *retainedObserverList;
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


qmfrederik:
The MinGW GCC build errors are weird.  I assume it's because you are using an iterator in your code.  What happens if you change:

```objc
  for (id listObserver in _retainedObserverArray) 
    {
      if (listObserver == observer)
        {
	  needToRelease = YES;
          break;
	}
    }
```

to

```objc
for (int i = 0; i < [cs count]; i++)
  {
     if ([_retainedObserverArray objectAtIndex: i] == observer)
       {
	  needToRelease = YES;
          break;
       }
  }
```

or just

```objc
if ([_retainedObserverArray containsObject:observer)
  {
    needToRelease = YES;
    break;
  }
```

?

rfm:
The best code for NSArray would use the -indexOfObjectIdenticalTo: since (as well as being a portable alternative) this is simpler and more efficient than the use if fast enumeration.

I'm not completely sure that holding the observer in a mutable array is the best approach though; an array is efficient if there are a small number of observers, but doesn't scale well if there are a lot of observers being added and removed.  To scale, the function API of NSHashTable API (or using the GSI macros) would probably be fastest.

rfm (commented):
It occurs to me that it would be better to add the new collection to the table rather than as a direct ivar of NSNotification (to hide it).

williameveretteggplant:
I have taken suggestions into account and updated.

rfm (approved):
That all looks workable.

PR #398, opened 2024-04-28
win64/clang+msvc: Skip failing tests and mark flaky test as hopeful

qmfrederik:
A number of NSURL-related tests fail or crash on 64-bit windows and compiled with the msvc front-end for clang, causing CI to be red.

This PR marks the failing tests as hopeful and skips the crashing tests altogether.

#367 tracks the root cause, but this hopefully gets CI back go green for now.


qmfrederik:
@rfm Are you good with this change?

rfm:
Yes ... it's a poor substitute for someone with access to a win64/clang+msvc system diagnosing/fixing, but I agree that it's better than just leaving it.

PR #400, opened 2024-05-06
GSFileURLHandle: Remove file handle caching

hmelder:
Currently, GSFileURLHandle caches file handles indefinitely. I opted to use NSCache with the cost being the size of the loaded data. By default, the size of the cache is 4MiB. The cache registration functionality was moved into `-[GSFileURLHandle loadInForeground]` as we need to know the size of the retrieved data.


hmelder:
Thinking about it, it might be better to just skip caching of file handles.

triplef:
This was previously discussed here: https://github.com/gnustep/libs-base/pull/309#pullrequestreview-1556770837

rfm:
> On 6 May 2024, at 17:16, Hugo Melder ***@***.***> wrote:
> 
> 
> Thinking about it, it might be better to just skip caching of file handles.

Yes, I think that's the best thing for file URLs

rfm (approved):

PR #402, opened 2024-05-07
NSURLSession: Invoke Completion Handler upon failed request

hmelder:


rfm (approved):
That looks to be correct behavior to me.

PR #404, opened 2024-05-14
Fix NSString ICU includes for Windows

triplef:
Fixes building with the Windows-bundled ICU libraries (was broken with #370).

This wasn’t caught by CI because it’s still using Windows 2019 agents which don’t have the ICU library. I’ll open a separate PR to update this.


rfm (approved):
That looks fine, thanks.

PR #405, opened 2024-05-14
Two small xml fixes

sensetalkdoug:
Fix execute_xpath() to return an NSError when expression can't be evaluated; Remove previous attributes when setting attributes; Add related tests

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2024-04-29  Doug Simons <douglas.simons@keysight.com>
+
+    * NSXMLNode.m:
+       Fix execute_xpath() to return an NSError when xpath expression can't
+	   be evaluated instead of just calling NSLog.
+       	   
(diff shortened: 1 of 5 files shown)


sensetalkdoug:
While working with these classes, I also noticed this problem, which I did not try to fix:
The [NSXMLElement setNamespaces:] method is broken. It calls xmlFreeNsList() which frees the memory associated with all of the old namespaces. But there can be many pointers to those namespaces still in the ns entries of various nodes in the tree, which can lead to memory access crashes. Finding all of those references and setting them to NULL would prevent the crashing, but I believe is the wrong answer since the replacement namespaces being added by the call to setNamespaces: may include new (or even identical) entries for the same prefixes, so ideally you’d like to reset each node to point to the new namespace for that prefix.

rfm (approved):
Both changes looks good, and nice to see testcases.  Thanks.

PR #407, opened 2024-05-18
Implement -[NSString enumerateLinesUsingBlock:]

hmelder:


hmelder:
@rfm

rfm (approved):
While I detest the API as over-engineered, it's what Apple chose to do, and looks like a correct implementation.
Perhaps we should add something cleaner like
- (NSString*) nextLineFrom: (NSUInteger*)pos;
as a simpler convenience method for people using gcc (or wanting clearer code)?

PR #411, opened 2024-05-25
NSURLSession Reimplementation

hmelder:
This Pull Request includes a complete reimplementation of the NSURLSession API.

Issues with the current implementation:
- Incorrect implementation of the libcurl Multi Socket API (Timer Resetting etc.)
- To much indirections (At least 9 custom Classes just for transfer setup)
- Linear search to associate handles with wrapper instances
- Poor error descriptions
- Hard to maintain and debug
- Missing features such as NSURLSessionUploadTask, Metrics, etc.
- Session and tasks were never released due to Retain Cycle in Block Dispose Function

I reduced the implementation to just two main classes: NSURLSession and NSURLSessionTask.

NSURLSession manages a libcurl multi handle and the required plumming via libdispatch. Timeout and socket actions are handled in the libdispatch work queue. Delegate messages and completion handlers are submitted to a seperate serial NSOperationQueue.

### Session Setup

When creating a new session, a libcurl multi handle is created and configured according to the NSURLSessionConfiguration object that was passed as an argument.

A serial dispatch queue is created for handling libcurl timers (dispatch_timer) and socket events (via dispatch_io).

The libcurl multi handle delegates all plumming the the application. This means we can drive the full transfer via libdispatch. libcurl instructs the application to set up timers. A call to a libcurl update function is made when the timer fires.

### Task Setup

The task object is first initialised by the session (i.e -[NSURLSession dataTaskWithRequest:]).
When creating the NSURLSessionTask, a new libcurl easy handle is created and configured with the settings from NSURLRequest and NSURLSessionConfiguration.
Before returning the newly created task object, the session configures the task by ORing Bit fields together and adds the task object to an array of ongoing transfers.

The bit field is required, as we have different paths along the request. There is the traditional delegate based path (See NSURLSessionTaskDelegate and friends) which has a large number of mostly optional methods for managing the transfer in detail.

There is also the possiblity of using a completion handler instead. Data is aggregated internally (or just written to a file in case of a downloadTask) and passed as an argument to the completion handler.

The internal options
```objc
typedef NS_ENUM(NSInteger, GSURLSessionProperties) {
  GSURLSessionStoresDataInMemory = (1 << 0),
  GSURLSessionWritesDataToFile = (1 << 1),
  GSURLSessionUpdatesDelegate = (1 << 2),
  GSURLSessionHasCompletionHandler = (1 << 3),
  GSURLSessionHasInputStream = (1 << 4)
};
```

The pointer to the task instance is now returned to the caller.

### Task Resumption

By default, the task is in suspended state. When -[NSURLSessionTask resume] is called for the first time, the session is retained
and -[NSURLSession _resumeTask:] is called. -[NSURLSession _resumeTask:] adds the underlying easy handle to the session’s multi handle (curl_multi_add_handle).
This in turn triggers a timer setup or reconfiguration (timer_callback).

The timer event handler calls curl_multi_socket_action and checks for task completion.

-[NSURLSession _checkForCompletion] reads the backlog of messages from the multi handle (curl_multi_info_read). If a easy handle is marked as done, the handle is removed from the multi handle (curl_multi_remove_handle), the task removed from the list of active tasks and -[NSURLSessionTask _transferFinishedWithCode:] is called.

For more information on how the libcurl multi socket API works, visit https://curl.se/libcurl/c/libcurl-multi.html

Related fixes and additions:
- NSBlockOperation: Fixed memory leak
- NSHTTPCookie: Fix ‘expires’ date parsing
- NSURLRequest: Implemented assumesHTTP3Capable
- NSURLRequest: Private method to access case-insensitive Header Dictionary

TODO (Priority High to Low):
- Unit Testing on Android
- Unit Tests for TLS (requires HTTPS test server)
- Custom Protocol Loading
- Caching
- Download with Resume Data
- Web Socket Extensions
- Background Downloading via Daemon (on macOS nsurlsessiond
﻿

Open Questions:

What is the desired behaviour when libcurl configuration or update functions return configuration errors? The old implementations just throws exceptions everywhere.


triplef:
This looks great. I'm traveling atm but will try to test this the week after next.

gcasa:
@hmelder I think with a change of this volume if you haven't submitted your copyright assignment, then you should get it in as soon as possible.

gcasa:
@hmelder as I had said before just email them and cc me on it.  As soon as you do that we should be okay accepting any changes as the FSF process can take a little time.  It's electronic now so it's faster than before but since they are chiefly a volunteer organization my policy has always been to proceed as long as the process has been started.

hmelder:
> Should we also get the author of the original implementation to take a look at this as well?

@theiostream would you be willing to take a look at my implementation? Not sure if you are still familiar with the code base ^^

Source/NSOperation.m
rfm, line 1: Looking at this change I see two issues:
1. Is the idea of removing an execution block as you go the correct behavior?  From the documentation it's not clear to me that there is anything wrong with leaving the block in the _executionBlocks array, to be deallocated when the NSBlockOperation instance is deallocated (it looks like that should happen as soon as all the blocks have been executed).
2. Assuming that we do need to remove blocks as they are executed, the current code is broken (you must not alter a mutable array while enumerating it) and will  go wrong if there is more than one block in the array.  I think the correct mechanism would be to take a shallow copy of the array and enumerate the copy (so the original array can be safely modified).

Source/NSHTTPCookie.m
rfm, line 1: This is a great change (I hadn't realised that the timestamp format in a cookie *must* be in the UTC time zone).
However, the change leaks an NSDateFormatter instance (and has trivial whitespace formatting errors).

rfm (commented):
I think it's great to simplify this code.
I'm pretty unfamiliar with it, so I guess we need a lot of testcases with such a large rewrite, but it generally looks good.
There are a few places where fixes could be merged separately from this patch because they are independent of the NSURLSession code itself, and I had a look at a couple: The NSCookie change looks fine to merge with minor tweaks, but I'm unsure of the NSOperation change (I don't have any practical experience using NSOperation with blocks).

hmelder:
> but I'm unsure of the NSOperation change (I don't have any practical experience using NSOperation with blocks).

Blocks are copied onto the heap (or ref count is increased) and transparently treated as normal Objective-C objects in the Runtime. We add the block to the array of blocks, but AFAIK never remove them after the operation (i.e. calling the block) is completed.

Did a quick check with an NSLog in a -dealloc as well :)

Source/NSHTTPCookie.m
hmelder: Ups you are right. Thanks for catching this.

Source/NSOperation.m
hmelder: >  From the documentation it's not clear to me that there is anything wrong with leaving the block in the _executionBlocks array, to be deallocated when the NSBlockOperation instance is deallocated

I will test this against Apple's implementation, but given that this would result in tasks never being dealloc'd when using the sharedSession.

rfm: Shouldn't the lifetime of an NSBlockOperation be as follows?
1. an instance is created
2. blocks are added to an instance (recorded in _executionBlocks)
3. the operation is added to a queue
4. the operation is executed ... causing all the blocks to be executed
5. the operation is removed from the queue and deallocated (causing _executionBlocks to be deallocated, which in turn causes the blocks to be deallocated)

Unless there is some sort of retain loop where a block which is added to the operation also retains the operation.
If that's the case, an explicit call to -removeAllObjects after enumerating the blocks in the array would be a safe way to break the retain loop.

hmelder: You are right. I have to investigate this further.

triplef:
@rupertdaniel could you please test this branch with our Unity project and if you get a chance with our Qt app (Win/Android) as well?

hmelder:
I do not understand why the CI is suddenly failing. It works just fine on Ubuntu, Rocky Linux, and Windows on my Machine (Objc sections are somehow discarded in the CI). Is there an easy way to reproduce a CI build? 

I've added more tests and added GSCACertificateFilePath which is a NSUserDefaults key for setting the path to a custom CA blob.

hmelder:
The bug was quite nasty. I have assigned a substring to a variable (`headerLine`) which is then released shortly after. However, the substring was also added to the `arc_tls` pool on return. After some time, an already deallocated object was being released on a thread managed by libdispatch causing a crash.

hmelder:
Turns out the latest `tools-windows-msvc` release build does not ship libcurl. I started a manual CI run which uses the master branch.

Here the test results on Windows 11 ARM64

```powershell
OsName                                                  : Microsoft Windows 11 Pro
OsType                                                  : WINNT
OsOperatingSystemSKU                                    : 48
OsVersion                                               : 10.0.22631
WindowsBuildLabEx                                       : 22621.1.arm64fre.ni_release.220506-1250
WindowsCurrentVersion                                   : 6.3
OSDisplayVersion                                        : 23H2
```

```sh
hugo@YellowBox MSYS /c/Users/hugo/Documents/libs-base/Tests/base/NSURLSession
$ gnustep-tests .
--- Running tests in . ---

     84 Passed tests

All OK!
```

hmelder:
@gcasa, you are also invited to review the PR, if you'd like to :)

gcasa:
> @gcasa, you are also invited to review the PR, if you'd like to :)

Naturally, of course, I prefer to leave it up to @rfm as he has more knowledge of this code than I do.

Source/NSURLSession.m
```diff
@@ -28,104 +28,28 @@
...
-#import "Foundation/NSURLError.h"
-#import "Foundation/NSURLSession.h"
-#import "Foundation/NSURLRequest.h"
-#import "Foundation/NSValue.h"
-
-GS_DECLARE const float NSURLSessionTaskPriorityDefault = 0.5;
-GS_DECLARE const float NSURLSessionTaskPriorityLow = 0.0;
-GS_DECLARE const float NSURLSessionTaskPriorityHigh = 1.0;
-
-GS_DECLARE const int64_t NSURLSessionTransferSizeUnknown = -1;
```
triplef, line 55: These constants should not be removed.

Headers/Foundation/NSURLSession.h
```diff
@@ -247,125 +295,77 @@ GS_EXPORT const int64_t NSURLSessionTransferSizeUnknown;
...
- * will be disabled while a task is suspended.
+/**
+ * Cancels the task and the ongoing transfer.
  */
-- (void) suspend;
-- (void) resume;
+- (void)cancel;
 
-- (float) priority;
-- (void) setPriority: (float)priority;
```
triplef, line 363: The `priority` property on `NSURLSessionTask` should not be removed.

hmelder:
I've implemented a custom redirection system for non-convenience tasks. See https://developer.apple.com/documentation/foundation/nsurlsessiontaskdelegate/1411626-urlsession

hmelder:
Feel free to do a complete review of the PR :)

triplef (approved):
While I haven’t reviewed the code changes in detail we’ve tested these changes with our app on different platforms.

triplef:
I’ve rebased and squashed all NSURSession-related commits. @hmelder would be great if you can amend a change log entry.

PR #414, opened 2024-05-30
fix: wrap aFormat with NSNumber

chenrui333:
seeing build failure as below:

```
  NSPropertyList+GNUstepBase.m:645:33: error: incompatible integer to pointer conversion passing 'NSPropertyListFormat' (aka 'enum NSPropertyListFormat') to parameter of type 'id' [-Wint-conversion]
      (self, _cmd, aPropertyList, aFormat, anErrorString);
                                  ^~~~~~~
  1 error generated.
```

relates to https://github.com/Homebrew/homebrew-core/pull/173267


rfm (commented):
The original code is correct but the type cast is wrong.  So as long as the compiler is told to warn rather than abort on a type cast issue, it should compile and work.

I changed the incorrect type cast in the git repository.

PR #371, opened 2024-02-05
Use absolute paths in ./configure

qmfrederik:
This caused errors when trying to run `./configure` outside of the libs-base directory (to do out-of-tree builds).  There were some further errors, but these changes seemed meaningful nonetheless.

PR #415, opened 2024-05-31
Use absolute paths in ./configure

qmfrederik:
This caused errors when trying to run ./configure outside of the libs-base directory (to do out-of-tree builds).

Originally submitted as #371, which was closed without merging.


rfm (approved):
That all looks good.

PR #416, opened 2024-06-04
Fix broken enumerateSubstringsInRange:options:usingBlock: implementation

hmelder:
The ICU ubrk API was used wrongly and rendered this method completely useless.


hmelder:
https://github.com/gnustep/libs-base/pull/370

PR #419, opened 2024-06-10
Implement -[NSFileManager URLsForDirectory:inDomains:]

2xsaiko:
This behaves like NSSearchPathForDirectoriesInDomains (at least, superficially. Not sure if Apple's implementation has some special behavior on top that their documentation doesn't mention.)

The documentation comment is copied from NSSearchPathForDirectoriesInDomains.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2024-06-10 Marco Rebhan <me@dblsaiko.net>
+
+  * Headers/Foundation/NSFileManager.h:
+  * Source/NSFileManager.m:
+  Implement -[NSFileManager URLsForDirectory:inDomains:].
+
 2024-06-10 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Source/GSFileHandle.m: debug log of closing a closed file handle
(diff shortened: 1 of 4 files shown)


rfm (changes requested):
The concept is fine, but there are two show-stopping issues of non-portable coding with the implementation:

1. The use of generics:  you need to replace the generic syntax with the macros provided in NSObjCRuntime.h
2. The use of blocks:  you need to replace any blocks code with portable code.  In this case it's simple enumeration so you might best use the FOR_IN macro provided in GSFastEnumeration.h

gcasa:
@2xsaiko I don't believe a copyright assignment is needed for this change.  Please write assign@gnu.org and request a copyright assignment for GNUstep and CC me on the request.   Also, please make the changes suggested via @rfm.  The reason is that GNUstep supports more compilers than just clang and not all support the same features.  Thank you, GC

gcasa:
Sorry, I accidentally hit the "close" button.  I reopened this.

2xsaiko:
Done!

rfm (approved):
One tiny issue; the dot syntax to access a property is non-portable; older compilers don't support properties, so we must call the method to get the count of objects in the paths array.

However, I'll merge and then fix that rather than holding things up.

rfm:
oops ... also an excess RELEASE to be removed and some formatting tweaks

PR #421, opened 2024-07-02
Revert "NSURLSession Reimplementation"

rfm:
Reverts gnustep/libs-base#411


rfm:
Revert accidental merge of unfinished rewrite breaking build
.

rfm:
Replacement for the existing implementation needs to actually build and work with all the common libdispatch implementations packaged with many systems (or conditionally be used with a configure time check for libdispatch capabilities).  That includes the testcases.
For instance, dispatch_cancel() is non-portable, as are the dispatch_io... functions.

hmelder:
> with all the common libdispatch implementations packaged with many systems  [...]
> For instance, dispatch_cancel() is non-portable, as are the dispatch_io... functions.

Everything other then https://github.com/apple/swift-corelibs-libdispatch is a fork. I am not aware of any base API being unavailable (apart from whole OSes not being supported).

libdispatch currently supports macOS, FreeBSD, Linux, and Windows

PR #422, opened 2024-07-03
NSURLSession rewrite

rfm:


rfm:
I noticed that the CI / Ubuntu x64 Clang gnustep-2.0 (push) build is working for this branch even though the code fails to build on Ubuntu normally ... that's because the integration is importing and building a version of libdispatch from source rather than using the version of libdisatch packaged  with Ubuntu.  We should improve on that.

hmelder:
> rather than using the version of libdisatch packaged with Ubuntu

`libdispatch0` which is packaged in 20.04LTS and 22.04LTS and removed in newer Ubuntu versions is sourced from `http://svn.macosforge.org/repository/libdispatch/trunk/`. It is version `r197` which was imported over 13 years ago.

```
libdispatch (0~svn197-1) unstable; urgency=low

  * Upgrade to r197 from the upstream source repository.
  * Remove all Debian-specific patches.
  * Remove libbsd as a dependency.
  * Add a dependency on libpthread-workqueue.

 -- [...] Sun, 6 Mar 2011 15:31:39 -0400
 ```

**Comment by hmelder** [comment]
Fedora, OpenSUSE, FreeBSD, EPEL, Arch and Alpine Linux are somewhat up-to-date and will work just fine.

**Comment by rfm** [comment]
Most installed versions of those distributions, along with others (the mainstream RedHat enterprise, CentOS, Alma, Debian etc) have the older version and probably will be dominant for some years to come.  I assume you would like improved NSURLSession code in the next release rather than (for instance) waiting several more years until the older installs have gone out of use.
So I've done some quick fixes to at least get this branch to compile and link again.  Of course it will be somewhat broken by the lack of dispatch_cancel() ... what alternative strategy is there?  Perhaps a way to get the timer to fire harmlessly if the session has been deallocated?

**Comment by rfm** [comment]
> waiting several more years until the older installs have gone out of use.
To be fair, that's a bit polemical/rhetorical:  we have the alternative of doing what we occasionally do, and saying we are breaking compatibility by dropping support for something old (we deprecate the old version).  The normal process for that is to announce it in once release and do it in the next release (so typically after 2 years).  I'd like to see the new code available faster than that.

**Comment by hmelder** [comment]
> the lack of dispatch_cancel()

is `dispatch_source_cancel()` available?

> Perhaps a way to get the timer to fire harmlessly if the session has been deallocated?

That seems reasonable.

**Comment by rfm** [comment]
> On 3 Jul 2024, at 11:54, Hugo Melder ***@***.***> wrote:
> 
> 
> the lack of dispatch_cancel()
> is dispatch_source_cancel() available?

Yes, dispatch_source_cancel() exists in the old version.
I don't know if they can be used interchangably for a timer though. How does dispatch_cancel() differ from it?

**Comment by hmelder** [comment]
> I don't know if they can be used interchangably for a timer though. How does dispatch_cancel() differ from it?

A dispatch timer is just a dispatch_source.

This is the definition of dispatch_cancel:
```objc
void
dispatch_cancel(void *object);
#if __has_extension(c_generic_selections)
#define dispatch_cancel(object) \
		_Generic((object), \
			dispatch_block_t:dispatch_block_cancel, \
			dispatch_source_t:dispatch_source_cancel \
		)((object))
#endif
```

Source/NSURLSession.m
```diff
@@ -921,6 +921,8 @@ - (void)dealloc
 
 #if	defined(HAVE_DISPATCH_CANCEL)
   dispatch_cancel(_timer);
+#else
```
hmelder, line 924: We can just use `dispatch_source_cancel` without the ifdef. 

The timer is under the `dispatch_source` prefix: https://developer.apple.com/documentation/dispatch/1385606-dispatch_source_set_timer

rfm:
OK, updated to use that if dispatch_cancel is not available.
I think/hope the code should work with those changes,  leaving only the problem of testing.
The main issue there seems to be that the HTTP server helper is completely non-portable, so that doesn't look as easy to fix.
I'm not familiar with either set of test cases, but I see that the NSURLConnection Helpers directory contains a portably coded server bundle with support for HTTPS as well as HTTP, so perhaps that could/should be modified and used for testing NSURLSession too.

rfm:
* We can just use dispatch_source_cancel without the ifdef.
That sounds good.

rfm:
FYI, I have to go out for a while now, so won't be able to respond for a bit.

hmelder:
>  so perhaps that could/should be modified and used for testing NSURLSession too.

I can make the HTTP server in the unit test more portable. Besides dispatch_io, what else is non-portable here? Just plain BSD sockets and WINSOCK on Win32.

>  but I see that the NSURLConnection Helpers directory contains a portably coded server bundle with support for HTTPS as well as HTTP
I do not like the architecture of it.

hmelder:
> So NSFileHandle and related gnustep extensions are probably the best (oldest and most reliably working on all platforms) foundation for networking test code. Directly using socket code comes close though.

I'll rewrite the webserver to use NSFileHandle instead of raw sockets then.

hmelder:
Should this be part of this PR or merged later on?

rfm:
I think it should all be done as one ideally.

Otherwise, anyone who is tracking the git repository to keep their codebase up to date will find the testcases crashing if their system happens to have the older libdispatch.

That's not really a big deal (because anyone tracking the latest changes on a frequent basis ought to be an able enough developer to handle that), but it gives a better impression not to make things unstable if we can help it.

rfm:
Any progress on this?  If using NSFileHandle is too much of a hassle, I think plain socket code is portable enough: as long as the tests run on all major platforms with the libdispatch they have been using for the last several years.

hmelder:
> If using NSFileHandle is too much of a hassle, I think plain socket code is portable enough

The server currently relies on the concurrency of libdispatch. I guess it is possible to rewrite it to only use NSFileHandle, but I couldn't get myself motivated just yet ;)

> as long as the tests run on all major platforms with the libdispatch they have been using for the last several years.

So replacing dispatch io with just source objects? Or are you ignoring the ancient Ubuntu and Debian versions now?

rfm:
I meant keeping libdispatch internal to the library, so the only portability issue in the tests would be the socket api.

hmelder:
I rewrote parts of the test HTTP server to use dispatch_source instead of dispatch_io.

While doing that, I found a nastly bug where [curl_easy_getinfo()](https://curl.se/libcurl/c/curl_easy_getinfo.html) writes a long to a local NSInteger variable with sizeof(intptr_t).

The local variable was never initialised. It was not a problem on Linux as sizeof(long) = 8 on LP-64 platforms, but things are different on Windows (sizeof(long) = 4).

The compiler did not catch this as `curl_easy_getinfo` is a variadic function.

hmelder:
Ready for review @rfm

rfm:
It's great to have spotted the windows/curl issue.

Unfortunately, the test code changes (tweaking the use of libdispatch rather than removing it entirely) aren't sufficient as the test web server fails to compile (at least on stock Ubuntu Jammy Jellyfish AMD64 architecture).

Here's the latest compiler output:

HTTPServer.m:68:37: warning: implicit declaration of function 'dispatch_data_get_size' is invalid in C99 [-Wimplicit-function-declaration]
    [NSMutableData dataWithCapacity:dispatch_data_get_size(dispatchData)];
                                    ^
HTTPServer.m:70:3: warning: implicit declaration of function 'dispatch_data_apply' is invalid in C99 [-Wimplicit-function-declaration]
  dispatch_data_apply(dispatchData, ^bool(dispatch_data_t region, size_t offset,
  ^
HTTPServer.m:70:43: error: type-id cannot have a name
  dispatch_data_apply(dispatchData, ^bool(dispatch_data_t region, size_t offset,
                                          ^~~~~~~~~~~~~~~~
HTTPServer.m:70:59: error: expected ')'
  dispatch_data_apply(dispatchData, ^bool(dispatch_data_t region, size_t offset,
                                                          ^
HTTPServer.m:70:42: note: to match this '('
  dispatch_data_apply(dispatchData, ^bool(dispatch_data_t region, size_t offset,
                                         ^
HTTPServer.m:72:30: error: use of undeclared identifier 'buffer'
    [mutableData appendBytes:buffer length:size];
                             ^
HTTPServer.m:72:44: error: use of undeclared identifier 'size'
    [mutableData appendBytes:buffer length:size];
                                           ^
HTTPServer.m:190:61: warning: incompatible pointer types passing 'NSUInteger *' (aka 'unsigned long *') to parameter of type 'socklen_t *' (aka 'unsigned int *') [-Wincompatible-pointer-types]
  if (getsockname(_socket, (struct sockaddr *) &serverAddr, &addrLen) == -1)
                                                            ^~~~~~~~
/usr/include/x86_64-linux-gnu/sys/socket.h:117:26: note: passing argument to parameter '__len' here
                        socklen_t *__restrict __len) __THROW;
                                              ^
HTTPServer.m:206:40: error: use of undeclared identifier 'DISPATCH_QUEUE_CONCURRENT'
                                       DISPATCH_QUEUE_CONCURRENT);
                                       ^
HTTPServer.m:219:67: warning: incompatible pointer types passing 'NSUInteger *' (aka 'unsigned long *') to parameter of type 'socklen_t *' (aka 'unsigned int *') [-Wincompatible-pointer-types]
  clientSocket = accept(_socket, (struct sockaddr *) &clientAddr, &sin_size);
                                                                  ^~~~~~~~~
/usr/include/x86_64-linux-gnu/sys/socket.h:307:28: note: passing argument to parameter '__addr_len' here
                   socklen_t *__restrict __addr_len);
                                         ^
HTTPServer.m:230:5: warning: implicit declaration of function 'close' is invalid in C99 [-Wimplicit-function-declaration]
    close(clientSocket);
    ^
HTTPServer.m:408:3: warning: implicit declaration of function 'close' is invalid in C99 [-Wimplicit-function-declaration]
  close(_socket);
  ^
6 warnings and 6 errors generated.

Tests/base/NSURLSession/Helpers/HTTPServer.m
```diff
@@ -0,0 +1,414 @@
...
+    [NSMutableData dataWithCapacity:dispatch_data_get_size(dispatchData)];
+
+  dispatch_data_apply(dispatchData, ^bool(dispatch_data_t region, size_t offset,
+                                          const void *buffer, size_t size) {
+    [mutableData appendBytes:buffer length:size];
+    return true; // Continue iterating
+  });
+
+  return [mutableData copy];
+}
```
hmelder, line 77: This function is never used and can be deleted

hmelder:
> HTTPServer.m:206:40: error: use of undeclared identifier 'DISPATCH_QUEUE_CONCURRENT'
DISPATCH_QUEUE_CONCURRENT);

I think `DISPATCH_QUEUE_CONCURRENT` is not exposed in ancient versions of libdispatch. The dispatch_data* conversion function can be deleted and is a left over.

hmelder:
I guess I need to use an NSOperationQueue then...

rfm:
Is DISPATCH_QUEUE_CONCURRENT needed?    If I replace it with 0 (and remove the unused code) it compiles.

The tests mostly appear to pass ... apart from an exception trying to launch gdnc, which may be unrelated.

Once I launched gdnc manually, the tests all passed

hmelder:
> Is DISPATCH_QUEUE_CONCURRENT needed?

You can execute it sequentially as well. I just wanted to test parallel tasks in NSURLSession.

> The tests mostly appear to pass ... apart from an exception trying to launch gdnc, which may be unrelated.

It would be good to use https://github.com/gnustep/libs-base/blob/04f0d769043fff7859f95e1237fc52e095aa784f/Source/GSDispatch.h#L49 the compat macro instead so that we can make use of `DISPATCH_QUEUE_CONCURRENT` when available.

hmelder:
> Once I launched gdnc manually, the tests all passed

Weird. I am not using gdnc anywhere in the code.

hmelder:
@rfm can you test it with the compat macro?

rfm:
Well it works with literal 0,  which is the same value as the compat macro.

hmelder:
> Well it works with literal 0, which is the same value as the compat macro.

So is it ready to merge then?

rfm:
I think we can merge ... it's a big change, so there are probably issues which will surface after merging, but if we don't merge we might never find them.

PR #426, opened 2024-07-30
NSUserDefaults: Retrieve native UI languages from Windows and Android System APIs

hmelder:
Note that `GetUserPreferredUILanguages` was introduced in Windows Vista.

### DIFF
--- a/Headers/GNUstepBase/GSConfig.h.in
+++ b/Headers/GNUstepBase/GSConfig.h.in
@@ -304,7 +304,7 @@ typedef	struct {
 
 #  if defined(__MINGW__)
 #    include <w32api.h>
-#    define GS_WINVER Windows2000
+#    define GS_WINVER WindowsVista
 #  elif defined(_MSC_VER)
 #    include <WinSDKVer.h>
 #    define GS_WINVER _WIN32_WINNT_WIN10
--- a/Source/NSProcessInfo.m
+++ b/Source/NSProcessInfo.m
@@ -1629,7 +1629,9 @@ GSInitializeProcessAndroid(JNIEnv *env, jobject context)
         localeList = strdup(localeListOrig);
       }
 
-      // FIXME: This is an IETF BCP 47 language tag and may not correspond to ll-CC format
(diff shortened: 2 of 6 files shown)


rfm (approved):
I like all this, but I have some backward compatibility reservation with the switch from Windows 2000 to Vista (unless I've missed something this code won't build on Windows 2000, but isn't protected by conditional compilation to be ignored on older versions).

Vista is old enough that I've no problem with the switch in principle, but we didn't say we would be making that change in the last release, and there may be many other things we want to update too.  For instance I see that configure.ac sets the default installation locations to paths suitable for older versions of windows, with a note to say that this should change for Vista:

    # TODO: Improve this hack.
    # According to Wikipedia, this is the default for Windows 2000,
    # Windows XP and Windows Server 2003.  For Windows Vista this will
    # change to C:\Users.  The directory name needs to be localized though
    # (and the disk may need changing as well ?).

So, I think I would tentatively approve this change, but we ought to also make any other changes appropriate for moving to Vista as a minimum version before the next release, and perhaps we should advertise (eg on the discussion list and website) that we are dropping support for earlier versions of Windows.

hmelder:
> but we ought to also make any other changes appropriate for moving to Vista as a minimum version before the next release, and perhaps we should advertise 

I can do this :) We should track this in an issue as well.

hmelder:
> So, I think I would tentatively approve this change

So should I squash and merge, or leave it open for now?

hmelder:
@rfm

rfm:
There was quite some pushback against moving away from XP,  but that was because there were a *lot* of old XP systems in use many years after official end-of-life (becaus XP was seen as much more stable/reliable).
I don't think the same is at all true of windows 2000, so dropping support there is much easier, and I expect few (if any) people building from git master, rather than a release, will be using such an old system.

So I think it's OK to merge.

rmottola:
Just as a drop-in comment. Yes Win-XP is incredibly ubiquitous and I randomly use(d) GNUstep on it. I do not have windows 2000 anymore to test, but if memory serves me we broke support for it last time I tried and there weren't many complaints.

PR #430, opened 2024-08-07
Agsdoc new style

rmottola:
Improvements to AutoGSDoc to enhance CSS styling and HTML5 compatibility

### DIFF
--- a/Tools/AGSHtml.m
+++ b/Tools/AGSHtml.m
@@ -648,7 +648,7 @@ static NSMutableSet	*textNodes = nil;
 	}
       else if ([name isEqual: @"br"] == YES)
 	{
-	  [buf appendString: @"<br />"];
+	  [buf appendString: @"<br/>"];
 	}
       else if ([name isEqual: @"category"] == YES)
 	{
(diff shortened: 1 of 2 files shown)


rmottola:
@rfm not yet done keeping this open for  further improvements, but so we can comment and collaborate on it and comment & keep track. When you feel enough we can merge.
We could also add Frame work on this branch

rfm (approved):
Those changes all look good, but I expect you will need to rebase because I committed other changes to implement basic graphing of class/superclass using graphviz.

rmottola:
@rfm I found that the main .html is declared as xhtml and uses XHTML styles in a mix with HTML. I think that is a little strange, especially today, would you agree to cleanup to HTML?

rfm:
Definitely a general cleanup would be a good idea.    I think we want to update the declared versions of files to be html 5 throughout.
  

  

  

  
  
  
  
  
>

rfm:
I didn't see this branch until after committing autogsdoc graphing changes.  Perhaps I should do further changes here?

rmottola:
@rfm there are no merge conflicts, so we were lucky, since it is one specific change I think it can go in, I merged.
Best to make a new common branch for style improvements!

PR #431, opened 2024-08-09
Modernise Windows Paths in `configure.ac` and use 'short' output of  relative_paths.sh by default

hmelder:
As we are planning to remove support for Windows Versions older than Vista, we can now change `C:\Documents and Settings` to `C:\Users`.

The move to short relatives paths was motivated by a `TODO` tag in `configure.ac`.

Additionally, some m4 macros were modernised.


rfm (approved):
This looks good to me.  Thanks.

Issue #432, opened 2024-08-09
`-[NSArray removeLastObject]` throws exception when array is empty


ERobsham:
The GNUStep implementation of `-[NSArray removeLastObject]` differs from the Apple implementation.   In GNUStep, exceptions are thrown when an array is empty and the `-removeLastObject` method is called:
``` objective-c
- (void) removeLastObject
{
  NSUInteger	count = [self count];

  if (count == 0)
    [NSException raise: NSRangeException
		 format: @"Trying to remove from an empty array."];
  [self removeObjectAtIndex: count-1];
}
```

In Apple implementations, the `-removeLastObject` method just silently exits when there are no elements to remove.

To improve compatibility for codebases used across both implementations, I think this exception raise should be dropped in favor of just a return.

ERobsham:
Changes merged to prevent raising exceptions with empty arrays -- closing.

PR #433, opened 2024-08-10
Update `-[NSMutableArray removeLastObject]`

ERobsham:
Updates the `-[NSMutableArray removeLastObject]` implementation to avoid raising an exception if the array is already empty. 

This brings the behavior of `-[NSMutableArray removeLastObject]` inline with the Apple implementation, improving compatibility for code bases which use both GNUStep and Apple objc/Foundation implementations.

RE: https://github.com/gnustep/libs-base/issues/432

### DIFF
--- a/Source/GSArray.m
+++ b/Source/GSArray.m
@@ -669,8 +669,7 @@ static Class	GSInlineArrayClass;
   _version++;
   if (_count == 0)
     {
-      [NSException raise: NSRangeException
-		  format: @"Trying to remove from an empty array."];
+      return;
     }
   _count--;
   RELEASE(_contents_array[_count]);
--- a/Source/NSArray.m
+++ b/Source/NSArray.m
@@ -2187,16 +2187,14 @@ compare(id elem1, id elem2, void* context)
 }
 
 /**
- * Removes the last object in the array.  Raises an exception if the array
- * is already empty.
+ * Removes the last object in the array if one exists (otherwise it just returns).
  */
 - (void) removeLastObject
 {
   NSUInteger	count = [self count];
 
   if (count == 0)
-    [NSException raise: NSRangeException
-		 format: @"Trying to remove from an empty array."];
+    return;
   [self removeObjectAtIndex: count-1];
 }
 
--- a/Source/NSKeyValueMutableArray.m
+++ b/Source/NSKeyValueMutableArray.m
@@ -158,7 +158,12 @@
 
 - (void) removeLastObject
 {
-  [self removeObjectAtIndex: ([self count] - 1)];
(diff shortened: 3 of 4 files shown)


fredkiefer (approved):
Looks good to me.

fredkiefer:
Maybe we also should a a small test case to prevent this behaviour from getting broken later on again.

ERobsham:
Well this is interesting...
The added test is failing with the original exception still being thrown..? 

(from the `Ubuntu x64 GCC` logs):
```
Failed test:     (2024-08-10 17:56:32.810 +0000) general.m:122 ... -removeLastObject does not raise exceptions on empty array
NSRangeException: <NSException: 0x55f3d1fdeae0> NAME:NSRangeException REASON:Trying to remove from an empty array. INFO:(null)Completed file:  general.m
```

fredkiefer:
This is really strange, maybe some caching issue? I will try to get the interest of the other GNUstep maintainers on this.

fredkiefer:
There is a second implementation of the `removelastobject` method on GSArray and this one is actually used. You need to change this as well. Great that you wrote the test :-)

ERobsham:
Well that makes _a lot_ more sense...!

fredkiefer:
There is also another implementation in NSKeyValueMutableArray.m. But maybe @rfm could have a look whether this also needs to be adjusted.

ERobsham:
Okay, went through and updated all the instances I found searching for: `- (void) removeLastObject`

I _think_ it makes sense to update the `NSKeyValueMutableArray.m` implementations too to avoid sending notifications to observers that are essentially noops.  
Additionally, if those weren't updated, we'd have situations where mutable arrays being watched via KVO behave differently from non-observed arrays ... which feels wrong.

ERobsham:
Looks like the two CI failures are unrelated to the change.

CI / Windows x64 MinGW GCC:
```
../GSPThread.h:40:3: sorry, unimplemented: '_Atomic' in Objective-C
   40 |   _Atomic(DWORD) owner;
      |   ^~~~~~~
make[4]: *** [/d/a/libs-base/libs-base/build/share/GNUstep/Makefiles/rules.make:534: obj/Additions.obj/GSObjCRuntime.m.o] Error 1
```

CI / Windows x86 MinGW GCC:
```
Installing additional packages through pacman...
  C:\Windows\system32\cmd.exe /D /S /C D:\a\_temp\setup-msys2\msys2.cmd -c "'pacman' '--noconfirm' '-S' '--needed' '--overwrite' '*' 'mingw-w64-i686-gcc-objc'"
  error: target not found: mingw-w64-i686-gcc-objc
```

Issue #429, opened 2024-08-07
MinGW CI Broken


hmelder:
It seems like the MinGW CI broke after an Update a few days ago...

hmelder:
`mingw-w64-i686-gcc-objc` was removed in the GCC 14.1.0 msys2 package update.

We can either remove the i686 CI job, or ask for a revert. @rfm what do you think?

Additionally, we now falsely enable Win32 native threads and locking on MinGW GCC after [this](https://github.com/gnustep/libs-base/commit/a2b9be628773831a273ffb975081df6106cb6e91) update.

PR #434, opened 2024-08-11
Remove MinGW 32-bit CI

hmelder:
MSYS2 is slowly phasing out 32-bit support and removed gcc-objc as part of the 14.1.0 update. If someone is interested in keeping 32-bit support in CI, self-build binaries must be supplied.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -181,14 +181,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


hmelder:
See #429

rfm (approved):
While we like to keep a high degree of backward compatibility, we really don't have enough people to build mingw/msys stuff that the maintainers of the projects don't support, so I agree with removing it.  If anyone is willing to put in thew effort they can always re-instate it.

PR #435, opened 2024-08-12
Make gs_mutex_* APIs public

hmelder:


rfm:
I'm unsure about this ... generally I've been trying to simplify things by deprecating/removing additions and exposed implementation details that don't seem to be actively used, but removing something we have added is a lot harder/slower than it was to add it.
So exposing this file is going against the trend, and we need to be confident.
I guess the idea is to provide a portable public mechanism to access simple mutex operations in the most efficient way possible for apps where performance is paramount?   Are you confident that what you are exposing contains what's needed, and nothing more?

fredkiefer:
I have to agree with Richard. This header file is not a great abstraction to export to the outside world. If we do so, we will depend on supporting it for a long time. We rather should try to find out which specific functionality is really required for application and if some of it is not properly supported by NSThread or the lock classes, try to find a suitable abstraction and only provide that.

PR #436, opened 2024-08-12
NSThread: Fix `threadPriority` and `setThreadPriority:` on Android by using `setpriority` instead of pthread

hmelder:
Android's pthread_setschedparam is currently broken, as it checks if the priority is in the range of the system's min and max priorities. The interval bounds are queried with `sched_get_priority_min`, and `sched_get_priority_max` which just return 0, regardless of the specified scheduling policy.
  
   The solution is to use `setpriority` to set the thread priority. This is possible because on Linux, it is not a per-process setting as specified by POSIX but a per-thread setting (See the `Bugs` section in `setpriority`).

### DIFF
--- a/Source/NSThread.m
+++ b/Source/NSThread.m
@@ -116,6 +116,11 @@ typedef struct {
 #  include <sys/file.h>
 #endif
 
+#if defined(__ANDROID__)
+#  include <string.h>       // For strerror
+#  include <sys/resource.h> // For getpriority and setpriority
+#endif
+
 #if	defined(HAVE_SYS_FCNTL_H)
 #  include <sys/fcntl.h>
 #elif	defined(HAVE_FCNTL_H)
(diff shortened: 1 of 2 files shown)


rfm (approved):
This makes perfect sense.

hmelder:
@rfm how about doing that on Linux as well? Currently we do not prioritise anything because of this.

hmelder:
Here is a similar discussion: https://github.com/dotnet/runtime/issues/91165

triplef:
Another discussion about this in the NDK project: https://github.com/android/ndk/issues/1255

Source/NSThread.m
```diff
@@ -1028,8 +1028,9 @@ + (BOOL) setThreadPriority: (double)pri
   if (pri > 1) { pri = 1; }
   if (pri < 0) { pri = 0; }
 
-  // Convert [0.0, 1.0] to [-20, 19] range
-  int priority = (int)(-20 + pri * 39);
+  // Convert [0.0, 1.0] to [-20, 19] range where -20 is the highest
+  // and 19 the lowest priority.
+  int priority = (int)(-20 + (1-pri) * 39);
```
triplef, line 1033: Do we need to do the same conversion when reading the priority?

PR #438, opened 2024-08-14
Add Branch Prediction Macros and Error checks in Android Initialisation

hmelder:
We have to discuss what should happen when a JNI call fails early during process initialisation. This should result in an abort IMO, as the environment is somehow corrupted.


rfm (approved):
I'm not in a position to really check this (having no android dev system), but from what I remember of using JNI for GNUstep (I worked on the JIGS integration Nicola did some years ago) the checks look reasonable/sane.

triplef (approved):
Looks good, thanks!

PR #442, opened 2024-08-29
Add some missing NSSearchPathDirectory enum types

buzzdeee:
As found here: https://web.archive.org/web/20131113010445/https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Constants/Reference/reference.html


rfm (commented):
While this looks relatively harmless, the fact that the new types are not implemented would cause potential confusion/breakage of apps that tried to use them, so I think it would be better to support those values in NSPathUtilities.m rather than ignoring them.

rfm:
I created rudimentary implementations for the directory types, so this patch is obsolete.

PR #443, opened 2024-08-29
Add some missing NSURLErrorDomain constants, as described here:

buzzdeee:
https://web.archive.org/web/20131113010445/https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Constants/Reference/reference.html


rfm (approved):
This seems harmless/good

PR #445, opened 2024-09-07
NSKeyValueCoding: Safe-Caching for -[NSObject valueForKey:]

hmelder:
It turns out that valueForKey: is a very expensive operation, and a major bottleneck for Key-Value Observing and other operations such as sorting an array by key.

The accessor search patterns for Key-Value observing are discussed in the Apple Key-Value Coding Programming Guide. The return value may be encapsulated into an NSNumber or NSValue object, depending on the Objective-C type encoding of the return value. This means that once valueForKey: found an existing accessor, the Objective-C type encoding of the accessor is retrieved. We then go through a huge switch case to determine the right way to invoke the IMP and potentially encapsulate the return type. The resulting object is then returned.
The algorithm for setValue:ForKey: is similar.

We can speed this up by caching the IMP of the accessor in a hash table. However, without proper versioning, this quickly becomes very dangerous. The user might exchange implementations, or add new ones expecting the
search pattern invariant to still hold. If we clamp onto an IMP, this invariant no longer holds.

Apple actually just caches the IMP without versioning, as there is no support for safe-caching in `objc4`.

We will make use of libobjc2's safe caching to avoid this.
Note that the caching is opaque. You will only need to redirect all
valueForKey: calls to the function below.

Note that the Implementation requires `method_getTypedSelector_np` which was added to `libobjc2` here: https://github.com/gnustep/libobjc2/commit/f7a38aa9979175f61e476f6d3a7e2d15aa50ba33.


Source/Additions/GSObjCRuntime.m
```diff
@@ -1334,7 +1334,7 @@ unsigned long long	(*imp)(id, SEL) =
                   }
                 val = [NSValue valueWithPoint: v];
               }
-            else if (GSSelectorTypesMatch(@encode(NSRange), type))
+            else if (strncmp(@encode(NSRange), type, strlen(@encode(NSRange))) == 0)
```
fredkiefer, line 1337: I am not sure about the other types but at least for NSRange we need to make sure that CFRange may be used interchangeable. This really is a rare case but we should not break things.

Tests/base/KVC/safe_caching.m
```diff
@@ -0,0 +1,66 @@
+#import <Foundation/NSKeyValueCoding.h>
+#import <Foundation/NSValue.h>
+#import <objc/runtime.h>
+
+#import "Testing.h"
+
+// Caching is only available for libobjc2
+#if defined(__OBJC2__)
```
fredkiefer, line 8: I don't understand why we cannot run this test with an old compiler. There isn't much need for it, but without caching this should just work out of the box.

fredkiefer:
Any idea why the new type test fails for gcc and would it pass without your changes to GSRuntime?

Source/Additions/GSObjCRuntime.m
hmelder: That makes sense. But there is also CGSize, CGRect, and CGPoint. The current approach is also not great and may lead to weird side-effects.

hmelder:
> Any idea why the new type test fails for gcc and would it pass without your changes to GSRuntime?

Not sure. I'd need to setup a GCC environment to debug this.
```sh
Failed test:     (2024-09-07 10:06:20.004 +0000)   types.m:312 ... Ivar returns NSPoint
Failed test:     (2024-09-07 10:06:20.004 +0000)   types.m:314 ... Ivar returns NSRange
Failed test:     (2024-09-07 10:06:20.004 +0000)   types.m:316 ... Ivar returns NSRect
Failed test:     (2024-09-07 10:06:20.004 +0000)   types.m:318 ... Ivar returns NSSize
Failed test:     (2024-09-07 10:06:20.004 +0000)   types.m:320 ... Ivar returns MyStruct
```

Source/Additions/GSObjCRuntime.m
hmelder: What about skipping the first two characters (NS, CF, CG) and just compare the remaining string?

hmelder: A bit ugly but it will work

fredkiefer: You mean test that the encode type starts with '{' and leave out the next two characters? Sounds good enough for me.

hmelder: It turns out that CFRange and NSRange are not toll-free bridged on macOS. For instance, you cannot pass a CFRange to +[NSValue valueWithRange:]. Additionally, all geometry structs in Foundation are actually CoreGraphics structs. We should use the same naming scheme to support interop.

```c
#import <CoreFoundation/CoreFoundation.h>
#import <CoreGraphics/CoreGraphics.h>

int main() {
	@autoreleasepool {
		NSLog(@"%s", @encode(NSRect));
		NSLog(@"%s", @encode(CGRect));
		NSLog(@"%s", @encode(NSPoint));
		NSLog(@"%s", @encode(CGPoint));
		NSLog(@"%s", @encode(NSSize));
		NSLog(@"%s", @encode(CGSize));
		NSLog(@"%s", @encode(NSRange));
		NSLog(@"%s", @encode(CFRange));

		CGRect cgRect = CGRectMake(0, 1, 1, 1);
		CFRange range = CFRangeMake(0, 1);
		NSValue *val = [NSValue valueWithRange: range];
		NSLog(@"%@", val);
	}
	return 0;
}
```

```sh
hmelder@Hugos-MacBook-Pro ~ % clang test.m -framework Foundation -framework CoreFoundation -framework CoreGraphics
test.m:18:43: error: sending 'CFRange' to parameter of incompatible type 'NSRange' (aka 'struct _NSRange')
                NSValue *val = [NSValue valueWithRange: range];
                                                        ^~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSRange.h:45:38: note: passing argument to parameter 'range' here
+ (NSValue *)valueWithRange:(NSRange)range;
                                     ^
1 error generated.
```

And otherwise

```sh
2024-09-10 13:11:46.478 a.out[54713:2550694] {CGRect={CGPoint=dd}{CGSize=dd}}
2024-09-10 13:11:46.479 a.out[54713:2550694] {CGRect={CGPoint=dd}{CGSize=dd}}
2024-09-10 13:11:46.479 a.out[54713:2550694] {CGPoint=dd}
2024-09-10 13:11:46.479 a.out[54713:2550694] {CGPoint=dd}
2024-09-10 13:11:46.479 a.out[54713:2550694] {CGSize=dd}
2024-09-10 13:11:46.479 a.out[54713:2550694] {CGSize=dd}
2024-09-10 13:11:46.479 a.out[54713:2550694] {_NSRange=QQ}
2024-09-10 13:11:46.479 a.out[54713:2550694] {?=qq}
```

I've renamed all structs in NSGeometry.h accordingly.

hmelder:
@rfm, @fredkiefer, it would be great if you could take another look at it.

Headers/Foundation/NSGeometry.h
```diff
@@ -97,8 +97,8 @@ typedef NSSize *NSSizePointer;
 }</example>
 
  <p>Rectangle.</p> */
-typedef struct _NSRect NSRect;
-struct _NSRect
```
triplef, line 101: Since we’re now defining `struct CGPoint/CGSize/CGRect`, can we please also add the following typedefs to match `CFCGTypes.h` from Apple’s CoreFoundation:

```c
typedef struct CGPoint CGPoint;
typedef struct CGSize CGSize;
typedef struct CGRect CGRect;
```

Headers/Foundation/NSGeometry.h
```diff
@@ -97,8 +97,8 @@ typedef NSSize *NSSizePointer;
...
+typedef struct CGRect NSRect;
+struct CGRect
```
triplef, line 101: Since we’re now defining `struct CGPoint/CGSize/CGRect`, can we please also add the following typedefs to match `CFCGTypes.h` from Apple’s CoreFoundation:

```c
typedef struct CGPoint CGPoint;
typedef struct CGSize CGSize;
typedef struct CGRect CGRect;
```

hmelder: I still don't like the idea that we do this at all, because it will lead to redefinitions.

hmelder: I moved all CG definitions in a CoreFoundation header file. It is not ideal that we install this as part of libs-base, but I cannot think of a better way to do this without major project restructuring.

hmelder:
@rfm it would be great if you could take another look at this PR :)
I've moved some known CoreGraphics struct types into their proper header to improve compatibility with Apple Frameworks.

rfm (approved):
All seems to be reasonable to me.

PR #446, opened 2024-09-07
Add getOffset:error: implementation to base

gcasa:
This PR adds the 10.15 method getOffset:error: to NSFileHandle.


rfm (commented):
As far as I can see the only point of this method is to provide error information in the case where the underlying device doesn't support the idea of an offset (or is closed), but the implementation doesn't provide that information ...

gcasa:
> As far as I can see the only point of this method is to provide error information in the case where the underlying device doesn't support the idea of an offset (or is closed), but the implementation doesn't provide that information ...

I am trying to figure out where and what error it should add.  I think I am going to move this back to Draft status until I do.

rfm:
added with a few other methods

PR #460, opened 2024-11-08
fix for builds without ICU

rmottola:


fredkiefer:
Richard has already fixed this. You should get a conflict for your change. And they are actually shown. Best to close this pull request.

PR #462, opened 2024-11-13
Use Bundle identifier for userdefaults

williameveretteggplant:


rfm (commented):
While using the bundle identifier is correct for OSX compatibility, the implementation is completely the wrong approach ... copying code is a maintainability issue, and trying to minimise the copied code by not actually duplicating the correct/full logic for resource lookup over all supported environment only makes  it worse.  A better approach is to expose the functionality from NSBundle rather than duplicating it, and that looked simple enough for me to do in master.

williameveretteggplant:
Question is, can you get the bundle identifier from NSBundle without invoking NSUserDefaults?

rfm:
Sure, because the Info file is not locale-specific, so we don't need the locale preferences.

PR #468, opened 2024-11-18
Added many missing features to NSGeometry

Matvii-Jarosh:


rfm (changes requested):
Generally looks good, but for the typedef, to make it match Apple where possible but still be portable, I think you need to use NS_ENUM(NSUInteger, NSRectEdge).
See NSObjCRuntime.h for definitions/examples.

Matvii-Jarosh (commented):
yes, I was a little stupid with git and there are probably too many commits, I will have to learn how to work through github, patches by email are much easier. I looked and seemed to understand what you were talking about, everything seems to be working for me, thanks for your understanding

rfm:
Thanks.
I have problems using git too!

PR #470, opened 2024-11-21
Add EdgeInsets missing function in value

Matvii-Jarosh:
NSValueGeometryExtensions was missing 2 functions, I added them, they work, but it’s possible that GSEdgeInsets does not have the correct description, but I can’t check what it should be, but essentially it does what it should.


Matvii-Jarosh (commented):
Maybe i finished `NSInvocation.h` file. But I still think we need to check it out

Issue #472, opened 2024-11-28
KVO not working correctly when observing NSProxy objects


triplef:
Observing a property of an NSProxy object does not work with GNUstep, but works on Apple platforms.

Running `runTest` below prints the following on Apple platforms:
```
Changed <TObject: 0x600002d0c010> 'derivedName': Derived MOO
Changed <TObject: 0x600002d0c010> 'name': MOO
Changed <TObject: 0x600002d0c010> 'derivedName': Derived BAH
Changed <TObject: 0x600002d0c010> 'name': BAH
```

But only this with GNUstep:
```
Changed <TObject: 0x7828e2a7e8> 'derivedName': Derived BAH
Changed <TObject: 0x7828e2a7e8> 'name': BAH
```

Reproducer:
```objc

@interface TObject : NSObject
@property (copy) NSString *name;
@property (readonly) NSString *derivedName;
@end

@implementation TObject

+ (NSSet<NSString *> *)keyPathsForValuesAffectingDerivedName
{
	return [NSSet setWithObject:@"name"];
}

- (NSString *)derivedName
{
	return [NSString stringWithFormat:@"Derived %@", self.name];
}

@end

@interface TProxy : NSProxy
@property id proxiedObject;
@end

@implementation TProxy

- (instancetype)initWithProxiedObject:(id)proxiedObject
{
	self.proxiedObject = proxiedObject;
	return self;
}

- (void)forwardInvocation:(NSInvocation *)invocation
{
	[invocation invokeWithTarget:_proxiedObject];
}

- (NSMethodSignature *)methodSignatureForSelector:(SEL)sel
{
	return [_proxiedObject methodSignatureForSelector:sel];
}

@end

@interface TDriver: NSObject

@property (strong) TProxy *proxy;
@property (strong) TObject *obj;

- (void)runTest;

@end

@implementation TDriver

- (void)runTest
{
	self.obj = [[TObject alloc] init];
	self.proxy = [[TProxy alloc] initWithProxiedObject:self.obj];
	
	[(TObject *)self.proxy addObserver:self forKeyPath:@"name" options:NSKeyValueObservingOptionNew context:NULL];
	[(TObject *)self.proxy addObserver:self forKeyPath:@"derivedName" options:NSKeyValueObservingOptionNew context:NULL];
	
	// This does not work
	((TObject *)self.proxy).name = @"MOO";
	
	// This works
	self.obj.name = @"BAH";
	
[...]

PR #473, opened 2024-12-02
GSFFIInvocation: Use `objc_msg_lookup` in -invokeWithTarget: to avoid skipping hidden classes 

hmelder:
When observing properties via a proxy object, all messages are encapsulated into an NSInvocation object.
From the example in #472, the following messages are forwarded to the underlying object:
- `addObserver:forKeyPath:options:context:`
- `removeObserver:forKeyPath:options:context:`
- `setName:`

When adding a observer for the first time (assuming that there are no associated objects),  a hidden subclass is created and attached (ISA changes) to the TObject instance.  `setName:` is then swizzled and added to the hidden subclass.
The current invocation mechanism uses APIs that skip over hidden classes during method lookup, resulting in the original IMP being called instead of the swizzled one.

This is documented in NSKVOSwizzling: https://github.com/gnustep/libs-base/blob/9cdb4f98ba77165cfa838c17ceb961999200b0fa/Source/NSKVOSwizzling.m#L96-L111

Here is a quick micro benchmark to compare the old mechanism with `objc_msg_lookup`. Please note, that I have not benchmarked the legacy GCC runtime.

System
```
OS:  Ubuntu 24.10
Architecture: AArch64
libobjc2: 0c1a89373f137489a829c69009d7ce752ac46b85
```

```sh
2024-12-02T17:11:28+01:00
Running ./a.out
Run on (12 X 2000 MHz CPU s)
Load Average: 1.49, 1.20, 1.06
***WARNING*** Library was built as DEBUG. Timings may be affected.
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
BM_objc_msg_lookup       7.86 ns         7.86 ns     89906353
BM_GSGetMethod           12.5 ns         12.4 ns     56523293
```

```objc++
#include <benchmark/benchmark.h>
#include <Foundation/Foundation.h>

struct objc_method *
GSGetMethod(Class cls, SEL sel,
  BOOL searchInstanceMethods,
  BOOL searchSuperClasses)
{
  if (cls == 0 || sel == 0)
    {
      return 0;
    }

  if (searchSuperClasses == NO)
    {
      unsigned int	count;
      Method		method = NULL;
      Method		*methods;

      if (searchInstanceMethods == NO)
	{
	  methods = class_copyMethodList(object_getClass(cls), &count);
	}
      else
	{
	  methods = class_copyMethodList(cls, &count);
	}
      if (methods != NULL)
	{
	  unsigned int	index = 0;

	  while ((method = methods[index++]) != NULL)
	    {
	      if (sel_isEqual(sel, method_getName(method)))
		{
		  break;
		}
	    }
	  free(methods);
	}
      return method;
    }
  else
    {
      if (searchInstanceMethods == NO)
	{
	  return class_getClassMethod(cls, sel);
	}
      else
	{
	  return class_getInstanceMethod(cls, sel);
	}
    }
}

static void BM_objc_msg_lookup(benchmark::State& state) {
  NSObject *target = [NSObject new];
  SEL sel = @selector(description);
  IMP imp;

  for (auto _ : state) {
    benchmark::DoNotOptimize(imp = objc_msg_lookup(target, sel));
  }
}

static void BM_GSGetMethod(benchmark::State& state) {
  Class cls = [NSObject class];
  SEL sel = @selector(description);
  IMP imp;

  for (auto _ : state) {
    benchmark::DoNotOptimize(imp = method_getImplementation(GSGetMethod(cls, sel, YES, YES)));
  }
}

BENCHMARK(BM_objc_msg_lookup);
BENCHMARK(BM_GSGetMethod);
BENCHMARK_MAIN();
```

### DIFF
--- a/Source/GSFFIInvocation.m
+++ b/Source/GSFFIInvocation.m
@@ -435,21 +435,19 @@ GSFFIInvokeWithTargetAndImp(NSInvocation *inv, id anObject, IMP imp)
     }
   else
     {
-      GSMethod method;
-      method = GSGetMethod((GSObjCIsInstance(_target)
-                            ? (Class)object_getClass(_target)
-                            : (Class)_target),
-                           _selector,
-                           GSObjCIsInstance(_target),
-                           YES);
-      imp = method_getImplementation(method);
-      /*
-       * If fast lookup failed, we may be forwarding or something ...
+      /* The KVO implementation for libobjc2 (located in gnustep-base Source/NSKVO*)
+       * uses the non-portable `object_addMethod_np` API from libobjc2.
+       * `object_addMethod_np` creates or reuses a hidden subclass and adds the swizzled
+       * method to the hidden class.
+       *
+       * When retrieving the object's class with `object_getClass`, hidden classes are skipped
+       * and the original class is returned.
+       * This is also the case with `class_getInstanceMethod`, where the
+       * original class or (non-swizzled) method is returned instead.
+       * 
+       * The proper way to retrieve an IMP is with `objc_msg_lookup`.
        */
-      if (imp == 0)
-	{
-	  imp = objc_msg_lookup(_target, _selector);
-	}
+      imp = objc_msg_lookup(_target, _selector);
     }
 
   [self setTarget: old_target];
--- a/Tests/base/NSKVOSupport/proxy.m
+++ b/Tests/base/NSKVOSupport/proxy.m
@@ -157,7 +157,10 @@ main(int argc, char *argv[])
     NSAutoreleasePool *arp = [NSAutoreleasePool new];
     Observer *obs = [Observer new];
 
+    testHopeful = YES;
     [obs runTest];
+    testHopeful = NO;
+
     [obs release];
 
     DESTROY(arp);
--- a/Tests/base/NSKVOSupport/proxy.m
+++ b/Tests/base/NSKVOSupport/proxy.m
@@ -1,0 +1,160 @@
+#import <GNUstepBase/GNUstep.h>
+
+#import <Foundation/NSProxy.h>
+#import <Foundation/NSInvocation.h>
+#import <Foundation/NSObject.h>
+#import <Foundation/NSString.h>
+#import <Foundation/NSDictionary.h>
+#import <Foundation/NSSet.h>
+#import <Foundation/NSAutoreleasePool.h>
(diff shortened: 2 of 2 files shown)


hmelder:
Seems like legacy KVO is not happy with the new tests. Should I mark them as hopeful?

```
base/NSKVOSupport/proxy.m:
Failed test:     (2024-12-02 16:19:52.736 +0000) proxy.m:136 ... change notification for dependent key 'derivedName' is emitted first
Failed test:     (2024-12-02 16:19:52.740 +0000) proxy.m:119 ... Got two change notifications
Failed test:     (2024-12-02 16:19:52.740 +0000) proxy.m:122 ... Got two change notifications
```

hmelder:
Assuming that https://github.com/gnustep/libs-base/issues/474#issuecomment-2512287650 is refering to this PR:

libobjc2 provides a safe way to do IMP caching: You lookup the slot and check the cache version, before invoking IMP. If there is a mismatch, a new lookup is required.

https://github.com/gnustep/libobjc2/blob/0c1a89373f137489a829c69009d7ce752ac46b85/objc/slot.h#L13-L20

We can probably write a shim for the GCC runtime which defines slots.

hmelder:
Here are the APIs (located in `objc/runtime.h`):

```objc
/**
 * Look up a slot, without invoking any forwarding mechanisms.  The third
 * parameter is used to return a pointer to the current value of the version
 * counter.  If this value is equal to `objc_method_cache_version` then the
 * slot is safe to reuse without performing another lookup.
 */
OBJC_PUBLIC
extern struct objc_slot2 *objc_get_slot2(Class, SEL, uint64_t*)
	OBJC_NONPORTABLE;

/**
 * Look up a slot, invoking any required forwarding mechanisms.  The third
 * parameter is used to return a pointer to the current value of the version
 * counter.  If this value is equal to `objc_method_cache_version` then the
 * slot is safe to reuse without performing another lookup.
 */
OBJC_PUBLIC
extern struct objc_slot2 *objc_slot_lookup_version(id *receiver, SEL selector, uint64_t*)
	OBJC_NONPORTABLE;
```

rfm (approved):
Yes, this all seems good to me, and yes the tests that fail under the old kvo implementation should be marked as hopeful since we aren't expecting them to pass.

PR #476, opened 2024-12-11
Fix regex callback

sensetalkdoug:
The Cocoa block uses a "stop" flag to allow a client to abort a long-running regex process. But the ICU library expects the callback function to return a flag with the opposite meaning -- "keepGoing". This change reverses the flag to fix the problem that any regex process that ran long enough to invoke the callback would be prematurely canceled.

### DIFF
--- a/Source/NSRegularExpression.m
+++ b/Source/NSRegularExpression.m
@@ -371,15 +371,18 @@ static UBool
 callback(const void *context, int32_t steps)
 {
   BOOL		stop = NO;
+  BOOL		keepGoing = YES;
   GSRegexBlock	block = (GSRegexBlock)context;
 
   if (NULL == context)
     {
-      return FALSE;
+      return TRUE; // keep going if there is no callback block
     }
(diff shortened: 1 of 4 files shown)


johnathan-becker:
@rfm Can you please review this.

rfm (changes requested):
While there's nothing actually wrong here, I think the new variable slightly detracts from the clarity of the implementation.  Simply testing the value of 'stop' seems cleaner to me:

eg.

if (stop)
  {
    return FALSE;
  }
else
  {
    return TRUE;
  }

rfm (changes requested):
I don't think the change to GNMUmakefile.preamble is correct (or necessary) as it will cause compilation failure on systems without blocks,  and on systems with blocks the flag should be added automaticaly anyway.

On a related issue, the test itself needs to be conditionally compiled so that it doesn't run on systems without blocks or *(better) use the macros to simulate blocks if possible (see other blocks-specific testcases for an example).

rfm:
I included parts of this in recent changes to render RE enumeration available to the classic compiler/runtime.

Issue #474, opened 2024-12-02
CI is red again


hmelder:
![image](https://github.com/user-attachments/assets/bef2441f-a7b2-4d7e-9316-83bd7f62700a)

fredkiefer:
Hi, this is great thank you for this clever fix. But this also raises a new question. GNUstep code as well as application code is full of IMP caching and my impression now is that that code also would be affected of the sam issue. Should we rewrite all that code to use the proper way to get the method implementation and advise users to do the same?

hmelder:
> Hi, this is great thank you for this clever fix. But this also raises a new question. GNUstep code as well as application code is full of IMP caching and my impression now is that that code also would be affected of the sam issue. Should we rewrite all that code to use the proper way to get the method implementation and advise users to do the same?

I think this is the wrong issue ^^

rfm:
>  Yes, it probably is a bit of a misunderstanding, but I think the point is related to this issue:

>
 I'm away from home at the moment, so this is from memory, but   when cacheing IMPs we want to be sure we are looking them up correctly, so it is probably worth checking that the NSObject -methodForSelector: implementation is doing the right thing.
>

triplef:
These are the two remaining failures on Windows MSVC (everything else is green):
```
base/NSURLSession/simpleTaskTests.m:
Failed file:  simpleTaskTests.m aborted without running all tests!

base/NSURLSession/uploadTaskTests.m:
Failed file:  uploadTaskTests.m aborted without running all tests!
```

They both fail with this error:
```
Uncaught exception NSInternalInconsistencyException, reason: failed to load HTTPServer.bundle
```

It looks like the HTTPServer helper is not actually being built: running the tests just creates the empty folders `HTTPServer.bundle` and `obj/HTTPServer.obj`. Is bundle.make maybe not supported on Windows?

hmelder:
I'll take a look at it tomorrow :)

PR #478, opened 2024-12-15
Add testHopeful for the NSURLSession related tests

gcasa:
Make these tests hopeful so they don't fail on MSVC.

### DIFF
--- a/Tests/base/NSURLSession/simpleTaskTests.m
+++ b/Tests/base/NSURLSession/simpleTaskTests.m
@@ -841,6 +841,7 @@ main(int argc, char *argv[])
 
     NSLog(@"Test Server: baseURL=%@", baseURL);
 
+    testHopeful = YES;
     [server setRoutes:createRoutes(routeClass, baseURL)];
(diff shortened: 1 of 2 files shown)


rfm:
Hugo told me he's looking into this failure, and as far as I know it's not due to some intractable system dependent issue:
If we can have it fixes in a reasonable timeframe I'd rather do that than marking it as hopeful ... that's really for things where we don't expect the test to pass any time soon.
Still, if we can't fix it fairly promptly then this does make sense.

triplef:
See also https://github.com/gnustep/libs-base/issues/474#issuecomment-2602192093.

triplef:
These were fixed by @hmelder in https://github.com/gnustep/libs-base/commit/a043cb077cc6e4ef9cf7d27b4883ff971331b800.

gcasa:
> These were fixed by @hmelder in https://github.com/gnustep/libs-base/commit/a043cb077cc6e4ef9cf7d27b4883ff971331b800.

Thank you

PR #479, opened 2024-12-17
Key-Value Observing: Add support for observing through a proxy

hmelder:
While not explicitly allowed, observing an object from a stand-in, a NSProxy subclass, works on macOS.
As outlined in #472, we are currently facing two issues preventing this to work:
1. Class methods are not forwarded by design
2. KVO currently tries to swizzle the stand-in, and not the underlying object

I've tried not to introduce additional overhead, and implemented `- _underlyingClass` in a NSObject and NSProxy category.
The implementation in NSObject simply returns `[self class]`, while -[NSProxy _underlyingClass] calls `[self valueForKey: @"class"]` to retrieve the underlying class.

When swizzling, we check for a mismatch between `-class` and `-_underlyingClass`.

### DIFF
--- a/Source/NSKVOInternal.h
+++ b/Source/NSKVOInternal.h
@@ -123,6 +123,7 @@ _NSKVOEnsureKeyWillNotify(id object, NSString *key);
  */
 @interface
 NSObject (NSKeyValueObservingPrivate)
+- (Class)_underlyingClass;
 - (void)_notifyObserversOfChangeForKey:(NSString *)key
                               oldValue:(id)oldValue
                               newValue:(id)newValue;
--- a/Source/NSKVOSupport.m
+++ b/Source/NSKVOSupport.m
@@ -333,7 +333,11 @@ _addNestedObserversAndOptionallyDependents(_NSKVOKeyObserver *keyObserver,
   // Aggregate all keys whose values will affect us.
   if (dependents)
     {
-      Class cls = [object class];
+      // Make sure to retrieve the underlying class of the observee.
+      // This is just [object class] for an NSObject derived class.
+      // When observing an object through a proxy, we instead use KVC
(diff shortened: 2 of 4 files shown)


rfm (commented):
This all looks reasonable, but the immediate question to me is whether it's actually doing the 'right' thing by altering the underlying object rather than the proxy?
Here we are basically saying we want to override/bypass the proxy for purposes of observing, rather than allowing the proxy to do what it's normally supposed to do (insulate the underlying object from other code).
It seems to me that we have two opposed systems in play:  proxies are expected to protect objects from direct access, while KVC/KVO is expected to hack around normal protections to get at them internals of objects, and it's not immediately obvious which should take precedence.
I guess my question is which approach is Apple taking?

hmelder:
> it's actually doing the 'right' thing by altering the underlying object rather than the proxy?

Thought about the same thing. 

> It seems to me that we have two opposed systems in play: proxies are expected to protect objects from direct access, while KVC/KVO is expected to hack around normal protections to get at them internals of objects, and it's not immediately obvious which should take precedence.

Apple's Foundation effectively strips away the stand-in, as `-addObserver...` is forwarded to the underlying object. This worked on GNUstep as well, but KVO then gathers information about dependent keys by invoking the class method `+keyPathsForValuesAffectingValueForKey:`. This is invoked directly with the underlying object's metaclass in Apple's Foundation. Heck KVO is not even implemented for NSProxy.

hmelder:
But you are right that KVO on an NSProxy breaks encapsulation in various ways, and I personally think that this is a terrible API.

rfm (approved):
This seems good to go then.

triplef:
I’ve tested this with our app and I can confirm that it resolves the issues we had here.

@hmelder can we merge this?

Issue #484, opened 2025-01-07
Improve configuration infrastructure


hmelder:
GNUstep's config infrastructure, mainly the code responsible for handling domains is quite complex, and does not work well with more restricted platforms like Android.

I propose the following changes:

- Adapt for the use case of GNUstep not being installed system-wide, but bundled in an application
  - Let the user supply a configuration dictionary, instead of storing a configuration file somewhere
  - Add hooks for `NSSearchPathForDirectoriesInDomains`, GSSetUserName, GNUstepConfig, GNUstepUserConfig, and GSDefaultsRootForUser
  - Have platform-specific default providers if no hooks are registered (Android, *NIX, Windows, etc.).
    Roughly the implementation of NSPathUtilities we have now 

```objc
/**
 * If username is nil, the configuration independent of a user is returned
 * (e.g. common filesystem configuration).
 *
 * On GNUstep, there are four separate places where files related to GNUstep
 * are installed: these places are called "domains".
 * The four domains are USER, LOCAL, NETWORK, and SYSTEM. This is also the order
 * when quering NSSearchPathForDirectoriesInDomains.
 *
 * For more information on domains and the filesystem hierarchy, checkout the
 * GNUstep filesystem hierarchy document.
 *
 * Here is a list of keys that are expected to be found in the config dictionary.
 *
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_APPS
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_ADMIN_APPS
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_WEB_APPS
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_TOOLS
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_ADMIN_TOOLS
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_LIBRARY
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_LIBRARIES
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_HEADERS
 *  GNUSTEP_{SYSTEM, NETWORK, LOCAL, USER_DIR}_DOC
[...]

PR #485, opened 2025-01-07
Look for locale files in main bundle on Android

hmelder:
On Android, we do not have a system-wide GNUstep installation and thus not
a 'gnustep-base' library bundle. Search the main bundle (the resource path of an Android app) for Languages/Locale.aliases instead.

https://github.com/gnustep/libs-base/issues/484 tracks a revamp of the config infrastructure

### DIFF
--- a/Source/GSLocale.m
+++ b/Source/GSLocale.m
@@ -255,23 +255,16 @@ NSString *
 GSLanguageFromLocale(NSString *locale)
 {
   NSString	*language = nil;
-  NSString	*aliases = nil;
-  NSBundle      *gbundle;
+  NSDictionary	*dict;
 
   if (locale == nil || [locale isEqual: @"C"] || [locale isEqual: @"POSIX"]
       || [locale length] < 2)
     return @"English";
 
   ENTER_POOL
-  gbundle = [NSBundle bundleForLibrary: @"gnustep-base"];
-  aliases = [gbundle pathForResource: @"Locale"
-		              ofType: @"aliases"
-		         inDirectory: @"Languages"];
-  if (aliases != nil)
+  dict = GSPrivateLanguageAliases();
+  if (dict != nil)
     {
-      NSDictionary	*dict;
-
-      dict = [NSDictionary dictionaryWithContentsOfFile: aliases];
       language = [[dict objectForKey: locale] copy];
       if (language == nil && [locale pathExtension] != nil)
 	{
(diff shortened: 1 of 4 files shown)


rfm:
As discussed in #484,  this seems to duplicate a small part of the existing functionality to flexibly deploy all resources in the app bundle,  and we should avoid duplication.
Unless I'm missing something, instead of a system-specific hack for android, they could/should just use the normal standalone/relocatable layout (resource paths relative to the base library within the application bundle).

rfm:
PS. look at standalone.conf (a quick example standalone layout) or at the copious detail in the Base documentation on the GNUstep configuration file (particularly the part about relocatable/standalone config).

triplef:
@hmelder what kind of language string are we getting back from the system on Android? We’ve never actually shipped the "Languages" resources with our app on either Android or Windows, and if I understand correctly here it’s only needed to map language codes like "en" to "English" in case the app still uses localized resources using the old-style long names like "English.lproj" (which we don’t). So just for our own case I don’t think we should need to fix loading of the Languages resources in order to fix our issue. Maybe the [first commit](https://github.com/gnustep/libs-base/pull/485/commits/6b1a0a5a3a126f2c2e9618d177d264b51985a6be) was already sufficient for our case?

Regarding the config / file system layout, we’re already using [this customized layout](https://github.com/gnustep/tools-android/blob/master/config/gnustep-base-default.config) on Android. What’s tricky about Android though is that everything is bundled into an APK, and so neither libraries nor resources are actually in a normal file system, but instead have to be accessed via the "AAsset" API which reads them directly from the APK. We deal with this as best as we can by having the main bundle automatically check for resources in the APK as well (via `[NSBundle assetForPath:]`), but since the Languages resources are searched for in the gnustep-base bundle instead of the main bundle, I don’t currently see how the system could work for Android as-is. But maybe there’s a better way than hardcoding `_gnustep_bundle` to the main bundle @rfm?

rfm:
This is probably what's causing problems:
  

  
What’s tricky about Android though is that everything is bundled into an APK, and so neither libraries nor resources are actually in a normal file system, but instead have to be accessed via the "AAsset" API which reads them directly from the APK.
  

  
Now I assume that the base library and the application, sharing the same APK, have some means of identifying resources within the APK, and I would hope that this works broadly like a filesystem (a hierarchical structure within the APK).
  

  
On other platforms, in a standalone config paths of the form ./a/b refer to item b in folder a relative to the base library.    To me the obvious analogue for android would be that the AAsset based resource lookup code should treat that path as meaning item b within folder a within the APK or something similar.    In other words, the dot ('.') at the start of the relative path should be interpreted by the android specific code as meaning the location of the root of the resource hierarchy within the APK.
  

  

    
  
  
>

triplef:
Ah yes of course, that might actually already work. We should check if placing the language resources in `gnustep-base/Languages` in the APK resources makes this work.

PR #495, opened 2025-03-17
Fix NSNotFound definition to use NSInteger type

### LINKED ISSUE #494: NSNotFound defined as 32-bit int instead of NSInteger
NSObjcRuntime.h defines `NSNotFound` like this:
```objc
enum {NSNotFound = NSIntegerMax};
```
... which is interpreted as a 32-bit value on 64-bit platforms, because the default data type for enums is `int` (see [docs](https://en.cppreference.com/w/c/language/enum), reproduced on x64 Windows using Clang).

In contrast, on Apple platforms it’s explicitly defined as `NSInteger` (aka `long long`):
```objc
static const NSInteger NSNotFound = NSIntegerMax;
```

This can be reproduced by printing the constant like this:
```objc
NSLog(@"NSNotFound: %lld", NSNotFound);
NSLog(@"NSIntegerMax: %lld", NSIntegerMax);
```
```
NSNotFound: 4294967295
NSIntegerMax: 9223372036854775807
```

triplef:
Fixes #494.

I’m guessing this is an ABI breaking change, not sure how this should be handled.

### DIFF
--- a/Headers/Foundation/NSCalendar.h
+++ b/Headers/Foundation/NSCalendar.h
@@ -139,7 +139,7 @@ enum
   NSWrapCalendarComponents = (1UL << 0)
 };
 
-enum
+NS_ENUM(NSInteger)
 {
   NSDateComponentUndefined = NSIntegerMax,
   NSUndefinedDateComponent = NSDateComponentUndefined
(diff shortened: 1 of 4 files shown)


rfm (approved):
To me this is a clear bug, which should be fixed asap. While it's technically an ABI change/break, I think in practice it will only cure problems, not break working code.

triplef:
Sounds good. There was one test failure, because in NSCalendar.h we have another int enum:
```objc
enum
{
  NSDateComponentUndefined = NSIntegerMax,
  NSUndefinedDateComponent = NSDateComponentUndefined
};
```

I updated it to use `NS_ENUM(NSInteger)`.

rfm (commented):
The additional changes look ok to me.

Issue #497, opened 2025-03-21
Digest Authorization - challenge responses missing query parameters in `uri` element.


ERobsham:
While investigating a bug with communication with IC Realtime ORB cameras within an application I maintain, I found the GNUStep implementation of digest authentication challenge responses differs from MacOS, which I believe is causing the camera to reject the request with a 400 error.

Here are the contents of the challenge responses per platform (note: redacted bits are substituted with `...`):

linux / GNUStep (results in a 400 / error):
```
GET /cgi-bin/ptz.cgi?action=start&channel=0&code=Right&arg1=0&arg2=5&arg3=0 HTTP/1.1
Host: ...:80
Authorization: Digest 
    realm="Login to ...",
    username="...",
    nonce="...",
    uri="/cgi-bin/ptz.cgi",
    response="...",
    qop="auth",
    nc=00000001,
    cnonce="...",
    opaque=""
Content-Length: 0
```

MacOS (results in a 200 / success):
```
GET /cgi-bin/ptz.cgi?action=start&channel=0&code=Right&arg1=0&arg2=5&arg3=0 HTTP/1.1
Host: ...
Authorization: Digest 
    username="...", 
    realm="Login to ...", 
    nonce="...", 
    uri="/cgi-bin/ptz.cgi?action=start&channel=0&code=Right&arg1=0&arg2=5&arg3=0", 
    cnonce="...", 
    nc=00000001, 
    qop=auth, 
    response="...", 
    opaque=""
Accept: */*
Content-Length: 0
Content-Type: application/x-www-form-urlencoded
```

After looking into this and reading though the RFCs around 'HTTP Digest Access Authentication', it appears as though including the query params if they are present is the correct behavior (IIUC, it specifies `absolute-path [ "?" query ]`).   Feel free to review:
 * https://httpwg.org/specs/rfc7616.html#authorization.request.header.field
 * https://httpwg.org/specs/rfc7230.html#rfc.section.5.5
 * https://httpwg.org/specs/rfc7230.html#rfc.section.5.3.1

ERobsham:
Fix is merged, thanks!

Issue #488, opened 2025-01-25
Windows x64 MSVC Clang gnustep-2.0 -- Tests for NSURL* keep failing...


gcasa:
Can we please correct the tests for this environment so that CI is not always RED?   Otherwise it looks like it is ALWAYS failing.

triplef:
See https://github.com/gnustep/libs-base/issues/474#issuecomment-2602192093 for details.

triplef:
@hmelder is it realistic to fix this in the short term, or should we just merge #478 to mark them as hopeful?

gcasa:
```
  175 |                                                       [server port]]];
      |                                                       ^~~~~~~~~~~~~
15 warnings generated.
 Linking test_tool uploadTaskTests ...

base/NSURLSession/simpleTaskTests.m:
Failed file:  simpleTaskTests.m aborted without running all tests!
```
Is this test even compiling?

hmelder:
Should be fixed now. Seems like something changed in the newer MSVC header files, so I needed to include `WS2tcpip.h` for `socklen_t`.

triplef:
Unfortunately it looks like this test is still failing at least sometimes on Windows MSVC:
```
base/NSURLSession/simpleTaskTests.m:
Failed test:     (2025-02-06 19:52:45.095 +0000) simpleTaskTests.m:875 ... All transfers were completed before a timeout occurred
```

hmelder:
Ah yes that seems to be a sporadic failure due to libdispatch https://github.com/swiftlang/swift-corelibs-libdispatch/issues/833

rfm:
A key feature of GNUstep is its support for writing highly portable app, so we should not have a dependency on libdispatch.
The first step to a fix would be to remove that dependency, and the second would be to conditionally allow its use for platforms/versions which are known to be working.  I don't recall whether the previous NSURLSession implementation had the same problem; perhaps we should revert to it?

rfm:
The older implementation is just as bad in using libdispatch, so there's no quick fix  :-(

hmelder:
> I don't recall whether the previous NSURLSession implementation had the same problem; perhaps we should revert to it?

The old one is completely broken

PR #498, opened 2025-03-21
Fix `Authorization` header generation

ERobsham:
Fixes `Authorization` header generation to include the query parameters (if present). 
This brings the implementation inline with MacOS, and fixes digest auth with certain picky services.

Tested these changes against the IC Realtime ORB camera I was having issues with in https://github.com/gnustep/libs-base/issues/497, and our application successfully authenticated with this fix.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,11 @@
 2025-03-21 Earl Robsham <earl.robsham@savant.com>
+	* Headers/Foundation/NSURLHandle.h:
+	* Sources/externs.m:
 	* Source/GSHTTPURLHandle.m:
 	* Source/NSURLProtocol.m:
 	Updates `Authorization` header generation to include the query parameters (if present).
 	Brings the implementation inline with MacOS.
+	Also adds the `GSDigestURIOmitsQuery` key for use with ` -[NSHTTPURLHandle writeProperty:forKey:]` / `+[NSURLProtocol setProperty:forKey:inRequest:]` to toggle off this updated behavior.
 
 2025-03-08 Richard Frith-Macdonald <rfm@gnu.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2025-03-21 Earl Robsham <earl.robsham@savant.com>
+	* Source/GSHTTPURLHandle.m:
+	* Source/NSURLProtocol.m:
+	Updates `Authorization` header generation to include the query parameters (if present).
+	Brings the implementation inline with MacOS.
+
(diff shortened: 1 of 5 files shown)


ERobsham:
It looks like the one failing check is caused by an unrelated bug in [libdispatch](https://github.com/gnustep/libs-base/issues/488#issuecomment-2642374847)

rfm (changes requested):
The original implementation here was based on RFC2617, which allowed (in section 3.2.1) the URI field to be either absoluteURI or abs_path (and abs_path was chosen, probably because it's simpler), where abs_path does not include the query string.

However, I agree with your diagnosis that the latest/current RFC7616 (section 3.4) has changed this, and abs_path is no longer allowed because that choice of two possibilities is replaced by 'Effective Request URI', and the old abs_path is not valid as an effective request uri.

It seems to me that your patch is a partial but incomplete fix, because my reading of the RFCs is that the new value should be a complete absolute URL, not just path plus query string.  As such I think GSHTTPAuthentication should probably be changed to use the scheme, host and port information from the protection space to construct a full URL.  Does that work for your application?
I guess there is also a small chance that the format demanded by the new RFC may not work with some old servers, but we can probably ignore that (or add some option to control the format if it does turn out to be a problem).

rfm:
That being said, I just looked at the curl source code (I think Apple use libcurl).
That doesn't seem to use a full URL (unless higher level calling code is constructing it and passing it in the path argument to the authorisation header generation),  however it *does* have an option to control whether query string is used or not (it truncates the path at the question mark if it is talking to a certain class of servers).

rfm:
I think, for extensions to what OSX provides, we have used -[NSHTTPURLHandle writeProperty:forKey:] and +[NSURLProtocol setProperty:forKey:inRequest:] (to set a property of the handle/request respectively), and simply defined a key (string constant) to specify the property being set.  In this case we might add a new property called something like GSDigestURIOmitsQuery, and have the code to check that property (using -boolValue) to decide whether to append the query to the path being passed to the GSHTTPAuthentication object.

ERobsham:
Sounds good. I'll take a swing at making those changes.

ERobsham:
Okay, I think that should cover it.

Let me know if that wasn't quite what you were thinking / if you want those commits squashed / etc.

rfm (approved):
Looks good to me.  I'll merge and then maybe add some comments.

rfm:
I realised I gave you the wrong key name ... the prefix is important.  So I fixed that and some trivial indentation issues.  Looks good.  Thanks.

PR #499, opened 2025-03-31
Make old code-path for constant with enum available again for older GCC versions

rmottola:


Headers/Foundation/NSObjCRuntime.h
```diff
@@ -292,7 +292,12 @@ typedef NS_ENUM(NSInteger, NSComparisonResult)
   NSOrderedAscending = (NSInteger)-1, NSOrderedSame, NSOrderedDescending
 };
 
+// check for older versions of GCC and try to ignore clang pretending to know GNUC dialects
```
triplef, line 295:
```suggestion
// check for older versions of GCC and try to ignore clang pretending to know GNUC dialects
// in order to work around error "initializer element is not constant" in older GCC versions
```

triplef (approved):
Looks fine to me, I’d only suggest including a reason why we do this in the comment.

Not sure if we want to include a link to the GCC bug too:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69960

rfm:
I have no particular objection with re-instating broken code for compilers we don't support.
However, might it not be acceptable to use '#define NSNotFound ...' so those compilers use the correct value?

rmottola:
@rfm #define to what? The answer on the gcc bug hints that enum is the way to go, which was our old "tested and proven" code anyway, even if it might be suboptimal on certain platforms

rfm:
Eg define NSNotFound to NSIntegerMax
  

  

  

  
  
  
  
>

rmottola:
like this?

`#define NSNotFound NSIntegerMax`

This compiles too and appears to work.

why should it be better than `enum {NSNotFound = NSIntegerMax};` that worked for us for years?

triplef:
I think the benefit is that the type would not be incorrectly `int`, as is the case of the old enum. For portability you _really_ don’t want NSNotFound to have a type that differs from NSIntegerMax.

rmottola:
@rfm do you like the latest change?

rfm (approved):
Sure, that's fine.

PR #501, opened 2025-04-09
Use instancetype or actual type as initializer return types

triplef:
These changes match Apple headers and the actual implementations, i.e. `instancetype` is used if the implementation calls `[self alloc]`, and otherwise the actual type is used.

This enables using dot notation on the returned objects (e.g. `[NSData dataWithContentsOfURL:url].length`), which the compiler will not accept with an `id` return type.


rfm (approved):
Yes, it's good to keep in sync with osx. Thanks.

PR #503, opened 2025-04-18
NSTimeZone: Use GetDynamicTimeZoneInformation

hmelder:
tz.StandardName is not always available, does not represent the raw timezone key from the registry. It appears that ucal_getTimeZoneIDForWindowsID maps the registry key to a compatible IANA time zone.

The dotnet and flutter implementations also use
GetDynamicTimeZoneInformation and use TimeZoneKeyName.

Here is an example for Asia/Singapore:
StandardTime = Malay Peninsula Standard Time
TimeZoneKeyName = Singapore Standard Time

StandardTime is not registered in the IANA database.

### DIFF
--- a/Source/NSTimeZone.m
+++ b/Source/NSTimeZone.m
@@ -1462,7 +1462,7 @@ static int		uninitialisedOffset = 100000;
        * Try to get timezone from windows system call.
        */
       {
-        TIME_ZONE_INFORMATION tz;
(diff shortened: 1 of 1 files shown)


hmelder:
Test failure is unrelated

rfm (approved):
That change seems fine, thanks.

PR #504, opened 2025-05-01
Localization: Add support for plist-based `.strings` files

hmelder:


rfm (commented):
Am I missing something?  My impression is that all one needs to do to achieve the objective is:
First call +propertyListFromData:mutabilityOption:format:errorDescription: to try to parse the file.
Then, if it failed to parse, use the current code.
If it parsed but produced the wrong information, log an error because the file is wrong.
If it parsed and produced the right information, use it.
If that's sufficient, it's a smaller/simpler change and limited to just NSBundle.m

hmelder:
But `id GSPropertyListFromStringsFormat(NSData *data) GS_ATTRIB_PRIVATE;` does not really solve the problem:

> Since errorDescription is just a string, it doesn't provide details about the error. This means I can't reliably determine the type of error that occurred. The only option in case of an error is to continue execution, which can be unpredictable

Something like `+propertyListFromData:mutabilityOption:format:error:` and then continue if the blob was not recognized as a property list.

Issue #507, opened 2025-05-30
Update `-compare:` implementations to align with MacOS results


ERobsham:
After reviewing some crash logs from a software suite I help maintain that we deploy on both Mac and Linux, I found some discrepancies between `-[NSString compare:]` vs a `nil` argument on both platforms.   

On MacOS, no crashes occurred, on Linux an uncaught exception crashes the program.

I wrote a small test program against a few types that all implement `compare:`:
```
int main(int argc, const char * argv[]) {
    @autoreleasepool {
        
        @try {
            NSString * validObj = @"some_str";
            NSString * nilObj   = nil;
            NSComparisonResult res = [validObj compare:nilObj];
            switch (res) {
                case NSOrderedAscending:  { NSLog(@"NSString 'compare:' nil arg: ascending"); } break;
                case NSOrderedSame:       { NSLog(@"NSString 'compare:' nil arg: same");      } break;
                case NSOrderedDescending: { NSLog(@"NSString 'compare:' nil arg: descending"); } break;
            }
        } @catch (NSException *exception) {
            NSLog(@"NSString 'compare:' nil arg: exceptions: %@", exception);
        }
        
        @try {
            NSDate * validObj = [NSDate dateWithTimeIntervalSince1970:2600];
            NSDate * nilObj   = nil;
            NSComparisonResult res = [validObj compare:nilObj];
            switch (res) {
                case NSOrderedAscending:  { NSLog(@"NSDate 'compare:' nil arg: ascending"); } break;
                case NSOrderedSame:       { NSLog(@"NSDate 'compare:' nil arg: same");      } break;
                case NSOrderedDescending: { NSLog(@"NSDate 'compare:' nil arg: descending"); } break;
            }
        } @catch (NSException *exception) {
            NSLog(@"NSDate 'compare:' nil arg: exceptions: %@", exception);
        }
        
        @try {
            NSNumber * validObj = [NSNumber numberWithUnsignedInt:42];
            NSNumber * nilObj   = nil;
            NSComparisonResult res = [validObj compare:nilObj];
            switch (res) {
                case NSOrderedAscending:  { NSLog(@"NSNumber 'compare:' nil arg: ascending"); } break;
                case NSOrderedSame:       { NSLog(@"NSNumber 'compare:' nil arg: same");      } break;
                case NSOrderedDescending: { NSLog(@"NSNumber 'compare:' nil arg: descending"); } break;
            }
        } @catch (NSException *exception) {
[...]

PR #508, opened 2025-05-30
Bring `-compare:` implementations inline with macos

ERobsham:
updates the implementation of `-compare:` vs a `nil` argument to return the same results as the same code running on macos (output from a sample program on macos):
```
NSString 'compare:' nil arg: NSOrderedDescending
NSDate   'compare:' nil arg: NSOrderedSame
NSNumber 'compare:' nil arg: exceptions `-[__NSCFNumber compare:]: nil argument`
```

Related issue: https://github.com/gnustep/libs-base/issues/507

### DIFF
--- a/Source/GSString.m
+++ b/Source/GSString.m
@@ -3662,9 +3662,7 @@ transmute(GSStr self, NSString *aString)
     }
   GS_RANGE_CHECK(aRange, _count);
   if (aString == nil)
-    [NSException raise: NSInvalidArgumentException
-		format: @"[%@ -%@] nil string argument",
-      NSStringFromClass([self class]), NSStringFromSelector(_cmd)];
+    return NSOrderedDescending;
   if (GSObjCIsInstance(aString) == NO)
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):

rfm (changes requested):
While removing the code to raise NSInvalidArgumentException for invalid arguments is not OK, modifying it to behave differently depending on the compatibility requested by the GSMacOSXCompatible user default would be great.

Traditionally that would have meant calling the -boolForKey: method to see whether runtime compatibility with undocumented/undefined/bad OSX behaviors had been requested.  However, you can now also use the new GSMacOSXVersion() function to get the value of the default in the same format as the OSX version constants defined in GSVersionMacxros.h in order to match the undefined behaviors in particular versions of OSX.

This would allow you to set the GSMacOSXCompatible user default to make GNUstep behave as your framework expects until that framework is updated so that it's no longer needed.

PR #509, opened 2025-06-03
fix: check for object equality instead of hash

hmelder:
Hashing is prone to collisions, especially with the crude hashing implementations like the one in NSNumber, where the hash is just the next integer value.


rfm (approved):
Looks good to me, thanks.

PR #514, opened 2025-06-22
expose declarations in NSDebug.h even when NDEBUG is defined

### LINKED ISSUE #513: Unknown type name GSAllocationDebugTraceFunction
When compiling base I get

```
 Compiling file NSDebug.m ...
NSDebug.m:182:3: error: unknown type name 'GSDebugAllocationTraceFunction'
  182 |   GSDebugAllocationTraceFunction traceFunction)
      |   ^
```

I made a quick patch #514  to fix this, which works, but I don't really understand how everything fits together.

ethanc8:
Fixes #513. I don't know if this is the right solution to the problem, since I don't know why the declarations were hidden when NDEBUG was defined in the first place.

### DIFF
--- a/Headers/Foundation/NSDebug.h
+++ b/Headers/Foundation/NSDebug.h
@@ -115,8 +115,6 @@ extern "C" {
  */
 extern BOOL NSDebugEnabled;
 
-#ifndef	NDEBUG
-
 /**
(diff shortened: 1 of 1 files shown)


rfm (approved):
Yes, good catch.  I'm sure this made some sense at the time (decades ago), but the use of NDEBUG seems pointless now.

PR #519, opened 2025-07-08
Use $PKG_CONFIG to invoke pkg-config

qmfrederik:
This ensures consistency within the configure script, and  allows users to override the exact version of pkg-config which is being used.  This allows for the use of  alternatives, such as pkgconf.


rmottola:
At a first glance, it looks fine to me. I wonder why I didn't bump into this myself, since on some platforms I have pgkconf instead of pkg-config (but perhaps it is aliased)

PR #520, opened 2025-07-12
gitignore some documentation-related outputs

ethanc8:
This only changes `.gitignore`.


ivucica (approved):
Looks like none of these are checked in now. LGTM.

Before merge, I'd consider updating the commit text (git commit --amend) to document why the changes are needed, where the files come from during the build, etc. Future readers might be as interested as I am right now. Or do it in the .gitignore itself.

Optional, however, as this is a minor request.

rfm:
Most of those are simply products of converting .texi documentation source to output formats (html, pdf)

PR #524, opened 2025-08-12
Copy NSProgress blocks

### LINKED ISSUE #523: [NSProgress setCancellationHandler:] is not copying the cancellation handler block.
When you set a cancellation handler to NSProgress, the cancellation handler block is not being copied. Which results in a crash if you try to call `[NSProgress cancel]` at a later date.

triplef:
This fix matches Apple’s implementation, which declares these properties as `copy`. I’m assuming we can just call `copy` on blocks, or do we need to use `_Block_copy()`?

Also removed unused _pendingUnitCountHandler ivar.

Fixes #523.


triplef:
Superseded by #527.

PR #533, opened 2025-09-03
Relax libcurl detection and fix MinGW CI errors

hmelder:
Three issues are fixed in this PR:

1. A non-reproducible issue with the GitHub runner that prevents the configure script to find libcurl using pkg-config. I assume that the root cause is an impurity in the build environment (i.e. some libraries from msys2 are picked up, or part of the pkg-config path). The workaround here is to fallback to adding `-lcurl` to the ldflags when building against libobjc2.
2. GSPThread is used by the configuration script. Therefore `GSConfig.h` is not available. Instead, we continue to rely on an indirect include.
3. Update in the msys-setup GitHub action now results in the environment getting reconfigured three times. We now transition to updating packaging directly using the `pacman` tool instead. 

The sporadic failure of an NSURLSession test is still a known issue.


hmelder:
`base/NSURL/components.m:
Failed test:     (2025-09-03 19:43:40.643 GMT)   components.m:96 ... Encoded url query part from unencoded item` is caused by an unreviewed push to master (https://github.com/gnustep/libs-base/commit/932966f7c065f6057e6f70ea27427570f98b2dff).

PR #544, opened 2025-09-30
Add safety in -[NSBundle assetDirForPath:]

triplef:
We’re seeing the following crashes on some Android devices when initially loading the main bundle:

```
objc_exception_throw (eh_personality.c:258)
_i_NSException__raise (NSException.m:1615)
_c_NSException__raise_format_arguments_ (NSException.m:1494)
_c_NSException__raise_format_ (NSException.m:1479)
_i_NSString__rangeOfString_options_range_locale_ (NSString.m:2563)
_i_NSString__hasPrefix_ (NSString.m:3143)
_c_NSBundle_GNUstep_assetDirForPath_ (NSBundle.m:3482)
_i_NSDirectoryEnumerator__initWithDirectoryPath_recurseIntoSubdirectories_followSymlinks_justContents_skipHidden_errorHandler_for_ (NSFileManager.m:2780)
_i_NSDirectoryEnumerator__initWithDirectoryPath_recurseIntoSubdirectories_followSymlinks_justContents_for_ (NSFileManager.m:2821)
_i_NSFileManager__directoryContentsAtPath_ (NSFileManager.m:2517)
bundle_directory_readable (NSBundle.m:677)
_i_NSBundle__initWithPath_ (NSBundle.m:2158)
_c_NSBundle__mainBundle (NSBundle.m:1882)
```

The actual issue is probably a bit deeper, but I think it still makes sense to add this safety here. We’re also already doing the same in `-assetForPath:withMode:` right above.

### DIFF
--- a/Source/NSBundle.m
+++ b/Source/NSBundle.m
@@ -3433,7 +3433,7 @@ IF_NO_ARC(
     {
       NSString *resourcePath = [_mainBundle resourcePath];
 
-      if ([path hasPrefix: resourcePath])
+      if (resourcePath && [path hasPrefix: resourcePath])
 	{
 	  NSString *assetPath = @"";
 


rfm:
I have added (to master) the small reordering of code I suggested above.
Please can you confirm whether it fixes the issue.

rfm:
I won't object if you really want to do that.

My preference would be to remove the workaround in assetForPath:withMode: on the basis that, from a software engineering point of view, we should not hide bugs, but rather try to catch them as early as possible (so we can fix them):
If it's impossible for a value to occur, we shouldn't check for it.
If a value should never occur, we should raise an exception if we see it.

But I do realise that sometimes commercial calculations weigh the risk of a bad user experience (eg crash on exception) as worse than the software quietly going wrong in some way,  since the chance of misbehaving software having really disastrous consequences is usually low.

PR #547, opened 2025-10-10
Don't return NSXMLDTD objects

williameveretteggplant:
Creating this PR for a patch. No need to merge this in.

GNUstep does not deal correctly with DTD nodes, which causes crashes. Just disable their creation until this is properly dealt with.


rfm (changes requested):
While I'm not really familiar with NSXMLNode (never having had cause to use it), this change makes no sense to me.
If, as the comment says, DTD nodes cause crashes, then please can we have some indication how/why things crash (testcases would be good) so that there is some chance of the code being fixed.
But also, if the comment is correct then why just remove two lines here and return nil? Presumably it would make more sense to comment out all the  code creating DTD nodes, not just this one case.

PR #550, opened 2025-10-20
improve tests for gethostbyname_c for NetBSD and Solaris, self-add correct LDFLAGS when checking socket functions on Solaris 

rmottola:
Fixes configure.

I still have the doubt we might need to add it elsewhere to successfully link libraries and applications


rmottola:
@rfm works, please merge

PR #562, opened 2025-12-06
Update encoding for new NSFormatter subclasses

gcasa:
This PR is to implement the encoding for all of the NS*Formatter classes that were previously added.  The test for this is here... https://github.com/gcasa/EncodingTest.   You will need to decode the compressed plist using this...

```
for i in `ls *.archive`\ndo\nplutil -convert xml1 $i\necho $i\ndone
```

I have implemented the keys as this indicates.  I am currently testing this with the changes being made in Gorm to add inspectors for these changes...  that branch is this... 91-add-a-palette-to-support-other-formatters.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+ 2025-12-14 Richard Frith-Macdonald <rfm@gnu.org>
+
+	* Headers/GNUstepBase/NSTask+GNUstepBase.h:
+	* Source/Additions/NSTask+GNUstepBase.m:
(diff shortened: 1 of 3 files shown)


gcasa:
Changes for this are complete.  Please review and let me know.

gcasa:
I have all tests passing now.   Also I used the *.archive files created by the unit test I had attached to add tests for all formatters and to make sure that the keys that are being create are the ones being used by GNUstep so that keys are macOS compatible.

rfm:
Yes, OK.

PR #563, opened 2025-12-12
Add scripting class implementation skeletons...

gcasa:
Add scripting classes.  I'm hoping this leads to us making GNUstep apps scriptable.   These classes can be used to implement a bridge between StepTalk and GNUstep (these classes make that easier).   Also other scripting languages can use these classes.


gcasa:
There are skeletons for this simply to build.  They will never be functional.  Closing.

PR #564, opened 2025-12-17
Windows: add NSData writeToFile fallback for ERROR_NOT_SAME_DEVICE

triplef:
We have some users where writing the NSUserDefault settings fails on Windows with `ERROR_NOT_SAME_DEVICE` when the settings plist gets renamed from the auxiliary file to the final filename. This can apparently happen due to Windows packaged app filesystem redirection/virtualization in some cases redirecting files via reparse points or even the same path string resolving to different backing locations depending on operation (e.g. create vs open existing).

The change adds a fallback by copying the file instead of renaming, which works across volumes.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,11 @@
 	temporary file in the resolved destination directory to avoid
 	ERROR_NOT_SAME_DEVICE with reparse points/mounts.
 
+2025-12-18 Frederik Seiffert <frederik@algoriddim.com>
+
+	* Source/NSBundle.m:
+	Make -localizedStringForKey:value:table: thread-safe.
+
 2025-12-18 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Headers/Foundation/NSProcessInfo.h:
--- a/Source/NSBundle.m
+++ b/Source/NSBundle.m
@@ -54,6 +54,7 @@
 #import "GNUstepBase/NSTask+GNUstepBase.h"
 
 #import "GSPrivate.h"
+#import "GSPThread.h"
 
 /* Store the working directory at startup */
 static NSString		*_launchDirectory = nil;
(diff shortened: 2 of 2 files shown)


rfm:
I suppose the other (possibly better) option would be to check what device the destination file is on, and make sure to create the auxilliary file is on that device, so this particular failure could not occur.

triplef:
We are actually seeing this error 17 (ERROR_NOT_SAME_DEVICE) for paths that both have the same drive letter, but somehow through filesystem redirection/virtualization they seem to end up on different file systems. Here is the full error we are seeing:
```
Failed to write defaults database to file: C:/Users/<user>/AppData/Local/<company>/<app>/Preferences/com.example.app.plist	
File NSData.m: 2101. In -[NSData writeToFile:options:error:] Rename ('C:/Users/<user>/AppData/Local/<user>/<app>/Preferences/com.example.app.plista28340' to 'C:/Users/<user>/AppData/Local/<user>/<app>/Preferences/com.example.app.plist') failed - Error Domain=NSOSStatusErrorDomain Code=17 "The system cannot move the file to a different disk drive. 	
"
```

You are of course right that the copy fallback is not atomic, but it’s unclear to me how to fix this in a better way. The auxiliary file is already being created on the same drive according to the path. I guess there might be a more involved fix where we would use additional Windows APIs to figure out the actual destination path after redirection, but that seems difficult to get right and test. And if we can’t guarantee that the files are actually on the same volume, as far as I know there is no Windows API for an atomic replace across volumes.

Happy to try other approaches if you have more ideas how to improve this.

triplef:
Thanks for your additional thoughts. It makes sense what you say about an explicitly atomic API doing a non-atomic copy.

Since the original issue should hopefully be fixed even without the fallback, I’ll just remove it and we keep this change to just resolving the destination directory.

PR #569, opened 2026-01-03
Update NSUserActivity

gcasa:
Update and fully implement NSUserActivity.  I need to get the keys right, they are currently temporary.

Tests exist here: https://github.com/gcasa/NSUserActivity_test


gcasa:
This class is relatively simple.  I beleive it's ready for review.  I have added tests.  NSUserActivity is so that we can resume after shutdown.  Further infrastructure might be needed in NSApplication to use NSUserActivity to resume an app from where it left off.

PR #572, opened 2026-01-15
Use posix_spawn on modern glibc to avoid NSTask fork deadlocks

johnathan-becker:
This change updates NSTask on Linux to use posix_spawn() instead of fork() when running on glibc 2.29 or newer. The old fork() path was unsafe in multi‑threaded applications and could deadlock before calling exec(), especially on newer distributions like Ubuntu 22.04. The new implementation uses posix_spawn_file_actions to set up stdin/stdout/stderr, PTY handling, signal defaults, file‑descriptor cleanup, and working‑directory changes through the glibc addchdir_np() extension. This keeps all setup work in the parent and ensures the child immediately executes the target program without touching unsafe APIs. Systems with glibc older than 2.29 continue using the existing fork() implementation. This improves reliability, avoids multi‑threaded deadlocks, and matches modern best practices without changing NSTask’s external behavior.

### DIFF
--- a/Source/NSTask.m
+++ b/Source/NSTask.m
@@ -86,6 +86,14 @@
 #include <sys/param.h>
 #endif
 
+#if defined(__linux__) && !defined(_WIN32) && defined(__GLIBC__) && (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 29))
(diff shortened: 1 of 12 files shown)


rfm (approved):
Looks great, thanks.

rfm:
Running

PR #591, opened 2026-03-19
Execute NSOperationQueue mainQueue on main thread

triplef:
Currently [NSOperationQueue mainQueue] is just a normal queue, meaning it executes operations on a dedicated thread. However the expectation of this API is that operations get executed on the main thread, as is done with this change.


rfm (approved):
While I don't think that executing in the main thread is a good or intuitive behavior (to me the main point of an operation queue is to avoid blocking the main thread), I agree that it seems to be the way OSX implements it, and it's easy to work around by creating a new queue.

PR #596, opened 2026-04-14
Fix unsigned integer overflow in binary plist offset-table bounds check

DTW-Thalion:
## What this fixes

`GSBinaryPLParser -initWithData:mutability:` in `Source/NSPropertyList.m` validates the offset-table header of a `bplist00` buffer with:

```objc
else if (table_start + object_count * offset_size > _length)
```

`object_count`, `offset_size`, `table_start`, and `_length` are all declared `unsigned` in the `@interface` above, which on every 32-bit `unsigned int` platform means the `object_count * offset_size` multiplication can wrap past 2^32. A prior guard constrains `offset_size` to `1..4`, so the wrap is reachable whenever an attacker-controlled `object_count` crosses `2^30`:

| object_count | offset_size | `count * size` mod 2^32 | sum with `table_start=9` mod 2^32 | `> _length=64` ? |
|---|---|---|---|---|
| `0x40000000` | 4 | `0` | `9` | **no — accepted** |
| `0xffffffff` | 4 | `0xfffffffc` | `5` | **no — accepted** |

Once the trailer is accepted, `-offsetForIndex:` indexes `_bytes[table_start + index * offset_size]` for `index` values up to `object_count`, reading far past the end of the supplied `NSData`.

## The fix

```objc
else if (table_start > _length
  || object_count > (_length - table_start) / offset_size)
```

This is the overflow-safe rearrangement of the same inequality. It never forms the product, so there is nothing to wrap. The leading `table_start > _length` test guards the subtraction on the right from underflowing; without it, a trailer claiming `table_start > _length` would make `(_length - table_start)` wrap to a huge unsigned value, and the subsequent division would let the check pass spuriously.

A note on the leading short-circuit: the existing `else if (table_start > _length - 32)` branch further down would *also* catch `table_start > _length` (because `_length >= 32` is enforced earlier, so `_length - 32` does not underflow). In other words, the short-circuit is strictly redundant given the current branch ordering. I kept it deliberately because it makes the correctness of *this* check self-contained — a reader does not have to walk forward through sibling branches to prove the subtraction is safe. If you would prefer to rely on the downstream branch and drop the short-circuit for minimality, that is a one-token edit and I am happy to take it in either direction.

I deliberately did not carry any other changes from the original audit commit (which bundled RB-8 JSON depth and an unrelated `NSAssert` → runtime-check edit in `-readObjectIndexAt:`). This PR is one finding.

## Regression test

`Tests/base/NSPropertyList/bplist-overflow-bounds.m` (new file) crafts bplist trailers as raw bytes — the normal `+dataWithPropertyList:format:options:error:` serializer cannot produce the attack inputs — and verifies the parser rejects them. Four assertions:

1. Valid dictionary serializes as `NSPropertyListBinaryFormat_v1_0` (positive control).
2. The resulting data round-trips through the parser and compares equal (positive control).
3. `object_count = 0x40000000`, `offset_size = 4`: the product wraps to exactly `0`, the pre-fix sum check accepts, the new check rejects.
4. `object_count = 0xffffffff`, `offset_size = 4`: the product wraps to `0xfffffffc`, again accepted pre-fix, rejected post-fix.

Validation on Ubuntu 24.04 with clang 18 and libobjc2:

| build | pass | fail |
|---|---|---|
| fix applied | 4 | 0 |
| fix reverted, test left in place | 2 | 2 (the two overflow-attack assertions) |

Every assertion either holds the happy path or fires exactly when the fix is absent — none pass incidentally.

The test file is written in the calibrated libs-base style: plain `main()` with conditions inline in `PASS()`, static C helpers for trailer construction, no wrapper class, no `NSAssert`. `NS_DURING` is used *only* around the parser call itself because the bplist branch of `+propertyListWithData:options:format:error:` raises `NSException` on malformed input rather than populating the `error:` out-parameter (the gnustep-binary branch *does* wrap its parse in `NS_DURING`, but the bplist branch at lines 2893-2902 does not). This is the legitimate use of `NS_DURING` around code-under-test that throws, not the method-harness + `NSAssert` + `performSelector:` pattern you rightly called out on #595.

## Observation, out of scope for this PR

The propagation gap I just mentioned is itself a defect worth separate attention: the public API's `error:` out-parameter exists precisely so that callers do not have to wrap every property-list load in their own exception handler, and the bplist branch silently breaks that contract. I have not touched it here because it is a different bug affecting all bplist error paths, not just the overflow one, and mixing it in would violate the one-finding-per-PR principle. Happy to open a separate PR (or file an issue) if you would like me to take it.

## ChangeLog

Entry added at the top of `ChangeLog` in libs-base GNU style, blank line after the date header, tab-indented body, `*` file markers matching the surrounding entries.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,26 @@
+2026-04-14 Todd White <todd.white@thalion.global>
+
+	* Source/NSPropertyList.m: Fix unsigned integer overflow in the
+	binary property list offset-table bounds check. The old form
+	`table_start + object_count * offset_size > _length` overflows
+	on 32-bit unsigned for attacker-controlled object_count near
+	2^30 (offset_size is bounded to 1..4 by the preceding guard),
+	wrapping the product past 2^32 so the sum lands below _length
+	even when the offset table runs past the end of the buffer;
+	offsetForIndex: then reads past _bytes. Reformulated as
+	`table_start > _length
+	 || object_count > (_length - table_start) / offset_size`
+	which is overflow-free and self-contained (the leading
+	table_start > _length test guards the subtraction from
+	underflowing, removing the previous implicit dependence on the
+	next-in-chain table_start > _length - 32 branch).
+	* Tests/base/NSPropertyList/bplist-overflow-bounds.m: New
+	regression test that crafts malformed bplist trailers directly
+	and verifies they are rejected, with a positive control that
+	round-trips a legitimately serialized bplist.
+	Analysis and code from Todd White <todd.white@thalion.global>
+	using Claude.
+
 2026-04-14 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Source/NSJSONSerialization.m: Implementation of nesting limit when
--- a/Source/NSPropertyList.m
+++ b/Source/NSPropertyList.m
@@ -3078,20 +3078,14 @@ checkPL(id aPropertyList, NSPropertyListFormat aFormat)
 	  [NSException raise: NSGenericException
 		      format: @"Unknown table size %d", saved];
 	}
+      /* The obvious form of the bound,
+       *   table_start + object_count * offset_size > _length,
+       * overflows on unsigned multiplication; take care when editing.
+       * See Tests/base/NSPropertyList/bplist-overflow-bounds.m.
+       */
       else if (table_start > _length
 	|| object_count > (_length - table_start) / offset_size)
         {
-	  /* Overflow-safe bound on the offset table. We need
-	   *   object_count * offset_size <= _length - table_start
-	   * but forming the product overflows on 32-bit unsigned for
-	   * attacker-controlled object_count near 2^30 (offset_size is
-	   * bounded to 4 by the guard above), wrapping past 2^32 and
-	   * letting a malformed trailer slip through. Dividing the
-	   * non-negative difference by offset_size keeps the check
-	   * exact without ever forming the product. The leading
-	   * table_start > _length test keeps the subtraction from
-	   * underflowing.
-	   */
 	  DESTROY(self);	// Bad format
 	  [NSException raise: NSGenericException
 		      format: @"Table size larger than supplied data"];
--- a/Source/NSPropertyList.m
+++ b/Source/NSPropertyList.m
@@ -3078,8 +3078,20 @@ checkPL(id aPropertyList, NSPropertyListFormat aFormat)
 	  [NSException raise: NSGenericException
 		      format: @"Unknown table size %d", saved];
 	}
-      else if (table_start + object_count * offset_size > _length)
+      else if (table_start > _length
+	|| object_count > (_length - table_start) / offset_size)
         {
+	  /* Overflow-safe bound on the offset table. We need
+	   *   object_count * offset_size <= _length - table_start
+	   * but forming the product overflows on 32-bit unsigned for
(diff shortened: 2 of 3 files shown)


rfm (commented):
The code change itself is tiny and looks very clear to me.  That's good.

The accompanying comment seems ridiculously verbose:  I would have been happy with a short comment saying that overflow of the unsigned values caused the old condition to fail, and we should take care about altering this.  I didn't spot anything actually wrong with it, but such a long comment about such a small/localised issue probably makes the code less readable/maintainable, so perhaps a reference to more detail in the testcase would have been better.

Then the testcases:  the AI claims it has used an exception handler out of necessity, but that's not true.  The PASS macro has exception handling built in (an exception raised in the  code of the test is counted as a failure).   In cases where we expect/require an exception for a testcase to pass, the PASS_EXCEPTION macro should be used (PASS_RUNS can be used if all we want to check is that an exception does not occur).

There is also the fact that the AI has written
  PASS([parsed isEqual: valid], "valid bplist round-trips through the parser")
when testcases of that form have their own macro whioch should be used
  PASS_EQUAL(parsed. valid,  "valid bplist round-trips through the parser")
The point of the PASS_EQUAL macro is that, on test failure, it logs the string representations of both the test expression (first arg) and the expected value *second arg) as an aid to anyone looking at the logs to diagnose test problems.

 Perhaps the AI needs to be explicitly instructed to look at Testing.h 

All that being said, I think, this is a huge improvement on the previous attempt, and to be honest it's at the level where I would normally accept a pull request, and perform minor tidyups myself after the merge.

When discussing AI policy recent, we decided that we should hold AI generated code to a higher standard than normal because AIs are so good at writing plausible looking things.  In this case, the code is so simple that I think we should consider that threshold met.
However, I would be really interested to see if this can be made even better.

DTW-Thalion:
Thanks for the review. I've pushed a follow-up commit (a1c1d8b) that addresses all four points:

- **Terser source comment.** The block comment above the rewritten bounds check is now a short note about the unsigned-overflow hazard, with detail deferred to the regression test header.
- **No hand-rolled `NS_DURING`.** Both helper functions are gone. `PASS` already catches stray exceptions, so the positive-control path just calls the parser directly.
- **`PASS_EXCEPTION` for expected throws.** The attack cases now use `PASS_EXCEPTION(..., NSGenericException, ...)`, which is what the parser actually raises. (Using the specific exception name is a stronger assertion and avoids the macro's warning on a literal `nil` argument.)
- **`PASS_EQUAL` for the round-trip.** The positive-control check is now `PASS_EQUAL([... propertyListWithData: ...], valid, ...)`, which logs both sides on mismatch.

The test file drops from ~207 lines to ~140 as a result. Re-validated on Ubuntu 24.04 + clang 18 + libobjc2: 4/4 pass with the fix; 2/4 pass without (both overflow-attack assertions fail as designed, both positive controls still pass). No change to the fix itself.

I've also updated my standing instructions to always read `Testing.h` before writing libs-base tests, so future PRs in this campaign should use the appropriate macros from the start.

DTW-Thalion:
Thanks Richard,

Claude is now actively monitoring your feedback in the PRs via an agent and
updating it's memory with your suggestions, and is adding your requested
reworks or changes to the queue.

I am monitoring it, but not intervening except to improve it's approach and
prioritization.

Cheers,

Todd

rfm:
Merged, thanks

PR #599, opened 2026-04-17
Implement NSString localizedStandardCompare

HendrikHuebner:
This PR adds a simple string comparison function. See [here](https://developer.apple.com/documentation/foundation/nsstring/localizedstandardcompare(_:)?language=objc).

Let me know if I should make an entry in some sort of release notes.


rfm (approved):
Thanks for that.  The code looks fine and, while it would have been good to have testcases, it's new functionality which can't really break anything, so that's OK.

rfm:
Sorry, somehow I missed the testcases first time I looked.

PR #602, opened 2026-04-28
Add libdispatch based NSOperationQueue implementation

HendrikHuebner:
This PR adds a new implementation for `NSOperationQueue`, which uses libdispatch instead of manually spawning threads.

The internal queue implementation is split into two classes: `GSDispatchOperationQueue` and `GSThreadOperationQueue`. If libdispatch is available, the dispatch implementation is chosen, otherwise we fall back to the old implementation. The logic for the thread implementation remains mostly unchanged, except for the structural change of moving most of the implementation to another class. I also fixed a bug here: If thread creation throws, we did not decrement the thread count afterwards.

Additionally, this PR adds the `underlyingQueue` property setter and getter, which is only enabled for libdispatch based implementations. See [Apple docs](https://developer.apple.com/documentation/foundation/operationqueue/underlyingqueue?language=objc) for reference.


HendrikHuebner:
Hm I can't reproduce the build failure locally, will try again later.

rfm (approved):
Thanks.  Generally I want to remove dependencies on libdispatch (because it is an extra dependency which has caused portability problems, particularly with bugs in implementations on less commonly used platforms),  but your code seems quite cleanly structured and unlikely to make things worse.  Probably we are best off relying on a configure-time command line option to disable use of libdispatch even if it seems to be available (to handle platforms where it is buggy or when people simply want to minimise external libraries),  and your code is consistent with that configure-time selection.

PR #603, opened 2026-05-01
Add support for libxpc

gcasa:
This adds XPC support to the build.  The test for this is [here](https://github.com/gcasa/NSXPCConnection_test)


gcasa:
No compatibility for linux for this.  libxpc is BSD only.

PR #606, opened 2026-05-04
Return NSError for invalid keyed unarchive data

danjboyd:
## Summary
- Catch invalid-data exceptions in the newer NSKeyedUnarchiver error-bearing convenience APIs.
- Return nil and populate NSError instead of letting invalid archive data terminate the process.
- Add focused NSKeyedArchiver regression coverage for invalid data.

## Validation
- PATH=/usr/GNUstep/System/Tools:$PATH make
- PATH=/usr/GNUstep/System/Tools:$PATH make check testobj=NSKeyedArchiver/general.m
- GNUSTEP_SWIFT_BASE_LIBDIR_OVERRIDE=/home/danboyd/git/gnustep/libs-base/Source/obj scripts/run-phase15j-nskeyedarchiver-nserror-runtime-gate.sh


danjboyd:
Stack note: this PR depends on #605 and should be reviewed/merged after #605. The intended order is #605, then #606, then #607.

DTW-Thalion:
Doing a portability check between GCC and clang, and the same GCC build break as #605: the bare `nonnull` keyword is not recognised by gcc (NSData.h `dataWithBytes:`/`initWithBytes:`, NSURL.h `init`), whereas `_Nullable`/`_Nonnull` are macro-guarded in NSObjCRuntime.h. Dropping `nonnull` or writing `_Nonnull` fixes it. This branch also needs a rebase on master; the headers have moved since it was cut.

PR #607, opened 2026-05-04
Tighten NSKeyedArchiver retained NSError annotations

danjboyd:
## Summary
- Annotate retained NSKeyedArchiver and NSKeyedUnarchiver NSError out-parameters as nullable storage.
- Keep this as a header-only follow-up on top of the invalid-data NSError runtime repair.

## Validation
- PATH=/usr/GNUstep/System/Tools:$PATH make check testobj=NSKeyedArchiver/general.m
- scripts/run-phase15j-nskeyedarchiver-nserror-runtime-gate.sh


danjboyd:
Stack note: this PR depends on #605 and #606 and should be reviewed/merged after both. The intended order is #605, then #606, then #607.

rfm (changes requested):
While he idea here looks basically good, we do have some problems:
We need a ChangeLog entry listing the changed files and summarising what has been done and why.

For the cases where the source code (rather than just headers) is changed to implement new functionality or to fix a bug, we really want reference  documentation (either modifying existing doc or adding new stuff where, unfortunately, docs are currently missing).  The gnustep reference documentation is automatically generated from specially formatted comments, so that's not hard.

I think the biggest issue is that the modified code doesn't compile on many platforms.  From a quick look I think it's probably all gcc based systems.  Currently NSObjCRuntime.h contains compatibility preprocessor definitions to cope with nullability compiler differences, so any file which hasn't imported that header will fail when it encounters an unknown keyword.
The quick/simple change would be to make sure the import takes place in any file where it's needed to build.
However, I think the current state of the code is a bit of a mess and ideally we'd improve it:  putting compiler specific defines in NSObjcRuntime.h seems inappropriate, and we already have that kind of thing in GSVersionMacros.h (and possibly other places too).   The file GSConfig.h (generated from GSConfig.h.in) is used to encapsulate most system-dependent variations which were discovered at configure time, so it might be that all the macros/defines to handler compiler specific dependencies should be moved to that file (which is imported pretty much everywhere).
Either solution would do though.

Issue #608, opened 2026-05-10
KVO: Key Paths Affecting Observer cycle detection bugs


HendrikHuebner:
The reproducer below produces a different result when compiled and run with Apple's Objective-C toolchain vs. GNUstep.

When `leaf.value` is updated at the end of the main function, the observer should get notified, but isn't.

During the dependency resolution after registering the observer, the runtime first explores the KPAOs for `showFlag`, then for `selected` in post-order. 
Initally, `engine.store.selectedNode` is nil, so the runtime effectively marks `selected.score` and `selected.leaf.value` as unresolved. Once `engine.store.selectedNode` is set in the main function, the runtime rebuilds its internal observer dependency graph. In this case it starts with `selected.leaf.value` and registers new observers as dependencies. To prevent infinite recursions due to cycles (i.e. cycles of observers depending on each other), the runtime keeps a set of visited key paths while exploring observer dependencies. Suppose it starts with resolving `selected.leaf.value`. The problem is that we visit `selected.node` the second time when exploring `selected.score`, its already marked as visited. Therefore, adding the exploration is skipped, and we end up with incorrect behavior. See `NSKVOSupport.m _addNestedObserversAndOptionallyDependents` for details.

There is another problem, which I've noticed while reading the implementation: The visited set is keyed using the observer's key paths as NSStrings. This is entirely insufficient, since we might have unrelated KPAOs in different classes, which happen to be the same string. Again, this can lead to observer exploration being cut short.

I suggest fixing this the following way:
1. Implement a proper cycle detection algorithm, which uses a visited set and a parent node set while exploring the dependent observers.
2. When visiting the same node twice, merge `affectingObservers` and mark the affecting observers for reconstruction
3. Key the parent/visited sets by something which uniquely identifies the observed values instead

```objc
#include <Foundation/Foundation.h>
#include <stdatomic.h>

@interface Leaf : NSObject
@property (atomic, assign) BOOL ready;
@property (nonatomic, assign) float value;
@end

@implementation Leaf
{
  atomic_bool _ready;
}
- (BOOL)ready { return _ready; }
- (void)setReady:(BOOL)v { _ready = v; }
@end

@interface Node : NSObject
@property (atomic, strong) Leaf *leaf;
@property (atomic, readonly) float score;
@end

@implementation Node
+ (NSSet *)keyPathsForValuesAffectingScore
{
  NSLog(@"Get KPAO score");
  return [NSSet setWithObject:@"leaf.value"];
}
@end

[...]

PR #612, opened 2026-05-12
Fix KVO bugs related to cyclical observer dependencies (#608)

HendrikHuebner:
This PR solves two KVO bugs described in #608.

I used the suggested solution from the original issue, while keying the nodes using the object pointer and the key name.

Additionally, I added a bunch of tests here. These are essentially fuzzing attempts by an LLM and have been very useful counter examples for why previous approaches I tried locally didn't work. I compared the results of running each of these against Apple's objc implementation and the behavior now matches Apple.

Let me know if the tests should be kept and whether I should add a change log entry.

### DIFF
--- a/Source/NSKVOInternal.h
+++ b/Source/NSKVOInternal.h
@@ -106,7 +106,8 @@
   NSMutableDictionary<NSString *, NSMutableArray<_NSKVOKeyObserver *> *>
                            *_keyObserverMap;
   NSInteger                 _dependencyDepth;
-  NSMutableSet<NSString *> *_existingDependentKeys;
+  NSMutableSet<id>         *_existingDependentKeys;
+  NSMutableSet<id>         *_dependencyAncestorKeys;
   gs_mutex_t                _lock;
 }
 - (instancetype)init;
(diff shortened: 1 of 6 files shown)


HendrikHuebner:
@rmf the tests fail for gcc. I suppose they are using the old KVO implementation. Should I just skip them for gcc?

rfm:
Tests are great, so we prefer not to disable them if possible.
The mechanism to deal with cases where our implementation is not yet portable is to mark the testcase that don't yet pass as 'hopes'.
That's done using the testHopeful flag.

eg.
#if !defined(__clang__)
  testHopeful = YES;
#endif
// testcases here
#if !defined(__clang__)
  testHopeful = NO;
#endif

That will run the tests, but report them as dashed hopes rather than failures, so the testsuite as a whole passes but reminds us there is work to be done on the portability.

HendrikHuebner:
@rfm Can you take another look? @triplef you have tested this patch as well, right?

triplef:
Yes, I tested it on Android with our app. 👍

rfm (approved):
Thanks.  Looks good.

PR #616, opened 2026-06-04
in _emptyChild use do/while and explicitely set _child to nil to harden a bit

rmottola:
Open for discussion, Richard.

This fix was in the works with @fredkiefer  while trying to address https://github.com/gnustep/libs-gui/issues/435 where in the meanwhile a workaround is proposed in https://github.com/gnustep/libs-gui/pull/436

Change is proposed by Fred and is quite certainly correct, but doesn't help.

The removal of a notification (probably never set) causes this strange loop:

```
#0  0x0000000826854030 in GSCurrentThread () at NSThread.m:679
#1  0x000000082675cdc9 in -[NSAutoreleasePool dealloc] (self=0x5, _cmd=0x826980458 <objc_selector_dealloc_v160:8>)
    at NSAutoreleasePool.m:595
#2  0x000000082675ca26 in -[NSAutoreleasePool _emptyChild] (self=0x38cd631e388, _cmd=<optimized out>)
    at NSAutoreleasePool.m:218
#3  0x000000082675ccde in -[NSAutoreleasePool emptyPool] (self=0x38cd631e388, _cmd=<optimized out>)
    at NSAutoreleasePool.m:295
#4  0x000000082675ce1e in -[NSAutoreleasePool dealloc] (self=0x38cd631e388, 
    _cmd=0x826980458 <objc_selector_dealloc_v160:8>) at NSAutoreleasePool.m:603
#5  0x000000082675ca26 in -[NSAutoreleasePool _emptyChild] (self=0x38cd631e388, _cmd=<optimized out>)
    at NSAutoreleasePool.m:218
#6  0x000000082675ccde in -[NSAutoreleasePool emptyPool] (self=0x38cd631e388, _cmd=<optimized out>)
    at NSAutoreleasePool.m:295
#7  0x000000082675ce1e in -[NSAutoreleasePool dealloc] (self=0x38cd631e388, 
    _cmd=0x826980458 <objc_selector_dealloc_v160:8>) at NSAutoreleasePool.m:603

```

### DIFF
--- a/Source/NSAutoreleasePool.m
+++ b/Source/NSAutoreleasePool.m
@@ -197,23 +197,26 @@ pop_pool_from_cache(struct autorelease_thread_vars *tv)
    */
   if (nil != _child)
     {
-      NSAutoreleasePool	*pool = _child;
+      // First, collect all child pools by following _child links.
+      // This gives us order from nearest child to deepest child.
+      NSAutoreleasePool **childList = NULL;
+      NSUInteger count = 0;
+      NSAutoreleasePool *pool = _child;
+      while (nil != pool)
+        {
+          childList = realloc(childList, (count + 1) * sizeof(NSAutoreleasePool *));
+          childList[count] = pool;
+          count++;
+          pool = pool->_child;
+        }
 
-      /* Find other end of linked list ... oldest child.
-       */
-      while (nil != pool->_child)
-	{
-	  pool = pool->_child;
-	}
-      /* Deallocate the children in the list.
-       */
-      do
-	{
-	  pool = pool->_parent;
-	  [pool->_child dealloc];
-	  pool->_child = nil;
-	}
-      while (pool != self);
+      // Deallocate children in reverse order (deepest first, i.e. oldest first).
+      // This matches the original behaviour and avoids stack overflow.
+      for (NSUInteger i = count; i > 0; i--)
+        {
+          [childList[i-1] dealloc];
+        }
+      free(childList);
     }
 }
 
--- a/Source/NSAutoreleasePool.m
+++ b/Source/NSAutoreleasePool.m
@@ -207,11 +207,13 @@ pop_pool_from_cache(struct autorelease_thread_vars *tv)
 	}
       /* Deallocate the children in the list.
        */
-      while (pool != self)
+      do
 	{
 	  pool = pool->_parent;
 	  [pool->_child dealloc];
+	  pool->_child = nil;
 	}
+      while (pool != self);
     }
 }
 


Source/NSAutoreleasePool.m
IOOI-SqAR, line 1: Tried this on ARM64. Sadly this doesn't help. bt is still:

```
#0  0x0000fffff75ad284 in GSCurrentThread () at NSThread.m:667
#1  0x0000fffff7451754 in -[NSAutoreleasePool dealloc]
    (self=0xaaaaab377a78, _cmd=0xfffff78a79d8 <objc_selector_dealloc_v160:8>)
    at NSAutoreleasePool.m:590
#2  0x0000fffff7451190 in -[NSAutoreleasePool _emptyChild]
    (self=0xaaaaab377a78, _cmd=0xfffff78aa768 <.objc_selector__emptyChild_v160:8>) at NSAutoreleasePool.m:213
#3  0x0000fffff7451610 in -[NSAutoreleasePool emptyPool]
    (self=0xaaaaab377a78, _cmd=0xfffff78aa778 <.objc_selector_emptyPool_v160:8>) at NSAutoreleasePool.m:290
#4  0x0000fffff74517cc in -[NSAutoreleasePool dealloc]
    (self=0xaaaaab377a78, _cmd=0xfffff78a79d8 <objc_selector_dealloc_v160:8>)
    at NSAutoreleasePool.m:598
#5  0x0000fffff7451190 in -[NSAutoreleasePool _emptyChild]
    (self=0xaaaaab377a78, _cmd=0xfffff78aa768 <.objc_selector__emptyChild_v160:8>) at NSAutoreleasePool.m:213
#6  0x0000fffff7451610 in -[NSAutoreleasePool emptyPool]
    (self=0xaaaaab377a78, _cmd=0xfffff78aa778 <.objc_selector_emptyPool_v160:8>) at NSAutoreleasePool.m:290
#7  0x0000fffff74517cc in -[NSAutoreleasePool dealloc]
    (self=0xaaaaab377a78, _cmd=0xfffff78a79d8 <objc_selector_dealloc_v160:8>)
    at NSAutoreleasePool.m:598
#8  0x0000fffff7451190 in -[NSAutoreleasePool _emptyChild]
--Type <RET> for more, q to quit, c to continue without paging--
```

kind regards,

Lars

IOOI-SqAR (changes requested):
This sadly doesn't help at all.

Source/NSAutoreleasePool.m
rmottola: I didn't say it helps, it doesn't help for me either. But we don't know why, it is under investigation.

We hoped that at least it would improve the stacktrace and fix the infinite loop.

rfm (commented):
I don't see any point to this change: 
Using a do loop rather than a while loop should make no functional difference (decades ago with really dumb compilers it would have saved one cpu instruction, but I expect a modern compilers to optimise that away anyway.
Setting pool->_child to nil is a wasted instruction in each loop iteration since the -dealloc method already does that.

Now, I get that this is inspired by seeing a loop in the autorelease code, but the solution to that is to fix whatever is causing the data structure to be corrupted, and this does not help.

I think the thing to do is develop a minimal testcase with the simplest code possible which reproduces the problem, then go through that to see how/why things are being corrupted.  Something anywhere in the app, writing over random memory is a possiblity, but that kind if thing is more likely to cause a crash than a loop.  Corruption of the links in the data structure such that the code loops are more likely a re-entrancy issue or a threading issue.  Both of those seem really unlikely/hard to cause though. Every autorelease pool is owned by a thread and should never be modified by another thread (it's hard to see how a pool from one thread could even be passed into another).  Similarly, it's an error to retain or release an autorelease pool (or to autorelease an autorelease pool) so it's hard to see how re-entrancy during d4eallocation of a pool could occur.

If running under gdb to find the bug is too hard or does not work, I suppose we could conditionally compile in test code  to check the consistency of the parent/child links before/after deallocating each pool, and to check that the thread deallocating a pool is the thread that owns it.

Finally, looking at the GUI issue being related to removing a notification observer;  that *is* code which has changed relatively recently, to add support/use of zeroing weak pointers.  So my best guess would be that the issue is to do with some difference between the way that libobjc2 handles that and the code for the gcc runtime (where weak pointers are implemented in the base library), while I can't see how that would interact with the autorelease pools, it seems the most recent change that might be in some way related.

fredkiefer:
I think Richard is correct. This PR should be closed, as it does neither help to solve the issue nor to pin it down. We need to come up with an idea that does at least one of these two things.

IOOI-SqAR:
> I think Richard is correct. This PR should be closed, as it does neither help to solve the issue nor to pin it down. We need to come up with an idea that does at least one of these two things.

Hey @fredkiefer , before you close that PR please consider this patch, which I created with the help of DeepSeek (I know, you hate this, but have please at least a look. I tested it and it works):

```diff
diff --git a/Source/NSAutoreleasePool.m b/Source/NSAutoreleasePool.m
index 9a37e52c8..129d08373 100644
--- a/Source/NSAutoreleasePool.m
+++ b/Source/NSAutoreleasePool.m
@@ -197,21 +197,26 @@ pop_pool_from_cache(struct autorelease_thread_vars *tv)
    */
   if (nil != _child)
     {
-      NSAutoreleasePool	*pool = _child;
+      // First, collect all child pools by following _child links.
+      // This gives us order from nearest child to deepest child.
+      NSAutoreleasePool **childList = NULL;
+      NSUInteger count = 0;
+      NSAutoreleasePool *pool = _child;
+      while (nil != pool)
+        {
+          childList = realloc(childList, (count + 1) * sizeof(NSAutoreleasePool *));
+          childList[count] = pool;
+          count++;
+          pool = pool->_child;
+        }
 
-      /* Find other end of linked list ... oldest child.
-       */
-      while (nil != pool->_child)
-	{
-	  pool = pool->_child;
-	}
-      /* Deallocate the children in the list.
-       */
-      while (pool != self)
-	{
-	  pool = pool->_parent;
-	  [pool->_child dealloc];
-	}
+      // Deallocate children in reverse order (deepest first, i.e. oldest first).
+      // This matches the original behaviour and avoids stack overflow.
+      for (NSUInteger i = count; i > 0; i--)
+        {
+          [childList[i-1] dealloc];
+        }
+      free(childList);
     }
 }
```

resulting in this method:

```objc
- (void) _emptyChild
{
  /* If there are NSAutoreleasePool below us in the list of
   * NSAutoreleasePools, then deallocate them also.
   * The (only) way we could get in this situation (in correctly
   * written programs, that don't release NSAutoreleasePools in
   * weird ways), is if an exception threw us up the stack.
   * However, if a program has leaked pools we may be deallocating
   * a pool with LOTS of children. To avoid stack overflow we
   * therefore deallocate children starting with the oldest first.
   */
  if (nil != _child)
    {
      // First, collect all child pools by following _child links.
      // This gives us order from nearest child to deepest child.
      NSAutoreleasePool **childList = NULL;
      NSUInteger count = 0;
      NSAutoreleasePool *pool = _child;
      while (nil != pool)
        {
          childList = realloc(childList, (count + 1) * sizeof(NSAutoreleasePool *));
          childList[count] = pool;
          count++;
          pool = pool->_child;
        }

      // Deallocate children in reverse order (deepest first, i.e. oldest first).
      // This matches the original behaviour and avoids stack overflow.
      for (NSUInteger i = count; i > 0; i--)
        {
          [childList[i-1] dealloc];
        }
      free(childList);
    }
}
```

If you want to read the whole story please have a look at: https://chat.deepseek.com/share/2cevhvuhv984zlyfx6

Kind regards,

Lars

IOOI-SqAR (commented):
Btw. I tested this on Ubuntu 24.04 ARM64 with AddressManager

fredkiefer:
Hi Lars,

just to make things clear. I don't hate AI generated code, what I really don't like is AI slob. And what DeepSeek came up with here, does not fall in this category. Most likely it correctly identified the cause of the problem an autorelease pool that gets created in the dealloc of an object. This really will corrupt the pool chain and what it came up with is almost a solution. There are two issues with that though. One is minor, we should not reallocate a buffer for each step in the loop. We normally start of with an array on the stack and only move to heap memory when that is exhausted. There is even a macro for this, so it should be easy to fix.
The other issue is worse. When the analysis is correct and a new pool gets created during the dealloc, this pool will now itself be in an incorrect state as its parent was dealloced. I see no simple way to address this. Perhaps with a flag in NSAutorelasePool, that signals that we are in a dealloc? Maybe we can postpone this to a later time? And as Richard already wrote, we need to find the root cause of this behaviour and fix that.

rfm (commented):
I agree with Fred.  While this analysis sounds plausible there are a few problems:
0. LLMs are good at doing plausible, which leads to ..
1. If the analysis is  correct, why did it not provide a simple testcase to reproduce the problem: I think we need something we can put in the base library regression tests.
2. I think the code is supposed to be protected against re-entrancy: so if there's a bug in that protection, it's what we need to fix, rather than doing a workaround.
3. As Fred said, the workaround with realloc/free is very inefficient, so even if we found fixing reentrancy earlier more problematic, we would want this code to be much more efficient.

rfm:
One option to avoid re-entrancy while maintaining the active pool might be to detach all children to begin with and loop to catch pools created while children are being emptied.

  while (nil != _child)
    {
      NSAutoreleasePool *pool = _child;

      _child = nil;
      pool->_parent = nil;
      /* Find other end of linked list ... oldest child.
       */
      while (nil != pool->_child)
        {
          pool = pool->_child;
        }
      /* Deallocate the children in the list.
       */
      while (pool != nil)
        {
          NSAutoreleasePool     *parent = pool->_parent;

          [pool dealloc];
          pool = parent;
        }
    }

rfm:
>>> and loop to catch pools created while children are being emptied.
But that's not necessary since the code calling -emptyPool already does it.

rfm:
Neither copying the children into an array nor detaching the children from the linked list solves the issue of an uncaught exception during deallocation (they both leak).  To right really safe code we need to really understand the root cause (ie we need the testcase).

IOOI-SqAR:
Hi Fred,

> Hi Lars,
> 
> just to make things clear. I don't hate AI generated code, what I really don't like is AI slob. And what DeepSeek came up with here, does not fall in this category. Most likely it correctly identified the cause of the problem an autorelease pool that gets created in the dealloc of an object. This really will corrupt the pool chain and what it came up with is almost a solution. There are two issues with that though. 

> One is minor, we should not reallocate a buffer for each step in the loop. We normally start of with an array on the stack and only move to heap memory when that is exhausted. There is even a macro for this, so it should be easy to fix. 

I wasn't aware of such a macro

> The other issue is worse. When the analysis is correct and a new pool gets created during the dealloc, this pool will now itself be in an incorrect state as its parent was dealloced. I see no simple way to address this. Perhaps with a flag in NSAutorelasePool, that signals that we are in a dealloc? 

That is something the AI came up with too, but this didn't work when tested. I copied this from the session I had with DeepSeek: https://chat.deepseek.com/share/2cevhvuhv984zlyfx6

```diff
--- a/Source/NSAutoreleasePool.h
+++ b/Source/NSAutoreleasePool.h
@@ -32,6 +32,8 @@
   unsigned int _flags;
   // The low bit indicates that the pool is a 'full' pool (as opposed to a 
   // single-object pool).  The second bit indicates it's a 'reserved' pool.
+  // The third bit (0x04) indicates that -_emptyChild is currently running
+  // on this pool, to prevent re‑entrant corruption.
   id _token;        // For internal use.
   struct _NSAutoreleasePoolList *_list;
   struct _NSAutoreleasePoolPage *_page;
```

```diff
--- a/Source/NSAutoreleasePool.m
+++ b/Source/NSAutoreleasePool.m
@@ -180,6 +180,9 @@
   return self;
 }

+// Bit in _flags to indicate that _emptyChild is in progress
+#define POOL_EMPTYING_FLAG 0x04
+
 /**
  * Remove and deallocate all the children of the receiver.  This is used to
  * clean up after exceptions.
@@ -187,6 +190,12 @@
 - (void) _emptyChild
 {
   GS_ASSERT(_parent != nil);
+
+  // Prevent re‑entrant execution that could corrupt the parent chain.
+  if (_flags & POOL_EMPTYING_FLAG) {
+    return;
+  }
+  _flags |= POOL_EMPTYING_FLAG;
 
   // Deallocate all our children (oldest first)
   if (nil != _child)
@@ -216,6 +225,7 @@
         }
       free(childList);
     }
+  _flags &= ~POOL_EMPTYING_FLAG;
 
   // Remove from the thread's pool list.
   if (_list != 0)
```

Note, that those patches aren't perfect, the AI had some hallucinations here (`GS_ASSERT(_parent != nil);`) and the patch doesn't apply, but I guess you get the idea.

> Maybe we can postpone this to a later time? And as Richard already wrote, we need to find the root cause of this behaviour and fix that.

I strongly recommend to read into https://chat.deepseek.com/share/2cevhvuhv984zlyfx6 and see all the ways that the AI proposed but didn't work in the end. It wasn't simply: "Hey AI, I need a fix!" and then took the first output of it and post it here, it was merely a back and forth between multiple attempts and failures to fix this.

The main point seems to be that we need to decouple the deallocation of the children from finding those.

I hope this clears things up a bit,

Lars

IOOI-SqAR:
Hi @rfm,

> I agree with Fred. While this analysis sounds plausible there are a few problems: 0. LLMs are good at doing plausible, which leads to ..
> 
> 1. If the analysis is  correct, why did it not provide a simple testcase to reproduce the problem: I think we need something we can put in the base library regression tests.

The AI proposed some test code which didn't reproduce the issue when run. You can find it inside https://chat.deepseek.com/share/2cevhvuhv984zlyfx6 , maybe we can start from this.

> 2. I think the code is supposed to be protected against re-entrancy: so if there's a bug in that protection, it's what we need to fix, rather than doing a workaround.

The AI also produced several attempts to protect against reentrancy, but when testing those patches it turned out that reentrancy wasn't the real problem here, in AddressManager it happened on the first call to _emptyChild. As I said to @fredkiefer earlier, it is helpful to read the whole conversation I linked above.

> 3. As Fred said, the workaround with realloc/free is very inefficient, so even if we found fixing reentrancy earlier more problematic, we would want this code to be much more efficient.

If you have a better idea, I am all open to it.

It seems like we have several ARP referencing each other in a cycle, but a larger one, where more than one ARP seems to be involved. At least that's my gut feeling.

Kind regards,

Lars

rmottola:
> I don't see any point to this change: Using a do loop rather than a while loop should make no functional difference (decades ago with really dumb compilers it would have saved one cpu instruction, but I expect a modern compilers to optimise that away anyway. Setting pool->_child to nil is a wasted instruction in each loop iteration since the -dealloc method already does that.

the do-loop conversion is not for efficiency, but for logical reasons. Given the lines above, the loop body should execute at least once. I think it is nice to be explicit.

We set the pool->_child for safety.

In dealloc, I see
```
      _parent->_child = nil;
      _parent = nil;
```

If it is really certain to be equivalent, then it is a wasted instruction.



> I think the thing to do is develop a minimal testcase with the simplest code possible which reproduces the problem, then go through that to see how/why things are being corrupted. Something anywhere in the app, writing over random memory is a possiblity, but that kind if thing is more likely to cause a crash than a loop. Corruption of the links in the data structure such that the code loops are more likely a re-entrancy issue or a threading issue. Both of those seem really unlikely/hard to cause though. Every autorelease pool is owned by a thread and should never be modified by another thread (it's hard to see how a pool from one thread could even be passed into another). Similarly, it's an error to retain or release an autorelease pool (or to autorelease an autorelease pool) so it's hard to see how re-entrancy during d4eallocation of a pool could occur.


If you see the error in the stack trace, the last stack is indeed about accessing thread variables.
 
> If running under gdb to find the bug is too hard or does not work, I suppose we could conditionally compile in test code to check the consistency of the parent/child links before/after deallocating each pool, and to check that the thread deallocating a pool is the thread that owns it.

Stack is really bad leading in my attempts a hang of gdb. I could only analyse core files where the stack itself would loop untilgdb dies.

> Finally, looking at the GUI issue being related to removing a notification observer; that _is_ code which has changed relatively recently, to add support/use of zeroing weak pointers. So my best guess would be that the issue is to do with some difference between the way that libobjc2 handles that and the code for the gcc runtime (where weak pointers are implemented in the base library), while I can't see how that would interact with the autorelease pools, it seems the most recent change that might be in some way related.

The issue is removing an observer which was not set. As I wrote at the beginning, it is indeed a libobjc2 only issue, the reason why I didn't confirm this bug until recently, reproducing after Lars' QEMU findings.

rfm:
I did read the AI thread, but the fact that it can't write a testcase demonstrating the issue suggests that its analysis is wrong;  which leaves us where we were before the AI was involved (memory corruption, threading issue, or re-entrancy issue).  It agreed that a re-entrancy issue was most likely, but since it hasn't pinpointed it, we aren't really any further forward.  Applying it's patch has its own problems and if it hides the underlying issue that's worse than useless.

>If you have a better idea, I am all open to it.

Well, I think my initial suggestions are better:  try to create a simple testcase, and if that needs more understanding, get that by stepping through the crash process in gdb and/or by adding consistency checks to a local copy of the code to spot the point at which the list is corrupted.

rmottola:
I don't fully understand why dealloc'ing in the allocation order instead of reverse-order should really bring a benefit.

I think that the patch helps getting around corruption: a separate list that doesn't get corrupted allows for going through, while walking back the doub-linked lists exposes corruption.

> The other issue is worse. When the analysis is correct and a new pool gets created during the dealloc, this pool will now itself be in an incorrect state as its parent was dealloced. I see no simple way to address this. Perhaps with a flag in NSAutorelasePool, that signals that we are in a dealloc? Maybe we can postpone this to a later time? And as Richard already wrote, we need to find the root cause of this behaviour and fix that.

Do you mean a pool gets created during the ARP release or in another object? at by checking only NSAutoreleasePool, no new allocation is done.

However, it empties the pool only at the end:
```
  /* Don't deallocate ourself, just save us for later use. */
  push_pool_to_cache(tv, self);
  GSNOSUPERDEALLOC;
```

rfm:
PS. as you suggest, we can also try to build a testcase from the DeepSeek suggestion, but since that didn't work it may not be much better than starting from scratch.  I will have a try though.

rfm:
>> I don't fully understand why dealloc'ing in the allocation order instead of reverse-order should really bring a benefit.

I think originally because it simplifies the linked list management to keep the code maintainable.  Now, with libobjc2 it matters because it has to integrate with the libobjc2 autorelease pool stack management (push and pop) so you can't just unlink from the middle.

rfm:
I updated/built the latest libraries and the AddressManager app, and was able to reproduce the issue.
I had problems with gdb crashing; I don't have Wolfgang's patched version for libobjc2 on my system :-( so I made do with a combination of gdb and adding the sanity checks I suggested, and was able to narrow down the issue. It seesm that libobjc2 is corrupting the autorelease pool links list behind the back of the base library in objc_autoreleasePoolPop(),  and probably the fix is simply additional calls to get rid of child pools after that function is called.  I still haven't worked out the exact sequence of events to trigger the issue though.

rfm:
So what initially happens is this:
We drain/release/deallocate a pool, and that calls -emptyPool
The libobjc2 version of -emptyPool calls _emptyChild, which removes any children, then it calls objc_autoreleasePoolPop() which deallocates objects and in doing so creates child pools, and returns
Then -dealloc removes the pool from the linked it and pushes it to the cache (with an attached child pools)

Now, in and of itself, that might not be a huge issue because when we need a new pool we pop the old one from the cache and use it along with the attached child pools, which should therefore get emptied when the pool is next deallocated (late but not a permanent leak).]
However, the code linking a new pool pool into the list does not work correctly when the *previously* popped pool had children, and this is what actually corrupts the linked list.

We can (and should) fix the code to cope with the unexpected child pools, which will prevent the crash, but that leaves the issue of objects not being deallocated when they should.

We can fix the late deallocation by calling _emptyChild *after* objc_autoreleasePoolPop() as well as before it.

The _emptyChild method has nothing to do with this issue.
However, as a completely separate issue, to avoid possible problems if an exception is raised while executing _emptyChild we should probably rewrite it as follows (perhaps inline rathe than as a method since it's so small):

  while (_child)
    {
      NSAutoreleasePool *pool = _child;

      while (pool->_child)
        {
          pool = pool->_child;
        }
      [pool dealloc];
    }

IOOI-SqAR:
Hi @rfm, @rmottola, @fredkiefer , @gcasa  I finally got a minimal test to reproduce the bug. Co-developed with DeepSeek again but to get it working was my part. At first, here's the test:

```objc
#import <objc/runtime.h>

static void setIvar(id object, const char *name, id value) {
    Ivar ivar = class_getInstanceVariable(object_getClass(object), name);
    if (ivar) {
        void *addr = (char *)object + ivar_getOffset(ivar);
        *(id *)addr = value;
    }
}

int main() {
    // Create a normal chain: outer (A) -> middle (B) -> inner (C)
    NSAutoreleasePool *A = [NSAutoreleasePool new];
    NSAutoreleasePool *B = [NSAutoreleasePool new];

    // Create a separate corrupt pool X
    NSAutoreleasePool *X = [NSAutoreleasePool new];

    // Corrupt B's parent to point to X (instead of A)
    setIvar(B, "_parent", X);
    setIvar(B, "_child", nil);

    // Set X's parent to nil and its child to A
    setIvar(X, "_child", A);

    // Drain A – this will cause infinite recursion in unpatched GNUstep Base
    [A drain];

    NSLog(@"Test passed – bug not triggered (patched base)");
    return 0;
}
```

and here is the (long) way to it: https://chat.deepseek.com/share/s2twff3633ps4jitsc

Kind regards,

Lars

gcasa:
@IOOI-SqAR Let’s start from the top…. What concerns me about this test is this… it starts off from a questionable position.   It manually corrupts the autorelease pool and then treats the result as a legitimate runtime state.  It doesn’t test the normal case.   It deliberately corrupts the pool.

I think it helps identify the issue… but if the issue is happening in GUI, then I think the fix needs to be localized there.

So… what the NSTableView code is doing is likely corrupting the pool like in this example… but that still means the issue is in GUI.   What this fix does is fix a case that, in normal usage, isn't likely to happen and, if it does, *SHOULD* cause a crash.

gcasa:
Allow me to further clarify...  I strongly believe the issue is in MY implementation of NSTableView viewBased cells.   I think that this fix, while it could mask the crash is not the right fix because it could hide the issue and WORSE hide potentially sloppy implementations like my viewBased cell implementation apparently was.   Failures are important... it shows us where the issue REALLY is.  If we put this in, it could hide potential failures and make debugging harder going forward.

I am working on a fix in GUI.  I have submitted it as https://github.com/gnustep/libs-gui/pull/437.   I have yet to test it. Once I do i will update the PR over there.  Please make any comments you have over there.

gcasa:
# Proof of the fix's incorrectness:
---
## LEMMA: Only correct code should execute properly on GNUstep.

### ASSUME: This app would work on macOS

* If the app works on macOS, then the inconsistent state of the autorelease pool created by the code in NSTableView cannot be correct, therefore the fix masks the issue instead of fixing an actual problem.
* Therefore: the fix is incorrect.

### ASSUME: This app would NOT work on macOS (so the code in AddressManager is wrong):

* If the app does NOT work on macOS, then there is no reason for this fix as incorrect code should not function.
* Therefore: the fix is incorrect

## THEREFORE: By the assumption put forward in the LEMMA: The proposed fix either hides an issue or allows incorrect code to execute, so the fix cannot be right. 
QED. (Thus it is shown)

* Sorry, this is me showing a bit of my academics. ;) Rather, the issue is the NSTableView changes, which I am working on fixing.

rfm:
In summary, the actual problem was unrelated to emptying pools and the AI analysis was wrong, as was the test it generated.  

I have fixed the actual autorelease issue.

I also modified _emptyChild to be safe in the event of re-entrance (though that was not an issue in this case, and I'm not sure it ever can be).

NB the fact that there was an issue here does not necessarily mean there is no problem in the GUI ...

PR #618, opened 2026-06-19
NSJSONSerialization: reject over-long numeric tokens (stack buffer overflow)

DTW-Thalion:
## Problem

`parseNumber()` accumulates a JSON numeric token into a fixed 255-byte stack buffer (`char number[255]`) via the `BUFFER()` macro. The overflow guard raises a parse error when the buffer is full but does not stop:

```c
#define BUFFER(x) do {\
  if (parsedSize == sizeof(number))\
    {\
      parseError(state);\
    }\
  number[parsedSize++] = (char)x; } while (0)
```

`parseError()` only sets `state->error` and returns normally, so control falls through to `number[parsedSize++] = (char)x`, writing one byte past the end of the buffer — and then continuing for every further character. The digit-consuming loops (`while (isdigit(c = consumeChar(state))) { BUFFER(c); }`) test only `isdigit()` and never inspect `state->error`, so a numeric token longer than 255 characters overruns the stack buffer without bound. This is an attacker-controlled stack buffer overflow reachable from `+JSONObjectWithData:` / `+JSONObjectWithStream:` on untrusted JSON.

## Fix

Return from `parseNumber()` as soon as the error is raised, so an over-long numeric token is rejected via the error out-parameter instead of overflowing the buffer. One added line.

## Validation (Ubuntu 24.04, clang 18, libobjc2, current master)

- **Unpatched:** parsing a 4096-character numeric token via `+JSONObjectWithData:` segfaults — the new regression test runs 2 assertions, then the process takes SIGSEGV (core dumped).
- **Patched:** the token is rejected via the error out-parameter; the regression test passes all 4 assertions with no crash.
- An AddressSanitizer replica of the exact `BUFFER()` loop reports `stack-buffer-overflow WRITE of size 1` one byte past the 255-byte buffer.

Adds `Tests/base/NSJSONSerialization/number.m` (a normal number still parses; an over-long token is rejected) and a ChangeLog entry.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2026-06-19 Todd White <todd.white@thalion.global>
+
+	* Source/NSJSONSerialization.m (parseNumber): the BUFFER() macro raised a
+	parse error when the 255-byte numeric buffer was full but did not stop, so
+	a numeric token longer than the buffer overran the stack array.  Return as
+	soon as the error is raised so an over-long token is rejected instead of
+	overflowing the buffer.
+	* Tests/base/NSJSONSerialization/number.m: new regression test.
+
 2026-06-18 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Documentation/Base.gsdoc: minor tidy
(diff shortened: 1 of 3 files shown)


rfm:
Thanks.  That looks good (I tried it on windows/gcc as an additional check).

PR #633, opened 2026-06-20
GSString: range-check before the tiny-string fast path in GSCString substring

DTW-Thalion:
## Summary

`-[GSCString substringFromRange:]` and `-[GSCString substringWithRange:]` run the "tiny string" fast path before validating the requested range:

```objc
- (NSString*) substringFromRange: (NSRange)aRange
{
  if (!_flags.wide)
    {
      id tinyString;
      tinyString = createTinyString((char*)_contents.c + aRange.location,
        aRange.length);                       // <-- no range check yet
      if (tinyString) { return tinyString; }
    }
  if (_flags.owned)
    {
      GS_RANGE_CHECK(aRange, _count);          // too late / skipped
      ...
```

`createTinyString(str, length)` reads up to 9 bytes from `str` (it dereferences `str[8]` for `length == 9` and `str[i]` for `i < length`). Because `str = _contents.c + aRange.location` and `length = aRange.length` come straight from the caller's `NSRange`, an out-of-range range reads past the end of the character buffer and folds the over-read bytes into the returned tagged-pointer string.

`-substringWithRange:` has the same defect on the non-owned branch. The sibling `GSMutableString` implementations already do `GS_RANGE_CHECK` first; these two did not.

Reachable via e.g. `-[NSString initWithBytesNoCopy:length:encoding:freeWhenDone:NO]` (a non-owned `GSCString`) followed by `-substringWithRange:`/`-substringFromRange:` (and `-substringFromIndex:`/`-substringToIndex:`, which funnel through them) with an out-of-range range. The fast path is active under the libobjc2 tagged-pointer runtime.

## Fix

Hoist `GS_RANGE_CHECK(aRange, _count)` to the first statement of both methods (and drop the now-redundant check in the owned branch), matching `GSMutableString`.

## Test

`Tests/base/NSString/substring-bounds.m`: in-range substrings still work; an out-of-range range raises `NSRangeException`.

## Validation (Ubuntu 24.04, clang 18, libobjc2; tiny strings active)

- **Positive control** (patched, in-tree lib): 4/4 pass.
- **Negative control** (unpatched in-tree lib): the out-of-range substring runs against a non-owned `GSCString` whose 16 bytes end exactly at a page boundary backed by a `PROT_NONE` guard page; `createTinyString` reads past the page and the test program **aborts (SIGSEGV)** — the harness reports "1 Failed file". The patched build rejects the range before the read. (The installed system copy already carries an older audit-era check, so the negative control must be run against a clean build of current master.)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,14 @@
+2026-06-20 Todd White <todd.white@thalion.global>
+
+	* Source/GSString.m (-[GSCString substringFromRange:],
+	-[GSCString substringWithRange:]): both ran the tiny-string fast path
+	- createTinyString((char*)_contents.c + aRange.location,
+	aRange.length) - before validating aRange against the receiver length.
+	createTinyString reads up to 9 bytes from that pointer, so an
+	out-of-range range read past the end of the character buffer.  Perform
+	GS_RANGE_CHECK first, as the GSMutableString implementations do.
+	* Tests/base/NSString/substring-bounds.m: new regression test.
+
 2026-06-20 Richard Frith-Macdonald <rfm@gnu.org>
 
         * Headers/Foundation/NSDecimalNumber.h:
--- a/Source/GSString.m
+++ b/Source/GSString.m
@@ -3880,6 +3880,8 @@ agree, create a new GSCInlineString otherwise.
 
 - (NSString*) substringFromRange: (NSRange)aRange
 {
+  GS_RANGE_CHECK(aRange, _count);
+
   if (!_flags.wide)
     {
       id tinyString;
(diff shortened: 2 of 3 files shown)


rfm:
That's a lot of fixes, but all good.  Small, simple, easy to review changes and straightforward testcases.  Thanks very much.
I was able to apply each pull locally on a windows (mingw with gcc) sysytem and validate testcases there with before/after checks (in addition to the Unbuntu/clang ones you did).

DTW-Thalion:
I am trying to resurface them in small groups from our internal testing program, separate PRs for each change and trying not to overwhelm - hopefully the descriptions are suitable.

PR #637, opened 2026-06-29
NSDecimal: fix mantissa buffer overflow on maximum-length values

DTW-Thalion:
## Summary

Two `NSDecimal` operations write one element past the fixed
`cMantissa[NSDecimalMaxDigit]` array (38 bytes, the last field of the struct)
when given a maximum-length 38-digit mantissa.

### GSSimpleAdd — carry-out overflow

When the sum carries out to a 39th digit and the result already holds 38
digits, the guard tries to make room:

```c
if (NSDecimalMaxDigit == result->length)
  {
    NSDecimalRound(result, result, NSDecimalMaxDigit - 1 - result->exponent, mode);
  }
```

That scale leaves `l = scale + exponent + length = 37 + length > length`, so
`GSDecimalRound`'s `if (result->length <= l) return;` makes the round a **no-op**.
The carry-shift then runs with `length == 38` and writes `cMantissa[38]`.

### GSDecimalRound — `l == 0` left-shift overflow

Rounding a full-length value at a scale that keeps no significant digits
prepends a leading zero by shifting the mantissa right:

```c
x = result->length;            /* 38 */
result->length += 1;           /* 39 */
while (x > 0) { result->cMantissa[x] = result->cMantissa[x-1]; x--; }  /* writes cMantissa[38] */
```

```objc
NSDecimal a, r;
NSDecimalFromString(&a, @"99999999999999999999999999999999999999", nil); /* 38 nines */
NSDecimalAdd(&r, &a, &a, NSRoundPlain);          /* heap-buffer-overflow, GSSimpleAdd:1481 */
NSDecimalRound(&r, &a, -38, NSRoundPlain);       /* heap-buffer-overflow, GSDecimalRound:304 */
```

Both are reachable from the public `NSDecimalAdd` / `NSDecimalRound` and the
`NSDecimalNumber` wrappers (e.g. `[big decimalNumberByAdding: big]`).

## Fix

- **GSSimpleAdd**: make room correctly — decide rounding from the least
  significant digit, drop it, raise the exponent, then prepend the carried `1`.
  The prepended digit is `1`, so the subsequent round-up cannot overflow the
  mantissa again. (The previous code's no-op round also produced a wrong value
  when it *did* round, because round-then-prepend doesn't compose; the rewrite
  fixes the value too — `99…9 + 99…9` now yields `2e38`.)
- **GSDecimalRound**: stop the prepend shift at the last array element. The
  dropped least significant digit cannot change the single digit that survives
  (length becomes 1), and the length used for the exponent adjustment is kept,
  so `round(99…9, scale −38)` still yields `1e38`.

## Test

`Tests/base/NSDecimalNumber/mantissa_overflow.m` checks both operations produce
the correct value without overflowing, and that ordinary (non-maximal) add and
round are unaffected. It aborts under AddressSanitizer without the fix and
passes with it (4/4).

### DIFF
--- a/Source/NSDecimal.m
+++ b/Source/NSDecimal.m
@@ -295,10 +295,19 @@ GSDecimalRound(GSDecimal *result, int scale, NSRoundingMode mode)
       if (l == 0)
         {
           int x;
-             
+
           x = result->length;
           result->length += 1;
           l += 1;
+          if (x >= NSDecimalMaxDigit)
+            {
+              /* No room to prepend a digit into the fixed-size mantissa: stop
+                 the shift at the last element, dropping the least significant
+                 digit.  Only the prepended digit survives the rounding below
+                 (length becomes l == 1), so the dropped digit cannot change
+                 the result. */
+              x = NSDecimalMaxDigit - 1;
+            }
           while (x > 0)
             {
                result->cMantissa[x] = result->cMantissa[x-1];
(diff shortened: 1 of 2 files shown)


rfm (approved):
Thanks ... merging.

PR #640, opened 2026-06-30
NSPropertyList: bound OpenStep typed-value token length

DTW-Thalion:
## Problem

The old-style (OpenStep) property-list parser handles the GNUstep-extended
typed values `<*I...>` (integer), `<*D...>` (date) and `<*R...>` (real) by
copying the token body into a stack buffer whose size is taken straight
from the token length:

```c
len = pld->pos - min;        /* distance to the closing '>' */
...
char    buf[len+1];          /* <*I... and <*R... */
unichar buf[len];            /* <*D...  (2*len bytes) */
```

`len` is bounded only by the size of the input, and these are C99
variable-length arrays on the stack, so a token of a few hundred KB
overflows the stack.

The token is attacker-controlled and reachable from the public
`+propertyListWithData:options:format:error:`: any data that is not a
binary or XML plist is parsed as `NSPropertyListOpenStepFormat`, and a
typed value may be the top-level item.

### Reproduced

Built `libs-base` with AddressSanitizer and parsed `<*I` + 1 MB of `0` + `>`
on a worker thread with a 512 KB stack:

```
ERROR: AddressSanitizer: stack-overflow ... in parsePlItem NSPropertyList.m:1140
    #0 parsePlItem NSPropertyList.m:1140                                  (buf[i] = (char)ptr[i])
    #1 +[NSPropertyListSerialization propertyListWithData:options:format:error:] NSPropertyList.m:2851
    #2 worker thread (512 KB stack)
```

The main thread does not crash until the token exceeds its (much larger)
stack, but any secondary thread with a normal smaller stack crashes on a
sub-megabyte token.

## Fix

A valid typed value is a short number or a fixed-format date (the writer
emits `[NSNumber stringValue]` / a `%Y-%m-%d %H:%M:%S %z` date), so reject
an over-long token before allocating the buffer rather than overflow the
stack. A single guard covers all three branches.

## Test

`Tests/base/NSPropertyList/openstep-typed-value-bounds.m` checks that a
normal `<*I...>`, `<*R...>` and `<*D...>` still parse, and that an
over-long `<*I...>` / `<*R...>` token is rejected (returns `nil`) instead
of being parsed. Validated before/after with an ASan build:

- without the fix: the two over-long cases are parsed instead of rejected (2 failed, 3 passed)
- with the fix: 5/5 pass, and the full `NSPropertyList` suite is 17/17.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,16 @@
+2026-06-29 Todd White <todd.white@thalion.global>
+
+	* Source/NSPropertyList.m (parsePlItem): the GNUstep-extended
+	OpenStep typed values <*I...>, <*D...> and <*R...> copied their
+	token body into a stack buffer (char buf[len+1] / unichar buf[len])
+	whose size was taken straight from the token length, so a token of
+	a few hundred KB overflowed the stack.  The token is reachable from
+	+propertyListWithData:options:format:error: with attacker-supplied
+	data.  A valid typed value is a short number or date, so reject an
+	over-long token before allocating the buffer.
+	* Tests/base/NSPropertyList/openstep-typed-value-bounds.m: new
+	regression test.
+
 2026-06-21 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Documentation/BaseAdditions.gsdoc:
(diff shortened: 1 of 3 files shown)


rfm (approved):
Tested on ubuntu gcc and windows gcc

PR #641, opened 2026-06-30
NSXMLDTDNode: copy strings passed to the entity-ID setters

DTW-Thalion:
## Problem

`-[NSXMLDTDNode setNotationName:]`, `-setPublicID:` and `-setSystemID:`
store `XMLSTRING(arg)` — the transient `[arg UTF8String]` buffer —
directly into the persistent libxml2 entity fields:

```objc
- (void) setNotationName: (NSString*)notationName
{
  internal->node.entity->name = XMLSTRING(notationName);        // transient buffer
}
- (void) setPublicID: (NSString*)publicID
{
  internal->node.entity->ExternalID = XMLSTRING(publicID);
}
- (void) setSystemID: (NSString*)systemID
{
  internal->node.entity->ExternalID = XMLSTRING(systemID);      // also: wrong field
}
```

`-[NSString UTF8String]` returns a buffer owned by the (usually
autoreleased) argument, so once that pool drains the entity field
dangles. This is reached from ordinary public API; no parsing of
untrusted input is required.

### Reproduced (AddressSanitizer)

Storing an autoreleased string and reading it back after the pool drained:

```
ERROR: AddressSanitizer: heap-use-after-free READ of size 12
    #0 strlen
    #1 StringFromXMLStringPtr NSXMLPrivate.h:94
    #2 -[NSXMLDTDNode notationName] NSXMLDTDNode.m:102
  freed by: object_dispose            (the argument string was deallocated)
```

and at `-dealloc`, `xmlFreeEntity` frees the same non-owned pointer:

```
ERROR: AddressSanitizer: attempting free on address which was not malloc()-ed
    ... object_dispose -> ... -> xmlFreeEntity
```

## Fix

Copy the argument with `XMLStringCopy`, exactly as the sibling
`-[NSXMLDTD setPublicID:]` / `-setSystemID:` already do (those store an
owning `malloc`-ed copy that `xmlFreeEntity` can free).

`-setSystemID:` additionally wrote the `ExternalID` field rather than
`SystemID`, so `-systemID` never returned the value that was set and a
`setPublicID:` / `setSystemID:` pair clobbered each other. Store the
system ID in `SystemID`.

## Test

`Tests/base/NSXMLDTDNode/entity-id-persistence.m` sets the three IDs from
strings whose autorelease pool then drains, churns the allocator, and
checks the values survive. Validated before/after with an ASan build:

- without the fix: `publicID`/`systemID` fail deterministically (the
  wrong-field bug), and `-release` aborts under ASan with the invalid free
- with the fix: 3/3 pass; the full NSXMLNode/NSXMLDocument/NSXMLElement
  suites (245/36/132) remain green.

The test is gated on `GS_USE_LIBXML` and `SKIP`s when libxml2 is absent.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,18 @@
+2026-06-29 Todd White <todd.white@thalion.global>
+
+	* Source/NSXMLDTDNode.m (-setNotationName:, -setPublicID:,
+	-setSystemID:): these stored XMLSTRING(arg) - the transient
+	[arg UTF8String] buffer - directly in the persistent libxml2 entity
+	fields, so once the autorelease pool that produced the string
+	drained the field dangled: a getter read freed memory and -dealloc
+	xmlFreeEntity freed a pointer libxml2 never allocated.  Copy the
+	argument with XMLStringCopy, as -[NSXMLDTD setPublicID:] and
+	-setSystemID: do.  -setSystemID: also wrote ExternalID instead of
+	SystemID, so -systemID never returned the value set; store it in
+	SystemID.
+	* Tests/base/NSXMLDTDNode/entity-id-persistence.m: new regression
+	test.
+
 2026-06-21 Richard Frith-Macdonald <rfm@gnu.org>
 
 	* Documentation/BaseAdditions.gsdoc:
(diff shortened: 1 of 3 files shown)


rfm (approved):
Thanks.  Looks fine.

PR #643, opened 2026-06-30
NSPropertyList: fix one-byte over-read parsing old-style <hex> data

DTW-Thalion:
Parsing an old-style (OpenStep) `<...>` data value reads one byte past the end of the buffer.

After the initial scan finds the closing `>`, the hex octets are parsed with `skipSpace()` called between them to allow whitespace. `skipSpace()` also skips `/* ... */` comments — and a comment can contain the `>` that the initial scan found and run to the end of the buffer, leaving `pos == end`. Its return value is ignored (`(void)skipSpace(pld)`), so the terminator check then dereferences `ptr[pos]` with `pos == end`:

```objc
(void)skipSpace(pld);          // can consume the '>' and reach end
...
if (pld->ptr[pld->pos] != '>') // ptr[end] — one byte past the buffer
```

AddressSanitizer confirms it — `heap-buffer-overflow READ of size 1 in parsePlItem` (NSPropertyList.m:1329), reachable from the public `+propertyListWithData:options:format:error:` with input such as `<41/*>`.

This guards the dereference with a `pos < end` check; the over-long/unterminated input is then reported as the existing `unexpected character (wanted '>')` parse error.

### Test

`Tests/base/NSPropertyList/data-comment-bounds.m` places the malformed input at the end of a page backed by a `PROT_NONE` guard page, so the one-byte over-read faults deterministically rather than silently reading adjacent memory (Unix only; it skips elsewhere). It also parses a well-formed `<4142>` value as a positive control. Before the fix the test aborts on the guard page; after, it passes (2/2).

### DIFF
--- a/Source/NSPropertyList.m
+++ b/Source/NSPropertyList.m
@@ -1326,7 +1326,10 @@ static id parsePlItem(pldata* pld) NS_RETURNS_RETAINED
                 // We permit (but do not require) space between/after hex octets
 		(void)skipSpace(pld);
 	      }
-            if (pld->ptr[pld->pos] != '>')
+            // skipSpace() can consume a comment that runs to the end of the
+            // buffer (taking the closing '>' with it), leaving pos == end, so
+            // the terminator check must guard against reading past the end.
+            if (pld->pos >= pld->end || pld->ptr[pld->pos] != '>')
               {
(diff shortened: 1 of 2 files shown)


rfm (approved):
All looks good. Thanks.

PR #644, opened 2026-06-30
NSURL: avoid NULL deref resolving a relative URL against a scheme-less base

DTW-Thalion:
`-[NSURL initWithString:relativeToURL:]` dereferences NULL when the base URL has no scheme.

When the relative string carries its own scheme, the parser compares it with the base's scheme to decide whether the relative URL is really absolute:

```objc
if (buf->scheme != 0 && base != 0
  && 0 != strcmp(buf->scheme, base->scheme))   // base->scheme may be NULL
```

The guard only checks `base != 0`, not `base->scheme != 0`. A base `NSURL` parsed from a scheme-less string has `scheme == NULL` (the field is zero-initialised and only set when a `scheme:` prefix is present), so `strcmp()` dereferences NULL.

Reachable from the public API:

```objc
NSURL *base  = [NSURL URLWithString: @"foo"];                        // scheme == nil
NSURL *child = [NSURL URLWithString: @"http:bar" relativeToURL: base]; // crash
```

AddressSanitizer: `SEGV ... READ ... in strcmp` ← `NSURL.m:982` ← `+URLWithString:relativeToURL:`.

This checks `base->scheme` before the `strcmp`: a relative URL that carries a scheme the base lacks is treated as an absolute URL — the same handling already used when the two schemes merely differ. With the fix, the example above resolves to the absolute `http:bar`.

### Test

`Tests/base/NSURL/relative-scheme-base.m` builds the scheme-less base, resolves a scheme-bearing relative URL against it, and checks it does not crash and resolves to `http:bar`. The NULL dereference crashes deterministically, so the test aborts before the fix and passes (3/3) after.

### DIFF
--- a/Source/NSURL.m
+++ b/Source/NSURL.m
@@ -979,10 +979,13 @@ static NSUInteger	urlAlign;
       start = end;
 
       if (buf->scheme != 0 && base != 0
-        && 0 != strcmp(buf->scheme, base->scheme))
+        && (base->scheme == 0 || 0 != strcmp(buf->scheme, base->scheme)))
         {
-          /* The relative URL is of a different scheme to the base ...
-           * so it's actually an absolute URL without a base.
+          /* The relative URL has a scheme which the base lacks or which
+           * differs from the base ... so it's actually an absolute URL
+           * without a base.  (base->scheme may be NULL when the base was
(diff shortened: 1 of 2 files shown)


rfm (approved):
Looks good, thanks.

PR #645, opened 2026-06-30
GSMime: append, don't overwrite, when decoding a body across several calls

DTW-Thalion:
A MIME body encoded with base64 or quoted-printable decodes incorrectly when it arrives across more than one `-parse:` call.

`GSMimeBase64DecoderContext` and `GSMimeQuotedDecoderContext` set their output cursor to the **start** of the destination buffer:

```objc
[dData setLength: size + ...];
dst = (unsigned char*)[dData mutableBytes];   // should be + size
beg = dst;
```

`size` is the amount already decoded into `dData`. Because the body decoder is invoked once per `-parse:` call (i.e. once per network packet for a streamed message), each call overwrote the previously decoded bytes, so only the last chunk's output survived. A base64 body `SGVsbG8sIFdvcmxkIQ==` ("Hello, World!") split across two `-parse:` calls decoded to `lo, World!`.

The base `GSMimeCodingContext` (`mutableBytes + size`) and the chunked decoder (`buf + size`) already append correctly; this makes the base64 and quoted-printable contexts do the same.

### Test

`Tests/base/GSMime/streamed-decode.m` feeds a base64 body in two parts, splitting the encoded data, and checks the document content is `Hello, World!`. It fails before the fix (decodes to `lo, World!`) and passes after.

### DIFF
--- a/Source/Additions/GSMime.m
+++ b/Source/Additions/GSMime.m
@@ -590,7 +590,7 @@ wordData(NSString *word, BOOL *encoded)
    * Expand destination data buffer to have capacity to handle info.
    */
   [dData setLength: size + (3 * (end + 8 - src))/4];
-  dst = (unsigned char*)[dData mutableBytes];
+  dst = (unsigned char*)[dData mutableBytes] + size;
   beg = dst;
 
   /*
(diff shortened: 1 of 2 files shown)


rfm (approved):
Thanks, looks good.

PR #646, opened 2026-06-30
GSMime: bound the RFC2047 encoded word decode buffer

DTW-Thalion:
A long RFC2047 encoded word in a MIME header overflows the stack.

`-[GSMimeParser _decodeHeader]` decodes the data part of an encoded word (`=?charset?enc?data?=`) into a stack VLA:

```objc
if (tmp > src)
  {
    unsigned char buf[tmp - src + 1];   // tmp-src is the encoded-word length
    ...
    ptr = decodeWord(buf, src, tmp, encoding);
```

`tmp - src` is the length of the encoded data, which comes straight from the header being parsed, so an attacker controls the size of the VLA. A header carrying a long encoded word overflows the stack.

AddressSanitizer confirms it — a ~400KB `=?utf-8?B?AAAA…?=` decoded on a 256KB thread stack crashes (`SEGV` in `decodeWord`/`decodebase64` ← `_decodeHeader` ← `parseHeaders:remaining:` ← `parse:`).

This allocates the buffer on the heap instead, matching the bounded-stack approach used elsewhere for untrusted lengths.

### Test

`Tests/base/GSMime/encoded-word-stack.m` decodes a long encoded word on a thread with a small (256KB) stack, so the regression overflows the stack and aborts before the fix and passes after (Unix only; it skips elsewhere).

### DIFF
--- a/Source/Additions/GSMime.m
+++ b/Source/Additions/GSMime.m
@@ -3073,9 +3073,14 @@ unfold(const unsigned char *src, const unsigned char *end, BOOL *folded)
 	       */
 	      if (tmp > src)
 		{
-		  unsigned char	buf[tmp - src + 1];
 		  unsigned char	*ptr;
 
+		  /* The encoded word length is controlled by the (untrusted)
+		   * header data, so GS_BEGINITEMBUF keeps a short word on the
+		   * stack but moves a long one to the heap, avoiding both a
+		   * stack overflow and an unconditional heap allocation.
+		   */
+		  GS_BEGINITEMBUF(buf, tmp - src + 1, unsigned char)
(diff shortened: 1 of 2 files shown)


rfm (changes requested):
The analysis looks fine, but always allocating on the heap seems an inefficient solution.  I think the fix should use the GS_BEGINITEMBUF() and GS_ENDITEMBUF() macros from GSPrivate,h (already included).

DTW-Thalion:
Good idea - didn't think about those macros.

Switched to `GS_BEGINITEMBUF()`/`GS_ENDITEMBUF()` . Revalidated under AddressSanitizer: a ~400KB encoded word still decodes (via the heap path) without overflowing a 256KB thread stack, and the regression test passes.

PR #647, opened 2026-06-30
GSToUnicode: don't over-read a partial UTF-16/UTF-32 unit

DTW-Thalion:
`GSToUnicode` over-reads the source when decoding a UTF-16 or UTF-32 buffer whose length is not a whole number of code units.

The decoding loops read a full unit each iteration but only checked that one byte remained:

```objc
// UTF-16
while (spos < slen)
  { ptr[dpos++] = *(uint16_t*)(src + spos); spos += 2; }   // reads 2 bytes
// UTF-32
while (spos < slen)
  { uint32_t c = *(uint32_t*)(src + spos); ... spos += 4; } // reads 4 bytes
```

So an odd-length UTF-16 buffer (or a UTF-32 buffer not a multiple of 4) reads 1–3 bytes past the end. This is reachable from the public `-[NSString initWithData:encoding:]` — e.g. decoding a truncated UTF-16 file or network buffer.

AddressSanitizer: `heap-buffer-overflow ... in GSToUnicode` (Unicode.m:1285 for UTF-32; the UTF-16 loops at :1255/:1266 have the same issue).

The fix requires a whole unit to remain (`slen - spos >= 2` / `>= 4`); any trailing partial bytes are ignored.

### Test

`Tests/base/NSString/utf16-utf32-bounds.m` decodes a short non-unit-aligned buffer placed at the end of a page backed by a `PROT_NONE` guard page, so the over-read faults deterministically (Unix only; skips elsewhere), plus a well-formed UTF-16 buffer as a positive control. Before the fix the test aborts on the guard page; after, it passes 3/3.

### DIFF
--- a/Source/Additions/Unicode.m
+++ b/Source/Additions/Unicode.m
@@ -1250,7 +1250,7 @@ GSToUnicode(unichar **dst, unsigned int *size, const unsigned char *src,
 	if ((GS_WORDS_BIGENDIAN && NSUTF16LittleEndianStringEncoding == enc)
 	  || (!GS_WORDS_BIGENDIAN && NSUTF16BigEndianStringEncoding == enc))
 	  {
-	    while (spos < slen)
+	    while (slen - spos >= sizeof(uint16_t))
               {
                 uint16_t	c = *(uint16_t*)(src + spos);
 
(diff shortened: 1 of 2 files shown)


rfm (commented):
One tiny thing, from the point of view of having self-documenting code, it would be nice to use sizeof(uint16_t) and sizeof(uint32_t) rather than 2 or 4 to make the point of the boundary check explicit.

DTW-Thalion:
Done — the three boundary checks now use `sizeof(uint16_t)` / `sizeof(uint32_t)` instead of the literals.

PR #649, opened 2026-06-30
NSKeyedUnarchiver: range-check the CF$UID object reference before indexing _objMap

DTW-Thalion:
## Summary

`-[NSKeyedUnarchiver _decodeObject:]` uses the object-reference index decoded from a `CF$UID` in the archive to index the `_objMap` `GSIArray` **before** the bounds-checked `-[_objects objectAtIndex:]` that follows:

```objc
obj = GSIArrayItemAtIndex(_objMap, index).obj;   /* index is an untrusted CF$UID */
if (obj != nil) { ... return obj; }
...
obj = [_objects objectAtIndex: index];           /* this one is range-checked */
```

`GSIArrayItemAtIndex` is not bounds checked in release builds — its only check is an assertion that `NS_BLOCK_ASSERTIONS` compiles out. So an archive whose root `CF$UID` is out of range reads past the end of `_objMap`: a heap buffer overflow when unarchiving an untrusted archive. The intended rejection (`objectAtIndex:` raising `NSRangeException`) is simply reached too late.

The fix range-checks `index` against `GSIArrayCount(_objMap)` up front and rejects an out-of-range reference with `NSRangeException`, matching the rejection `objectAtIndex:` already performs.

- **Positive + negative control** (in-tree `gnustep-tests`, assertions on): patched, **2/2 pass** — a normal object still round-trips through the keyed archiver, and an archive whose root `CF$UID` is one past the object table is rejected with `NSRangeException`. Against the unpatched lib the rejection case **fails** (`GSIArrayItemAtIndex`'s assertion raises `NSInternalInconsistencyException` instead).
- **Memory safety** (ASan, built with `NS_BLOCK_ASSERTIONS` so the `GSIArray` assertion is compiled out): the unpatched `-_decodeObject:` reads past `_objMap` — `AddressSanitizer: heap-buffer-overflow, READ of size 8` just past the map allocation. The patched build rejects the reference before the read.

A new regression test is added at `Tests/base/NSKeyedArchiver/uidbounds.m`.

### DIFF
--- a/Source/NSKeyedUnarchiver.m
+++ b/Source/NSKeyedUnarchiver.m
@@ -164,6 +164,17 @@ static NSMapTable	*globalClassMap = 0;
    * If the referenced object is already in _objMap
    * we simply return it (the object at index 0 maps to nil)
    */
+  /* The index comes from an (untrusted) CF$UID in the archive, and GSIArray
+   * is not bounds checked in release builds, so reject an out-of-range
+   * reference (as -[_objects objectAtIndex:] does below) rather than reading
+   * past the end of the map.
+   */
+  if (index >= GSIArrayCount(_objMap))
+    {
+      [NSException raise: NSRangeException
+		  format: @"[%@ -%@] archive object index out of range",
+	NSStringFromClass([self class]), NSStringFromSelector(_cmd)];
+    }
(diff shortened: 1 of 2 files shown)


rfm (approved):
Looks good, thanks.

PR #650, opened 2026-06-30
GSMime: avoid a stack VLA when scanning a long quoted header token

DTW-Thalion:
## Summary

`-[GSMimeParser scanToken:]` parses a header token. Its quoted-string branch copies the token into a **stack VLA sized from the quoted-string length**:

```objc
unichar buf[length];                 /* length = quoted-string content length */
...
[string getCharacters: buf range: NSMakeRange(start, length)];
```

A quoted header value is attacker-controlled, so a long quoted value overflows the stack. This is reachable from untrusted input through:

- **HTTP auth** — `+[GSHTTPAuthentication protectionSpaceForAuthentication:requestURL:]` (and `-authorizationForAuthentication:method:path:`) scan a malicious server's `WWW-Authenticate` value, e.g. `Digest realm="<very long>"`.
- **MIME header parameters** — any header with a long quoted parameter value (`Content-Type: …; name="<very long>"`, etc.) goes through `scanToken:` during a normal `-parse:`.

The fix uses `GS_BEGINITEMBUF` / `GS_ENDITEMBUF` (as in `-_decodeHeader`), keeping the common short token on the stack and moving only a large one to the heap.

- Reached through the public auth entry `+[GSHTTPAuthentication protectionSpaceForAuthentication:requestURL:]` with a 400KB quoted realm on a 256KB-stack thread: the unpatched build crashes (**AddressSanitizer: SEGV, stack overflow** on the worker thread); the patched build returns normally.
- Regression test (in-tree `gnustep-tests`, a 256KB-stack worker scanning a 400000-character quoted token): patched **1/1 pass**; against the unpatched lib the test binary **crashes (stack overflow) before completing**.

A new regression test is added at `Tests/base/GSMime/quoted-token-stack.m`.

### DIFF
--- a/Source/Additions/GSMime.m
+++ b/Source/Additions/GSMime.m
@@ -2253,10 +2253,17 @@ NSDebugMLLog(@"GSMime", @"Header parsed - %@", info);
 	}
       else
 	{
-	  unichar	buf[length];
-	  unichar	*src = buf;
-	  unichar	*dst = buf;
+	  unichar	*src;
+	  unichar	*dst;
+	  NSString	*result;
 
+	  /* The quoted string length is attacker-controlled (it comes from an
+	   * untrusted header value), so use GS_BEGINITEMBUF rather than a stack
+	   * VLA to avoid a stack overflow on a very long quoted string.
+	   */
+	  GS_BEGINITEMBUF(buf, length, unichar)
+	  src = buf;
+	  dst = buf;
 	  [string getCharacters: buf range: NSMakeRange(start, length)];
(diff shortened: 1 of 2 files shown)


rfm (approved):
I'm going to merge this since it's basically as efficient than the old code,  but this is actually a case where using the heap might actually be *more* efficient than the stack, because there exists an API to initialise a string directly from a buffer on the heap, where the string takes ownership of the buffer, so it's possible this could be optimised a tiny bit.  Probably not worth putting effort into testing that though.

PR #655, opened 2026-07-03
Add NSSet and NSMutableSet operation tests

DTW-Thalion:
This adds a coverage test for `NSSet` and `NSMutableSet` semantics. The existing `Tests/base/NSSet/basic.m` only checks empty/single-element creation and the generic protocol helpers (NSObject/NSCoding/NSCopying); it doesn't exercise the set operations themselves.

`Tests/base/NSSet/operations.m` covers:

- **Creation & de-duplication** — `+set`, `+setWithObject:`, `+setWithObjects:`, `+setWithArray:`, `+setWithSet:`, and that duplicates are not stored
- **Membership** — `containsObject:`, `member:` (returns the stored equal object / nil), `anyObject` (incl. empty set), `allObjects`, `count`
- **Predicates** — `isEqualToSet:` (order-independent), `isSubsetOfSet:` (incl. the empty set), `intersectsSet:` (incl. disjoint and empty)
- **Derived sets** — `setByAddingObject:` (and that the original is unchanged), `setByAddingObjectsFromArray:`, `setByAddingObjectsFromSet:`
- **NSMutableSet mutation** — `addObject:` (incl. duplicate no-op), `addObjectsFromArray:`, `removeObject:` (incl. non-member no-op), `removeAllObjects`
- **NSMutableSet set algebra** — `unionSet:`, `intersectSet:`, `minusSet:`, `setSet:`

This is test-only (no source changes) and portable — it deliberately avoids the block-based methods (`objectsPassingTest:`, block enumeration) so it builds regardless of block support.

37 assertions, all passing:

```
$ gnustep-tests base/NSSet/operations.m
     37 Passed tests
All OK!
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2026-07-03 Todd White <todd.white@thalion.global>
+
+	* Tests/base/NSSet/operations.m:
+	Add tests for NSSet and NSMutableSet: creation and de-duplication,
+	membership (containsObject:, member:, anyObject, allObjects), the set
+	predicates (isEqualToSet:, isSubsetOfSet:, intersectsSet:), the
+	setByAdding... derived-set methods, and NSMutableSet mutation and set
+	algebra (unionSet:, intersectSet:, minusSet:, setSet:).
+
 2026-07-01 Richard Frith-Macdonald <rfm@gnu.org>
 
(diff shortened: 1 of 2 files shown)


rfm:
Merged.  Thanks for all these testcases.

PR #685, opened 2026-07-06
Create NSProgress userInfo dictionary if needed

triplef:
Fixes `-setUserInfoObject:forKey:` not working if progress was initialized with nil object, and matches macOS behavior where calling `-userInfo` always returns an object.


rfm (approved):
Yep, looks good.  Thanks.

PR #689, opened 2026-07-06
NSInvocation: support structures containing bitfields

### LINKED ISSUE #290: `cifframe_type` does not support bitfields (`_C_BFLD`)
I'm seeing libs-base hitting the `NSCAssert(0, @"Unknown type in sig");` assertion in `cifframe_type`.

In this case, the type being parsed is `b0I1b1I1b2I1b3I1b4I1b5I1b6I1b7I1b8I1b9I1`, which seems to correspond to the [`rflags`](https://github.com/gnustep/libs-gui/blob/master/Headers/AppKit/NSView.h#L157-L170) struct used by `NSView`.

I hit this particular use case because as part of an effort to get libs-gui and libs-back to compile on msvc, [`rflags` was being accessed via a function call, rather than via the instance variable](https://github.com/qmfrederik/tools-windows-msvc/blob/features/libs-back/patches/gnustep-back-win32-fixes.patch#L71-L72).

The easy workaround appears to be to msvc port to use methods to access values of that struct, rather than the entire struct.  But thought it may be worth documenting this issue anyway.

DTW-Thalion:
Fixes #290.

Passing or returning a structure by value through an `NSInvocation` aborts in
`cifframe_type` with `Unknown type in sig` when the structure contains a
bitfield (the `_C_BFLD`, `b<offset><type><width>` encoding) — the `rflags`
case from the issue.

Fixing the abort exposed a second problem: `NSMethodSignature` sizes such a
structure as zero.  A bitfield reports a byte size of zero on its own (it
shares a storage unit with its neighbours), so summing the member sizes in
`next_arg` gives the whole structure a size of zero; `getReturnValue:` then
copies nothing and the returned structure comes back all-zero.

Both are addressed here:

* `cifframe_type` handles `_C_BFLD` by coalescing each run of consecutive
  bitfields into the storage units of their underlying type (one ffi element
  per unit), so the enclosing structure gets the correct libffi layout.
* `next_arg` takes the size of a bitfield-containing structure from the
  runtime (`objc_sizeof_type`) instead of summing the members.

### Test

`Tests/base/NSInvocation/bitfield.m` passes and returns three bitfield
structures through an invocation and checks the round-tripped field values: a
3/5/7-bit struct, a run of single-bit fields as in NSView's `rflags`, and a
struct wide enough to occupy two storage units (exercising the multi-unit
path).

Verified on the current tree:

* Before: the test aborts with the assertion from this issue.
* After: the values round-trip correctly, and the NSInvocation (75),
  NSMethodSignature (46) and NSValue (47) suites all pass.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,20 @@
 	* Headers/Foundation/NSKeyedArchiver.h: Add the _allowedClasses
 	instance variable.
 	* Tests/base/NSKeyedArchiver/secure_coding.m: New test.
+	* Source/cifframe.m (cifframe_type): Handle the _C_BFLD type encoding.
+	Passing or returning a structure that contains a bitfield through an
+	NSInvocation aborted with "Unknown type in sig".  Coalesce each run of
+	bitfields into the storage units of their underlying type so the
+	enclosing structure gets the correct libffi layout.
+	* Source/NSMethodSignature.m (next_arg): Take the size of a structure
+	that contains bitfields from the runtime.  Bitfields share a storage
+	unit and report a byte size of zero individually, so summing the member
+	sizes gave such a structure a size of zero and its return value was lost.
+	Work out a bitfield's alignment from its underlying type rather than
+	from objc_alignof_type, whose gnu runtime implementation aborts on a
+	bitfield encoding.
+	* Tests/base/NSInvocation/bitfield.m: New test.
+
 2026-07-05 Todd White <todd.white@thalion.global>
 
 	* Source/NSScanner.m (-[NSScanner scanDecimal:]): Implement; it
--- a/Source/NSMethodSignature.m
+++ b/Source/NSMethodSignature.m
@@ -266,6 +266,7 @@ next_arg(const char *typePtr, NSArgumentInfo *info, char *outTypes)
 	  unsigned int def_align = objc_alignof_type(typePtr-1);
 	  unsigned int acc_align = def_align;
 	  const char	*ptr = typePtr;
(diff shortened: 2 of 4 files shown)


rfm (changes requested):
The testcase here uses the reserved words in and out as variables.  Some compilers may accept that, but it's not portable so needs changing.  Apart from the it looks good.

DTW-Thalion:
Done — renamed the `in`/`out` locals and parameters to `input`/`output` throughout the test.

rfm (changes requested):
There's another portability issue: the implementation of objc_alignof_type in the gnu runtime does not handle bitfields, so the call in next_arg() in NSMethodSignature.m aborts with a message 'unknown type':  we need to calculate the size/alignment of the bitfield directly rather than assuming that the runtime functions will do it.

DTW-Thalion:
> There's another portability issue: the implementation of objc_alignof_type in the gnu runtime does not handle bitfields, so the call in next_arg() in NSMethodSignature.m aborts with a message 'unknown type': we need to calculate the size/alignment of the bitfield directly rather than assuming that the runtime functions will do it.

Will  look at that latter today.

DTW-Thalion:
Done. next_arg() now works out a bitfield's alignment from its underlying type rather than from objc_alignof_type, which aborts on the gnu runtime's bitfield encoding. The field carries no independent byte size, so the structure's size still comes from the single struct-level runtime call.

rfm (approved):
Great, thanks

PR #690, opened 2026-07-07
New key-value observing (KVO) implementation on all runtimes

DTW-Thalion:
Add a per-class swizzling backend, using an NSKVONotifying_<Class> subclass and object_setClass, for runtimes that
lack that extension, and build the new KVO on every runtime.

--disable-newkvo keeps the old implementation.

The GNU runtime needed portability work.  The observer classes get explicit instance variables, as gcc does not synthesize them from the property declarations, and a recursive gs_mutex replaces @synchronized, which cannot be used on Windows where the GNU runtime does not enable Objective-C exceptions.  

Performance is on par with the previous KVO implementation on the libobjc2 hot path.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -31,8 +31,19 @@
 	prototype takes the address as char*, so it is cast; it is still the
 	binary address rather than an ascii string.
 
-2026-07-07  Todd White <todd.white@thalion.global>
+2026-07-08  Todd White <todd.white@thalion.global>
 
+	* Source/NSKVOSwizzling.m: On runtimes without libobjc2 per-object
+	method addition, make an observed object notify for arbitrary-key
+	mutation, as NSMutableDictionary does, by overriding setObject:forKey:
+	and removeObjectForKey: on the notifying subclass.  This was done only
+	for the libobjc2 backend, so key-value observing of a mutable
+	dictionary fired no notifications on the GNU runtime.
+	* Tests/base/NSKVOSupport/general.m,
+	Tests/base/NSKVOSupport/newoldvalues.m: Remove the FLAKY_ON_GCC markers
+	that made the affected cases hopeful on the GNU runtime; they now pass.
+
+2026-07-07  Todd White <todd.white@thalion.global>
 	* Tests/base/NSKVOSupport/kvoToMany.m: Rework to avoid Objective-C
 	blocks and other constructs unavailable with the GNU runtime
 	(property auto-synthesis, the strong attribute, object subscripting),
--- a/ChangeLog
+++ b/ChangeLog
@@ -33,6 +33,40 @@
 
 2026-07-07  Todd White <todd.white@thalion.global>
 
+	* configure.ac: Build the new key-value observing implementation on
+	every Objective-C runtime rather than only the next-generation one.
+	Add --disable-newkvo to fall back to the old implementation.
+	* configure: Regenerate.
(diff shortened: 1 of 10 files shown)


rfm:
I think before we can look at this in detail, we need to have the testcases rewritten to work with objc-1 so that we can reliably compare before/after changes to the kvo implementation.  Ideally I guess we also want to be confident that our testcases cover all relevant behaviors.
In particular, the current kvoToMany.m currently uses blocks and needs to be reworked to avoid that.

DTW-Thalion:
I can look at the test cases today at some point.  Though it is worth noting that everything that is there currently is objc-1 clean except for KvoToMany.m - so that is where most of the effort will be. I will exercise that. Also, a lot of the tests are FLAKY_ON_GCC which hides failures. 

I'll start with deblocking kvoToMany and run the whole suite on the new KVO to confirm the collection tests pass.

DTW-Thalion:
Done. kvoToMany.m no longer uses blocks, nor the other clang-only constructs that surfaced once the objc-2 guard was removed (property auto-synthesis, the strong attribute, object subscripting). It builds and runs on the GNU runtime now and passes the same 203 assertions on both runtimes, so the before/after comparison works on objc-1.

I also rebased onto master, so the PR is no longer conflicting.

The rest of Tests/base/NSKVOSupport already builds on objc-1, but general.m and newoldvalues.m wrap some cases in FLAKY_ON_GCC (testHopeful on the GNU runtime), so those results are tolerated rather than asserted. Now that the new implementation runs on objc-1, I will remove the markers so the cases assert properly and see where that takes us.

DTW-Thalion:
Removing the markers which surfaced three failures on the GNU runtime, I forgot to cut and paste a section from the new KVO libobjc2 path to the GCC path - all key-value observing of an NSMutableDictionary: setObject:forKey: and removeObjectForKey: were not firing notifications. In the portable swizzling backend for GCC I had installed notifying setters for simple keys and to-many collections but not for arbitrary-key mutation, which I had done in the libobjc2 backend. I added that, removed the FLAKY_ON_GCC markers from general.m and newoldvalues.m, and the full NSKVOSupport suite now passes identically on both runtimes (506 tests, no hopeful cases).

rfm (approved):
While this is hard to review because it is a very big patch,  it's also rather straightforward in removing dependencies rather than making design/algorithmic changes, and its easy to see that it is generally safe/correct.  The test coverage provides a lot of confidence too.
So I'm happy that this is good to merge.  Thanks.

rmottola:
This makes compilation fail on older gcc, atomic is not defined.
(Here on Solaris with old gcc 4.9). Can we skip this with some kludge or did a config test miss? @rfm 

```
In file included from NSUserDefaults.m:57:0:
NSKVOInternal.h:123:18: error: unknown property attribute before ',' token
 @property (atomic, retain) NSMutableDictionary *pendingChange;
                  ^
NSKVOInternal.h:123:1: warning: object property 'pendingChange' has no 'assign', 'retain' or 'copy' attribute; assuming 'assign'
 @property (atomic, retain) NSMutableDictionary *pendingChange;
```

rfm:
We just didn't test with older GCCs.  I pushed the trivial fix: remove the keyword since atomic is the default behaviour.

rfm:
I have tried to keep properties out of the codebase  but I think we need to document that gnustep now requires gcc-4.6 or later (support for properties).
It seems reasonable since that was released 15 years ago.

Issue #552, opened 2025-10-20
support gethostbyname_r 5 arguments version


rmottola:
currently we support gethostbyname_r with 6 arguments or fallback to standard.
Some systems (e.g. solaris) have gethostbyname_r with 5 arguments (no result apparently) that would still be a better option.

We could improve configure checks as asterisk and support it.

DTW-Thalion:
Opened #692 for this. configure detects the 5-argument form and NSHost calls it. I validated it against a stub with the Solaris signature but have no Solaris system to test the actual resolution on. @rmottola @rfm - either of you running Solaris?

rfm:
Merged the change.  Please could you test it on your Solaris system Riccardo?

rmottola:
I don't think you properly "merged" it, @rfm ? I don't see the change in master.

I downloaded the diff of the PR of @DTW-Thalion  and tested on solaris 10 on UltraSPARC

I see 
`checking for gethostbyaddr_r... yes`

then I see in `config.h`
```
/* Number of arguments taken by gethostbyname_r (5 or 6). */
#define GS_GETHOSTBYNAME_R_ARGS 5
```

This looks fine.

Actual build fails though:
```
 Compiling file NSHost.m ...
NSHost.m: In function '-[NSHost(Private) _addHostInfo:withNames:addresses:]':
NSHost.m:431:35: warning: cast increases required alignment of target type [-Wcast-align]
        struct sockaddr_in *ipv4 = (struct sockaddr_in *)(tmp->ai_addr);
                                   ^
NSHost.m:436:36: warning: cast increases required alignment of target type [-Wcast-align]
        struct sockaddr_in6 *ipv6 = (struct sockaddr_in6 *)tmp->ai_addr;
                                    ^
NSHost.m: In function '-[NSHost(Private) _addHostAddress:withNames:addresses:]':
NSHost.m:509:32: warning: passing argument 1 of 'gethostbyaddr_r' from incompatible pointer type
       if (0 == gethostbyaddr_r(&ip_addr, sizeof(ip_addr), AF_INET,
                                ^
In file included from NSHost.m:82:0:
/usr/include/netdb.h:246:17: note: expected 'const char *' but argument is of type 'struct in_addr *'
 struct hostent *gethostbyaddr_r
                 ^
NSHost.m:510:32: warning: passing argument 7 of 'gethostbyaddr_r' from incompatible pointer type
[...]

PR #692, opened 2026-07-08
NSHost: support the 5-argument gethostbyname_r

DTW-Thalion:
Only the 6-argument glibc form of `gethostbyname_r` was detected and used. Solaris and some other systems provide a 5-argument form that returns the `struct hostent*` directly rather than through a result parameter.

`configure` now checks for the 6-argument form and, failing that, the 5-argument form, recording the count in `GS_GETHOSTBYNAME_R_ARGS`. `-[NSHost(Private) _addHostName:withNames:addresses:]` calls the matching form.

The 6-argument path is unchanged and verified on Linux (Tests/base/NSHost passes, 88 tests). I have no Solaris system to exercise the 5-argument path on, so I validated it against a stub with the Solaris signature: the 6-argument configure test fails to compile against a 5-argument declaration while the 5-argument test links, and the NSHost call site compiles and returns the entry correctly when built with the 5-argument form.

Implements #552.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,14 @@
+2026-07-07  Todd White <todd.white@thalion.global>
+
+	* configure.ac: Detect the 5-argument (Solaris) form of
+	gethostbyname_r in addition to the 6-argument (glibc) form, and
+	record the argument count in GS_GETHOSTBYNAME_R_ARGS.
+	* configure: Regenerate.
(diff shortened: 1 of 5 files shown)


rfm (approved):
As Solaris is a fringe platform nowadays, I think we can simply merge this and fix (if necessary) once Riccardo can test it.

PR #695, opened 2026-07-08
NSNumber: hash by numeric value rather than truncating

DTW-Thalion:
`-[NSNumber hash]` cast doubleValue to an integer, so every value sharing an integral part collided: 1.5 and 1.7 hashed alike.

This reduces doubleValue modulo a Mersenne prime, after the manner of CPython, so the fractional part contributes and an integral value still hashes to that integer.

The hash stays a function of doubleValue because that is what `-compare:` and `-isEqual:` use. NSNumber equality promotes to double, so a value held as int, float or double compares equal across those types and must hash equally. Hashing the raw storage, as #535 first tried, breaks that: the bit patterns of an int and a double of the same value differ, so equal numbers get different hashes and fail as dictionary keys.

One consequence worth noting: equality is not transitive across the precision boundary. A long long above 2^53 compares equal to the double it rounds to, so its hash must be that of the double, and distinct large integers can collide. That follows `-compare:` and cannot be avoided without changing comparison.

`Tests/base/NSNumber/NSNumber_hash.m` covers the `-hash`/`-isEqual:` contract, fractional scattering and non-finite values. 862 tests pass under both the GNU and libobjc2 runtimes.

This is an option for #535, following the doubleValue-based approach discussed there.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,16 @@
 2026-07-08  Todd White <todd.white@thalion.global>
 
+	* Source/NSNumber.m (GSHashDouble, -[NSNumber hash]): Derive the hash
+	from -doubleValue by reducing the value modulo a Mersenne prime, so
+	that equal numbers held in different types hash equally (as -isEqual:
+	requires) and fractional values no longer collide on their integral
+	part.
+	* Tests/base/NSNumber/NSNumber_hash.m: New test for the -hash /
+	-isEqual: contract, for fractional scattering and for non-finite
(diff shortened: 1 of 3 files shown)


rfm (approved):
Great,  Given that this is the approach I found looked best when I searched for alternative hashing solutions, I have nothing to complain about here.  I know Hugo was going to come back after reviewing the literature, but this is a good solution (and if he finds a better one we can always change later).

Issue #482, opened 2025-01-05
NSURLSession implementation needs to be portable between compilers


rfm:
At the moment, the impl,emenation is clang-specific.
We need to rewrite avoiding the use of blocks.

DTW-Thalion:
Some measurements on what the current state is, since the shape of the
rewrite depends on which part is meant.

From the config.log of the Ubuntu x64 GCC job on a recent run:
HAVE_BLOCKS is 0, HAVE_LIBCURL is 1, and GS_HAVE_NSURLSESSION is 0. So
NSURLSession is not built there at all. Nothing fails to compile, because
the whole of Headers/Foundation/NSURLSession.h sits inside
"#if GS_HAVE_NSURLSESSION": the class is simply absent.

The tests still run on that job and assert nothing.
base/NSURLSession/simpleTaskTests.m completes in 0.049s with no test
lines, so the file reports as completed while covering nothing.

Where the blocks are: the public header uses block types in 9 places,
including the completionHandler arguments of
-dataTaskWithRequest:completionHandler: and
-getTasksWithCompletionHandler:. Those are the API as Apple defines it,
so dropping them means the class no longer offers that API. The internal
use is separate and is the part a rewrite can reach: about 17 block
literals in NSURLSessionTask.m and 8 in NSURLSession.m.

There is also a second gate. configure sets the default from
OBJC_RUNTIME_LIB being ng as well as from HAVE_BLOCKS, and raises
"Missing ng runtime (needed for NSURLSession)" if the option is asked for
with any other runtime, so removing the block use alone would not make
the class available under the gnu runtime.

Which of those is the target here: the internal block use, or the class
being available where the completion handler API cannot be expressed?

cc @rfm

rfm:
It's not quite that simple.

Having the class avoid needing clang for internal use is obviously a good staging point, but it's pretty worthless if you then have to have clang to use it, so it's not the final objective.

My understanding is that most of the functionality is available without blocks (using the delegate methods).  So getting to s stage where we can build using GCC, and test all the API that can be tested from GCC is a good second stage.

A final stage would be to design/implement alternative methods where existing delegate methods don't suffice, and deprecate the parts if the API using blocks, pointing users to the new non-,block methods.  That gets us to the modern model based an a solid foundation (libcurl) with a portable API.

DTW-Thalion:
OK, sounds good - I'll see if I can figure out a strategy for this.

DTW-Thalion:
[...]

PR #696, opened 2026-07-08
NSURLSession: drive libcurl from a run loop work thread instead of libdispatch

DTW-Thalion:
Addresses the libdispatch dependency in NSURLSession tracked by #488. NSURLSessionStreamTask and downloadTaskWithResumeData: remain unimplemented as before; this change is limited to the event engine. Does not address portability across compilers per #482... yet.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -21,6 +21,32 @@
 
 2026-07-08  Todd White <todd.white@thalion.global>
 
+	* Source/NSURLSession.m, Source/NSURLSessionTask.m,
+	Source/NSURLSessionPrivate.h: Drive the libcurl multi handle from a
+	dedicated work thread running an NSRunLoop rather than libdispatch, so
+	the class no longer depends on libdispatch.  The serial work queue
+	becomes an NSThread, the timer source an NSTimer, and the per-socket
+	read and write sources become run loop descriptor watchers (ET_RDESC
+	and ET_WDESC on unix, a WSAEVENT registered as ET_HANDLE on Windows).
+	A socket watcher is now added or removed only on an actual direction
+	change rather than torn down and recreated.  dispatch_once is replaced
+	by a static gs_mutex.
+	* configure.ac, Source/GNUmakefile: Build NSURLSession whenever the ng
+	runtime, blocks and libcurl are present, no longer requiring libdispatch,
+	so the class is available (and its tests run) on systems without
+	libdispatch installed.
+	* Tests/base/NSURLSession/Helpers/HTTPServer.m,
+	Tests/base/NSURLSession/Helpers/GNUmakefile: Rewrite the test HTTP
+	server to serve each connection on its own thread rather than with
+	libdispatch queues and sources, so the tests no longer depend on
+	libdispatch either.
+	* Tests/base/NSURLSession/simpleTaskTests.m: Re-enable the suite, which
+	was disabled (except under MSVC, where pre-existing transfer failures
+	are tracked separately), and add tests for invalidateAndCancel and for
+	deallocating a session without invalidating it.
+
+2026-07-08  Todd White <todd.white@thalion.global>
+
 	* Source/NSNumber.m (GSHashDouble, -[NSNumber hash]): Derive the hash
 	from -doubleValue by reducing the value modulo a Mersenne prime, so
 	that equal numbers held in different types hash equally (as -isEqual:
--- a/Headers/GNUstepBase/config.h.in
+++ b/Headers/GNUstepBase/config.h.in
@@ -262,10 +262,6 @@
 /* Define to 1 if you have the <dispatch/private.h> header file. */
 #undef HAVE_DISPATCH_PRIVATE_H
 
-/* Define to 1 if you have the `dispatch_queue_create_with_target' function.
-   */
-#undef HAVE_DISPATCH_QUEUE_CREATE_WITH_TARGET
-
 /* Define to 1 if you have the `dladdr' function. */
 #undef HAVE_DLADDR
 
--- a/Source/GNUmakefile
+++ b/Source/GNUmakefile
@@ -386,15 +386,11 @@ ifneq ($(GNUSTEP_TARGET_OS), mingw32)
   endif
 endif
 
-ifeq ($(HAVE_BLOCKS), 1)
-ifeq ($(GNUSTEP_BASE_HAVE_LIBDISPATCH), 1)
-ifeq ($(GNUSTEP_BASE_HAVE_LIBCURL), 1)
+ifeq ($(GNUSTEP_BASE_WITH_NSURLSESSION), 1)
   BASE_MFILES += \
 	NSURLSession.m \
 	NSURLSessionTask.m \
 	NSURLSessionConfiguration.m
-endif
-endif
 endif
 
 ifeq ($(GNUSTEP_BASE_HAVE_MDNS), 1)
--- a/Source/NSURLSession.m
+++ b/Source/NSURLSession.m
@@ -466,6 +466,23 @@ static NSURLSession * sharedSession = nil;
   curl_multi_remove_handle(_multiHandle, easy);
 }
 
+/* Called on the work thread when the delegate refuses a redirect.  The
+ * intercepted transfer's completion was held back in -_checkForCompletion, so
+ * finish the task here with a cancellation. */
+- (void) _cancelRedirectForTask: (NSURLSessionTask *)task
+{
+  curl_multi_remove_handle(_multiHandle, [task _easyHandle]);
+
+  /* -_transferFinishedWithCode: may release the last reference to the
+   * session, so keep both alive across the call. */
+  RETAIN(self);
+  RETAIN(task);
+  [_tasks removeObject: task];
+  [task _transferFinishedWithCode: CURLE_ABORTED_BY_CALLBACK];
+  RELEASE(task);
+  RELEASE(self);
+}
+
 /* Called on the work thread from libcurl's timer_callback.  Schedules a
  * one-shot timer on the work thread run loop; libcurl re-arms it as needed.
  */
@@ -801,6 +818,20 @@ static NSURLSession * sharedSession = nil;
             task,
             eff_url,
             curl_easy_strerror(res));
+
+          /* With CURLOPT_FOLLOWLOCATION disabled libcurl reports an
+           * intercepted 3xx response as a completed transfer.  When the task
+           * is being redirected the easy handle is about to be re-added for
+           * the new location (or cancelled by the delegate), so hold back the
+           * completion instead of delivering it for the intermediate leg. */
+          if ([task _redirectInProgress])
+            {
+              NSDebugMLLog(
+                GS_NSURLSESSION_DEBUG_KEY,
+                @"Holding back completion for redirecting task %@",
+                task);
+              continue;
+            }
 
           curl_multi_remove_handle(_multiHandle, easyHandle);
 
--- a/Source/NSURLSession.m
+++ b/Source/NSURLSession.m
@@ -439,12 +439,27 @@ static NSURLSession * sharedSession = nil;
       [task _easyHandle],
       multiHandle,
       code);
+
+    /* Kick the transfer off now rather than waiting for the timer callback
+     * (see -_addHandle:). */
+    curl_multi_socket_action(multiHandle, CURL_SOCKET_TIMEOUT, 0,
+      &_stillRunning);
+    [self _checkForCompletion];
   }];
 }
 
 - (void) _addHandle: (CURL *)easy
 {
   curl_multi_add_handle(_multiHandle, easy);
+
+  /* Kick the added transfer off now rather than waiting for libcurl to fire
+   * the timer callback.  Relying on the timer alone races with the run loop,
+   * which shows up most on a handle that is re-added after a redirect: the
+   * transfer can stall until an unrelated event drives the multi handle.
+   * See https://curl.se/libcurl/c/curl_multi_socket_action.html . */
+  curl_multi_socket_action(_multiHandle, CURL_SOCKET_TIMEOUT, 0,
+    &_stillRunning);
+  [self _checkForCompletion];
 }
 - (void) _removeHandle: (CURL *)easy
 {
--- a/Source/NSURLSession.m
+++ b/Source/NSURLSession.m
@@ -31,7 +31,13 @@
 #import "NSURLSessionTaskPrivate.h"
 #import "Foundation/NSString.h"
 #import "Foundation/NSArray.h"
+#import "Foundation/NSDate.h"
+#import "Foundation/NSMapTable.h"
+#import "Foundation/NSPort.h"
+#import "Foundation/NSRunLoop.h"
 #import "Foundation/NSStream.h"
+#import "Foundation/NSThread.h"
+#import "Foundation/NSTimer.h"
 #import "Foundation/NSUserDefaults.h"
 #import "Foundation/NSBundle.h"
 #import "Foundation/NSData.h"
@@ -39,7 +45,6 @@
 #import "GNUstepBase/NSDebug+GNUstepBase.h"  /* For NSDebugMLLog */
 #import "GNUstepBase/NSObject+GNUstepBase.h" /* For -notImplemented */
 #import "GSPThread.h"                        /* For nextSessionIdentifier() */
-#import "GSDispatch.h"                       /* For dispatch compatibility */
 
 NSString * GS_NSURLSESSION_DEBUG_KEY = @"NSURLSession";
 
@@ -125,22 +130,80 @@ socket_callback(CURL * easy,           /* easy handle */
     }
 } /* socket_callback */
 
+#pragma mark - Work thread trampoline
+
+/* Runs the run loop for an NSURLSession's work thread.  It deliberately
+ * holds no reference to the session, so that the session's lifetime is not
+ * extended by the running thread; -[NSURLSession dealloc] stops and joins
+ * the thread before releasing anything the thread might touch.
+ */
+@interface GSURLSessionWorkThread : NSObject
+{
+@public
+  NSThread * thread;
+  NSPort * port;
+  BOOL shouldExit;
+}
+- (void) run;
+- (void) stop;
+@end
+
+@implementation GSURLSessionWorkThread
+- (void) run
+{
+  NSAutoreleasePool * pool = [NSAutoreleasePool new];
+  NSRunLoop * rl = [NSRunLoop currentRunLoop];
+
+  [rl addPort: port forMode: NSDefaultRunLoopMode];
+  while (!shouldExit)
+    {
+      NSAutoreleasePool * inner = [NSAutoreleasePool new];
+
+      [rl runMode: NSDefaultRunLoopMode beforeDate: [NSDate distantFuture]];
+      [inner release];
+    }
+  [rl removePort: port forMode: NSDefaultRunLoopMode];
+  [pool release];
+}
+
+- (void) stop
+{
+  shouldExit = YES;
+}
+@end
+
 #pragma mark - NSURLSession Implementation
+
+/* The session acts as its own run loop watcher for the sockets libcurl
+ * asks us to monitor. */
+@interface NSURLSession () <RunLoopEvents>
+@end
 
 @implementation NSURLSession
 {
   /* The libcurl multi handle associated with this session.
    * We use the curl_multi_socket_action API as we utilise our
-   * own event-handling system based on libdispatch.
+   * own event-handling system integrated with the work thread run loop.
    *
    * Event creation and deletion is driven by the various callbacks
    * registered during initialisation of the multi handle.
    */
   CURLM * _multiHandle;
-  /* A serial work queue for timer and socket sources
-   * created on libcurl's behalf.
+  /* Drives a dedicated thread running an NSRunLoop.  All libcurl multi
+   * handle activity (adding handles, socket events and the timer) happens on
+   * that thread, which serialises access in place of a dispatch queue and
+   * keeps GNUstep free of a libdispatch dependency.  The helper holds the
+   * session unretained so that it does not keep the session alive.
    */
-  dispatch_queue_t _workQueue;
+  GSURLSessionWorkThread * _workHelper;
+
+#if	defined(_WIN32)
+  /* Maps a registered WSAEVENT back to its socket, since the run loop only
+   * hands the event handle back to -receivedEvent:type:extra:forMode:.
+   */
+  NSMapTable * _socketForEvent;
+#endif
+
   /* This timer is driven by libcurl and used by
    * libcurl's multi API.
    *
@@ -150,13 +213,9 @@ socket_callback(CURL * easy,           /* easy handle */
    *
    * See https://curl.se/libcurl/c/CURLMOPT_TIMERFUNCTION.html
    * and https://curl.se/libcurl/c/curl_multi_socket_action.html
-   * respectively.
+   * respectively.  It is scheduled on the work thread run loop.
    */
-  dispatch_source_t _timer;
-
-  /* The timer may be suspended upon request by libcurl.
-   */
-  BOOL _isTimerSuspended;
+  NSTimer * _timer;
 
   /* Only set when session originates from +[NSURLSession sharedSession] */
   BOOL _isSharedSession;
@@ -169,7 +228,7 @@ socket_callback(CURL * easy,           /* easy handle */
    */
   int _stillRunning;
 
-  /* List of active tasks. Access is synchronised via the _workQueue.
+  /* List of active tasks. Access is synchronised via the work thread.
    */
   NSMutableArray<NSURLSessionTask *> * _tasks;
 
@@ -189,23 +248,27 @@ socket_callback(CURL * easy,           /* easy handle */
   gs_mutex_t _taskLock;
 }
 
+static NSURLSession * sharedSession = nil;
+
 + (NSURLSession *) sharedSession
 {
-  static NSURLSession * session = nil;
-  static dispatch_once_t predicate;
+  static gs_mutex_t lock = GS_MUTEX_INIT_STATIC;
 
-  dispatch_once(
-    &predicate,
-    ^{
-    NSURLSessionConfiguration * configuration =
-      [NSURLSessionConfiguration defaultSessionConfiguration];
-    session = [[NSURLSession alloc] initWithConfiguration: configuration
-                                                 delegate: nil
-                                            delegateQueue: nil];
-    [session _setSharedSession: YES];
-  });
+  GS_MUTEX_LOCK(lock);
+  if (nil == sharedSession)
+    {
+      NSURLSessionConfiguration * configuration =
+        [NSURLSessionConfiguration defaultSessionConfiguration];
 
-  return session;
+      sharedSession
+        = [[NSURLSession alloc] initWithConfiguration: configuration
+                                             delegate: nil
+                                        delegateQueue: nil];
+      [sharedSession _setSharedSession: YES];
+    }
+  GS_MUTEX_UNLOCK(lock);
+
+  return sharedSession;
 }
 
 + (NSURLSession *) sessionWithConfiguration:
@@ -247,9 +310,6 @@ socket_callback(CURL * easy,           /* easy handle */
       NSString * caPath;
       NSUInteger sessionIdentifier;
 
-      /* To avoid a retain cycle in blocks referencing this object */
-      __block typeof(self) this = self;
-
       sessionIdentifier = nextSessionIdentifier();
       queueLabel = [[NSString alloc]
                     initWithFormat: @"org.gnustep.NSURLSession.WorkQueue%ld",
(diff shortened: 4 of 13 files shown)


rfm (changes requested):
While I like this, and I think it's generally good, I haven't really been able to completely verify it to my satisfaction.  One problem is that while the patch removes libdispatch from NSURLSession it does not remove the dependency on the availability/presence of libdispatch at configure time, so when we build without libdispatch present the testcases do nothing.  Another is that the testcases themselves still depend on libdispatch.  This was an issue for testing on Debian (which has no libdispatch package).
Another issue is that my Ubuntu system (which does have libdispatch) was aborting during the tests: I tracked that down to the leak sanitiser (I routinely test with asan/lsan), but I think that's at least as much a bug (or bad interaction) with the leak sanitiser code on arm64 ... I wasted a lot of time on that.
I built and linked with libdispatch from  source on Debian, and was able to run there: with memory leaks reported by lsan ... apparently within libdispatch in the test helper (so either the test code is not cleaning up libdispatch as it should or the libdispatch implementation leaks).
So, I guess I'd really like to see this updated to build NSURLSession when libdispatch is not available (and for the tests to not use libdispatch).
On a side note, this reminded me of the bug where GSMimeParser/GSMimeDocument can deadlock, so I made some changes to GSMime.m intended to remove that potential for deadlock (and the NSURLSession code should no longer need to make sure to initialise them safely in the main thread).

DTW-Thalion:
Interesting. OK, I'll look at this later today and see if I can clean that up in a fresh environment with no libdispatch (my Ubuntu has libdispatch as well).

DTW-Thalion:
@rfm - done. Exorcism of libdispatch complete.

NSURLSession no longer needs libdispatch to be built. The libdispatch checks are out of configure.ac (both the NSURLSession dependency test and the `dispatch_queue_create_with_target` check), and `Source/GNUmakefile` now gates the sources on `GS_HAVE_NSURLSESSION`, which needs only the ng runtime, blocks and libcurl. I built with `--disable-libdispatch` and the class compiles and the test suite passes there (117 tests).

The test HTTP server no longer uses libdispatch either. It serves each connection on its own thread rather than through dispatch queues and sources, and its makefile drops `-ldispatch`, so the leaks lsan reported inside libdispatch in the helper should be gone.

I also dropped the `+[NSURLSession initialize]` pre-warm, since your `GSMime.m` change removes the `GSMimeParser`/`GSMimeDocument` `+initialize` interdependency it was working around. I ran the transfer suite 20 times against it with no hang.

rfm:
Thanks.  Interesting changes!
Inside my Ubuntu VM on arm64,  I now get 117 passed tests with or without LSAN doing a memory check.  Which seems perfect.
However my Debian bare metal amd64 system now gets 11 test failures (though the memory leaks are gone).  The set of failures seems consistent on each run (though interleaved with passes from  other threads I guess).

eg.
Failed test:     (2026-07-09 20:25:19.886 GMT+1) simpleTaskTests.m:743 ... Folded header with continuation space is parsed correctly
Expected 'Testing' and got 'Test ing'
Passed test:     (2026-07-09 20:25:19.887 GMT+1) simpleTaskTests.m:116 ... cookie contains request cookie
Failed test:     (2026-07-09 20:25:19.887 GMT+1) simpleTaskTests.m:745 ... Folded header with continuation tab is parsed correctly
Expected 'Testing' and got 'Test ing'
Passed test:     (2026-07-09 20:25:19.927 GMT+1) simpleTaskTests.m:403 ... <DataTransferWithRedirectAndBlock> data in completion handler is not nil
2026-07-09 20:25:19.926 simpleTaskTests[515459:515490] Truncating thread name 'NSOperationQueue 0x7c33b043b178_0' to 15 characters due to platform limitations
Passed test:     (2026-07-09 20:25:19.928 GMT+1) simpleTaskTests.m:404 ... <DataTransferWithRedirectAndBlock> response is not nil
Passed test:     (2026-07-09 20:25:19.928 GMT+1) simpleTaskTests.m:406 ... <DataTransferWithRedirectAndBlock> response is an NSHTTPURLResponse
Passed test:     (2026-07-09 20:25:19.929 GMT+1) simpleTaskTests.m:407 ... <DataTransferWithRedirectAndBlock> error is nil
2026-07-09 20:25:19.928 simpleTaskTests[515459:515493] Truncating thread name 'org.gnustep.NSURLSession.WorkQueue7' to 15 characters due to platform limitations
Passed test:     (2026-07-09 20:25:19.929 GMT+1) simpleTaskTests.m:412 ... <DataTransferWithRedirectAndBlock> received data is correct
2026-07-09 20:25:19.930 simpleTaskTests[515459:515496] Truncating thread name 'NSOperationQueue 0x7c33b043bf08_0' to 15 characters due to platform limitations
Passed test:     (2026-07-09 20:25:19.930 GMT+1) simpleTaskTests.m:484 ... <DataTransferWithRelativeRedirect> Session created a valid download task
Passed test:     (2026-07-09 20:25:19.931 GMT+1) simpleTaskTests.m:168 ... <DownloadTransferWithRedirect> URLManager Session is equal to session
Passed test:     (2026-07-09 20:25:19.931 GMT+1) simpleTaskTests.m:172 ... <DownloadTransferWithRedirect> didCreateTask: Count is correct
Passed test:     (2026-07-09 20:25:19.932 GMT+1) simpleTaskTests.m:265 ... <DataTransferWithCanceledRedirect> URLManager Session is equal to session
Passed test:     (2026-07-09 20:25:19.932 GMT+1) simpleTaskTests.m:785 ... <AbortAfterDidReceiveResponseTest> Session created a valid download task
Passed test:     (2026-07-09 20:25:19.932 GMT+1) simpleTaskTests.m:174 ... <DownloadTransferWithRedirect> didCreateTask: task is equal to returned task
Passed test:     (2026-07-09 20:25:19.933 GMT+1) simpleTaskTests.m:269 ... <DataTransferWithCanceledRedirect> didCreateTask: Count is correct
Passed test:     (2026-07-09 20:25:19.933 GMT+1) simpleTaskTests.m:178 ... <DownloadTransferWithRedirect> didCompleteWithError: No error occurred
2026-07-09 20:25:19.932 simpleTaskTests[515459:515497] Truncating thread name 'org.gnustep.NSURLSession.WorkQueue8' to 15 characters due to platform limitations
Passed test:     (2026-07-09 20:25:19.933 GMT+1) simpleTaskTests.m:271 ... <DataTransferWithCanceledRedirect> didCreateTask: task is equal to returned task
Passed test:     (2026-07-09 20:25:19.933 GMT+1) simpleTaskTests.m:180 ... <DownloadTransferWithRedirect> didCompleteWithError: Count is correct
2026-07-09 20:25:19.933 simpleTaskTests[515459:515499] Truncating thread name 'NSOperationQueue 0x7c33b0455348_0' to 15 characters due to platform limitations
Passed test:     (2026-07-09 20:25:19.935 GMT+1) simpleTaskTests.m:183 ... <DownloadTransferWithRedirect> didCompleteWithError: task is equal to returned task
Failed test:     (2026-07-09 20:25:19.935 GMT+1) simpleTaskTests.m:275 ... <DataTransferWithCanceledRedirect> didCompleteWithError: An error occurred
2026-07-09 20:25:19.934 simpleTaskTests[515459:515500] Truncating thread name 'org.gnustep.NSURLSession.WorkQueue9' to 15 characters due to platform limitations
Failed test:     (2026-07-09 20:25:19.935 GMT+1) simpleTaskTests.m:187 ... <DownloadTransferWithRedirect> didWriteData: count is correct
Passed test:     (2026-07-09 20:25:19.935 GMT+1) simpleTaskTests.m:277 ... <DataTransferWithCanceledRedirect> didCompleteWithError: Count is correct
Passed test:     (2026-07-09 20:25:19.936 GMT+1) simpleTaskTests.m:190 ... <DownloadTransferWithRedirect> didWriteData: Downloaded all expected data
2026-07-09 20:25:19.936 simpleTaskTests[515459:515501] Truncating thread name 'NSOperationQueue 0x7c33b0455f88_0' to 15 characters due to platform limitations
Passed test:     (2026-07-09 20:25:19.936 GMT+1) simpleTaskTests.m:280 ... <DataTransferWithCanceledRedirect> didCompleteWithError: task is equal to returned task
Failed test:     (2026-07-09 20:25:19.937 GMT+1) simpleTaskTests.m:192 ... <DownloadTransferWithRedirect> didWriteData: Download location is not nil
Passed test:     (2026-07-09 20:25:19.937 GMT+1) simpleTaskTests.m:223 ... <DataTransferWithRelativeRedirect> URLManager Session is equal to session
Failed test:     (2026-07-09 20:25:19.937 GMT+1) simpleTaskTests.m:296 ... <DataTransferWithCanceledRedirect> didCompleteError is NSURLErrorCancelled
Expected '-125' and got (nil)
Passed test:     (2026-07-09 20:25:19.938 GMT+1) simpleTaskTests.m:227 ... <DataTransferWithRelativeRedirect> didCreateTask: Count is correct
Failed test:     (2026-07-09 20:25:19.939 GMT+1) simpleTaskTests.m:193 ... <DownloadTransferWithRedirect> location is a fileURL
Passed test:     (2026-07-09 20:25:19.939 GMT+1) simpleTaskTests.m:229 ... <DataTransferWithRelativeRedirect> didCreateTask: task is equal to returned task
Passed test:     (2026-07-09 20:25:19.941 GMT+1) simpleTaskTests.m:233 ... <DataTransferWithRelativeRedirect> didCompleteWithError: No error occurred
Passed test:     (2026-07-09 20:25:19.941 GMT+1) simpleTaskTests.m:235 ... <DataTransferWithRelativeRedirect> didCompleteWithError: Count is correct
Failed test:     (2026-07-09 20:25:19.942 GMT+1) simpleTaskTests.m:196 ... <DownloadTransferWithRedirect> dataWithContentsOfURL is not nil
Passed test:     (2026-07-09 20:25:19.942 GMT+1) simpleTaskTests.m:238 ... <DataTransferWithRelativeRedirect> didCompleteWithError: task is equal to returned task
Passed test:     (2026-07-09 20:25:19.943 GMT+1) simpleTaskTests.m:200 ... <DownloadTransferWithRedirect> string from data is not nil
Passed test:     (2026-07-09 20:25:19.943 GMT+1) simpleTaskTests.m:265 ... <AbortAfterDidReceiveResponseTest> URLManager Session is equal to session
2026-07-09 20:25:19.940 simpleTaskTests[515459:515507] Truncating thread name 'NSOperationQueue 0x7c33b0439498_0' to 15 characters due to platform limitations
Passed test:     (2026-07-09 20:25:19.950 GMT+1) simpleTaskTests.m:269 ... <AbortAfterDidReceiveResponseTest> didCreateTask: Count is correct
Failed test:     (2026-07-09 20:25:19.950 GMT+1) simpleTaskTests.m:201 ... <DownloadTransferWithRedirect> data is correct
Expected 'Hello World!' and got ''
Failed test:     (2026-07-09 20:25:19.951 GMT+1) simpleTaskTests.m:242 ... <DataTransferWithRelativeRedirect> didReceiveData: Count is correct
Passed test:     (2026-07-09 20:25:19.951 GMT+1) simpleTaskTests.m:665 ... <DataTaskWithCookies> data in completion handler is not nil
Passed test:     (2026-07-09 20:25:19.951 GMT+1) simpleTaskTests.m:271 ... <AbortAfterDidReceiveResponseTest> didCreateTask: task is equal to returned task
Passed test:     (2026-07-09 20:25:19.951 GMT+1) simpleTaskTests.m:666 ... <DataTaskWithCookies> response is not nil
Passed test:     (2026-07-09 20:25:19.951 GMT+1) simpleTaskTests.m:243 ... <DataTransferWithRelativeRedirect> data in didReceiveData is not nil
Passed test:     (2026-07-09 20:25:19.951 GMT+1) simpleTaskTests.m:275 ... <AbortAfterDidReceiveResponseTest> didCompleteWithError: An error occurred
Passed test:     (2026-07-09 20:25:19.952 GMT+1) simpleTaskTests.m:247 ... <DataTransferWithRelativeRedirect> string from data is not nil
Passed test:     (2026-07-09 20:25:19.952 GMT+1) simpleTaskTests.m:668 ... <DataTaskWithCookies> response is an NSHTTPURLResponse
Passed test:     (2026-07-09 20:25:19.952 GMT+1) simpleTaskTests.m:277 ... <AbortAfterDidReceiveResponseTest> didCompleteWithError: Count is correct
Passed test:     (2026-07-09 20:25:19.952 GMT+1) simpleTaskTests.m:669 ... <DataTaskWithCookies> error is nil
Failed test:     (2026-07-09 20:25:19.952 GMT+1) simpleTaskTests.m:248 ... <DataTransferWithRelativeRedirect> data is correct
Expected 'Hello World!' and got ''

DTW-Thalion:
OK, now that is a mystery. Will see if I can Docker a Debian environment to sort this out. @rfm - clang I presume?

rfm:
IIRC  my Debian system is using clang 21, libcurl 8.21 ... More recent than the Ubuntu VM versions.  I have no real idea whether this is a library version issue or a hardware/thread scheduling thing, and I'm done for the day.

Maybe tomorrow I can find time to set it all up on an Intel/Ubuntu system.

DTW-Thalion:
> IIRC my Debian system is using clang 21, libcurl 8.21 ... More recent than the Ubuntu VM versions. I have no real idea whether this is a library version issue or a hardware/thread scheduling thing, and I'm done for the day.
> 
> Maybe tomorrow I can find time to set it all up on an Intel/Ubuntu system.

No need - I think I found the problem - I forgot to do a curl_multi_socket_action(CURL_SOCKET_TIMEOUT) kickstart after adding a new handle. should have this fixed fairly quickly.

DTW-Thalion:
@rfm patched - try again at your convenience, sir.

rfm:
Thanks, but that makes no difference (well,  I wasn't timing things, it might be that the tests are running a little faster): still  the same 11 failures.

DTW-Thalion:
> Thanks, but that makes no difference (well, I wasn't timing things, it might be that the tests are running a little faster): still the same 11 failures.

OK, I guess a full Debian setup is required to track it down - will look into it more.

rfm:
After updating, I tried running simpleTaskTests several times and noticed that the number of tests reported occasionally varied, and got a crash on one run.  So investigating with ASAN I found a use-after-free indicating a memory management issue.  I suspect that premature deallocation of one task and re-use of the memory by another task could explain the failures.

    #0 0x7ffff7bc4474 in _i_NSURLSession___performOnWorkThread_ /home/richard/gnustep-ng/build/libs-base/Source/NSURLSession.m:547:35
    #1 0x7ffff7bd1bdd in __header_callback_block_invoke_2.30 /home/richard/gnustep-ng/build/libs-base/Source/NSURLSessionTask.m:666:17
    #2 0x5555556b9e76 in _i_URLManager__URLSession_dataTask_didReceiveResponse_completionHandler_ /home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/./URLManager.h:153:3
    #3 0x7ffff7bd19ad in __header_callback_block_invoke.29 /home/richard/gnustep-ng/build/libs-base/Source/NSURLSessionTask.m:654:14
    #4 0x7ffff79ed3b9 in _i_NSBlockOperation__main /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:606:7
    #5 0x7ffff79ec767 in _i_NSOperation__start /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:469:4
    #6 0x7ffff79f15a3 in _i_GSThreadOperationQueueImpl___thread_ /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:1466:15
    #7 0x7ffff79f0814 in _i_NSOperationQueue_Private__thread_ /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:1344:3
    #8 0x7ffff79e7cfe in _i_NSObject__performSelector_withObject_ /home/richard/gnustep-ng/build/libs-base/Source/NSObject.m:2072:10
    #9 0x7ffff7adda3c in _i_NSThread__main /home/richard/gnustep-ng/build/libs-base/Source/NSThread.m:1360:7
    #10 0x7ffff7adeabf in nsthreadLauncher /home/richard/gnustep-ng/build/libs-base/Source/NSThread.m:1500:3
    #11 0x5555556723fa in asan_thread_start(void*) crtstuff.c
    #12 0x7ffff6fa9288 in start_thread nptl/pthread_create.c:448:8
    #13 0x7ffff7027d07 in __GI___clone3 /build/reproducible-path/glibc-2.42/misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

0x7cdff21e3a38 is located 56 bytes inside of 152-byte region [0x7cdff21e3a00,0x7cdff21e3a98)
freed by thread T7 here:
    #0 0x5555556748da in free (/home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/obj/simpleTaskTests+0x1208da) (BuildId: f33f970226d73f161af289ff047cbc3efee63da1)
    #1 0x7ffff7fa62a1 in object_dispose /home/richard/gnustep-ng/build/libobjc2/runtime.c:379:2
    #2 0x7ffff79e69fd in _i_NSObject__dealloc /home/richard/gnustep-ng/build/libs-base/Source/NSObject.m:1413:3
    #3 0x7ffff7bc7fe5 in _i_NSURLSession__dealloc /home/richard/gnustep-ng/build/libs-base/Source/NSURLSession.m:1177:3
    #4 0x7ffff7bd1e85 in __destroy_helper_block_e8_32o40o48o56o /home/richard/gnustep-ng/build/libs-base/Source/NSURLSessionTask.m
    #5 0x7ffff7fabcd0 in _Block_release /home/richard/gnustep-ng/build/libobjc2/blocks_runtime.m:289:5
    #6 0x7ffff77817e8 in _i_GSMutableArray__removeAllObjects /home/richard/gnustep-ng/build/libs-base/Source/GSArray.m:701:8
    #7 0x7ffff79ed3f5 in _i_NSBlockOperation__main /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:609:3
    #8 0x7ffff79ec767 in _i_NSOperation__start /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:469:4
    #9 0x7ffff79f15a3 in _i_GSThreadOperationQueueImpl___thread_ /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:1466:15
    #10 0x7ffff79f0814 in _i_NSOperationQueue_Private__thread_ /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:1344:3
    #11 0x7ffff79e7cfe in _i_NSObject__performSelector_withObject_ /home/richard/gnustep-ng/build/libs-base/Source/NSObject.m:2072:10
    #12 0x7ffff7adda3c in _i_NSThread__main /home/richard/gnustep-ng/build/libs-base/Source/NSThread.m:1360:7
    #13 0x7ffff7adeabf in nsthreadLauncher /home/richard/gnustep-ng/build/libs-base/Source/NSThread.m:1500:3
    #14 0x5555556723fa in asan_thread_start(void*) crtstuff.c

previously allocated by thread T0 here:
    #0 0x555555674d4d in calloc (/home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/obj/simpleTaskTests+0x120d4d) (BuildId: f33f970226d73f161af289ff047cbc3efee63da1)
    #1 0x7ffff7fa3215 in allocate_class /home/richard/gnustep-ng/build/libobjc2/gc_none.c:19:3
    #2 0x5555556bbefe in testParallelDataTransfer /home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/simpleTaskTests.m:560:13
    #3 0x5555556bb07b in main /home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/simpleTaskTests.m:941:5
    #4 0x7ffff6f3df76 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16

Thread T7 created by T0 here:
    #0 0x555555658cc5 in pthread_create (/home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/obj/simpleTaskTests+0x104cc5) (BuildId: f33f970226d73f161af289ff047cbc3efee63da1)
    #1 0x7ffff7ade81a in _i_NSThread__start /home/richard/gnustep-ng/build/libs-base/Source/NSThread.m:1572:7
    #2 0x7ffff7adb999 in _c_NSThread__detachNewThreadSelector_toTarget_withObject_ /home/richard/gnustep-ng/build/libs-base/Source/NSThread.m:921:3
    #3 0x7ffff79f1dfb in _i_GSThreadOperationQueueImpl__execute /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:1547:5
    #4 0x7ffff79f07ac in _i_NSOperationQueue_Private__execute /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:1339:3
    #5 0x7ffff79f05a0 in _i_NSOperationQueue_Private_observeValueForKeyPath_ofObject_change_context_ /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:1302:3
    #6 0x7ffff79ede06 in _i_NSOperationQueue__addOperation_ /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:934:4
    #7 0x7ffff79edea5 in _i_NSOperationQueue__addOperationWithBlock_ /home/richard/gnustep-ng/build/libs-base/Source/NSOperation.m:947:3
    #8 0x7ffff7bc5d9f in _i_NSURLSession___didCreateTask_ /home/richard/gnustep-ng/build/libs-base/Source/NSURLSession.m:879:7
    #9 0x7ffff7bc6834 in _i_NSURLSession__dataTaskWithRequest_ /home/richard/gnustep-ng/build/libs-base/Source/NSURLSession.m:935:3
    #10 0x7ffff7bc68b5 in _i_NSURLSession__dataTaskWithURL_ /home/richard/gnustep-ng/build/libs-base/Source/NSURLSession.m:945:10
    #11 0x5555556bbfa2 in testParallelDataTransfer /home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/simpleTaskTests.m:588:14
    #12 0x5555556bb07b in main /home/richard/gnustep-ng/build/libs-base/Tests/base/NSURLSession/simpleTaskTests.m:941:5
    #13 0x7ffff6f3df76 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16

SUMMARY: AddressSanitizer: heap-use-after-free /home/richard/gnustep-ng/build/libs-base/Source/NSURLSession.m:547:35 in _i_NSURLSession___performOnWorkThread_
Shadow bytes around the buggy address:
  0x7cdff21e3780: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 fa
  0x7cdff21e3800: fa fa fa fa fa fa fa fa 00 00 00 00 00 00 00 00
  0x7cdff21e3880: 00 00 00 00 00 00 00 00 00 00 00 fa fa fa fa fa
  0x7cdff21e3900: fa fa fa fa 00 00 00 00 00 00 00 00 00 00 00 00
  0x7cdff21e3980: 00 00 00 00 00 00 00 00 fa fa fa fa fa fa fa fa
=>0x7cdff21e3a00: fd fd fd fd fd fd fd[fd]fd fd fd fd fd fd fd fd
  0x7cdff21e3a80: fd fd fd fa fa fa fa fa fa fa fa fa fd fd fd fd
  0x7cdff21e3b00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fa
  0x7cdff21e3b80: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x7cdff21e3c00: fd fd fd fd fd fd fd fd fd fd fd fd fa fa fa fa
  0x7cdff21e3c80: fa fa fa fa 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==545193==ABORTING

rfm:
Or perhaps the premature deallocation of a block used by a task, rather than the task itself.
  

  

  

  
  
  
  
  
>

DTW-Thalion:
Thanks @rfm - will try to make sense of it this today.

rfm:
OK, with a bit more experimentation I can see that this is probably neither an O/S nor hardware issue.

I built/installed on an Ubuntu system running on bare metal AMD hardware (EPYIC 7313P 16 core): running the tests there produced occasional failures and differing number of tests reported, though the problems occurred less frequently than on my Debian system.
The Debian system is running on AMD Ryzen 9 3900X 12 core.

My development VM is running the same version of Ubuntu, but ARM 4 cores in the VM, rather than AMD.   I tried rerunning the tests there and found that, while rare, the tests occasionally report a different number of testcases passed and/or a failure.

So it seems there are thread/timing related memory management issues common to all platforms.

DTW-Thalion:
Yes, some of the changes I made last night to try to debug  this exposed a lock inversion issue in libobjc2, and I managed to reproduce some of your Debian issues on other configs. So today will be an Tsan/Asan heavy day.

rfm:
To be fair here, I'm not sure this has ever worked properly.  I decided to try some testing on master (having built/installed libdispatch on my Debian system) and found that the tests were simply not being executed there ... commented out because of an issue in libdispatch.  However, after uncommenting the testcases, and applying the workaround from https://github.com/swiftlang/swift-corelibs-libdispatch/issues/833 to my libdispatch,  I find that I'm getting failures (including similar ASAN results) with the code on master.

DTW-Thalion:
Yes, I did that last night, I unblocked the tests and now there are issues which have been latent I suspect for a while. Running Tsan runs now to try and isolate.

rfm:
A few observations about header folding/unfolding.
1. The NSURLSessionTask code unfolds incorrectly ... it concatenates the header before and after the fold, but the fold is semantically equivalent to a single space character.  This means we should append a single space to the old value before appending the new trimmedLine in the header callback.
2. The testcase for the unfolding of headers is expecting an incorrect result, so it needs to be corrected to expect the correct result (a space in the header at the point where it was folded).
3. The curl documentation explicitly says that it doesn't deal with folding, but on my system (libcurl4-gnutls/unstable,now 8.21.0-2 amd64) it appears to be handing us the unfolded header information, which appears to be a bug unless I'm missing something).  That makes the unfolding code in NSURLSessionTask redundant but harmless (for this version of libcurl).

So I think the failure I'm seeing in the testcase for folding is an error in what the testcase is expecting.

rfm:
That is to say, this particular failure does not appear to be related to the memory management issues.

DTW-Thalion:
You are correct, I have made a fix for the folding/unfolding which I found an issue in  - we now are doing proper  RFC 7230.
I also noted a synchronization issue within the testing itself, so I have made a fix there as well. The final issue I think is I need an explicit cancel in NSURLSessionResponseCancel - testing now.

DTW-Thalion:
@rfm - try this again on your Debian. I occasionally still get a red card on MinGW but those are for things outside this NSURL* work. But otherwise, everything is green on my end.

rfm:
Sometimes I get this:
./simpleTaskTests.m:
Failed test:     (2026-07-10 22:37:23.801 GMT+1) simpleTaskTests.m:708 ... Found both cookies

    120 Passed tests
      1 Failed test

Sometimes this:
    117 Passed tests

All OK!

or this
./simpleTaskTests.m:
Failed test:     (2026-07-10 22:38:36.669 GMT+1) Passed test:     (2026-07-10 22:38:36.669 GMT+1) simpleTaskTests.m:627 ... <DataTaskWithBlock> error is nil

    119 Passed tests
      1 Failed test

So, not all fixed, but hugely improved.

DTW-Thalion:
@rfm :708 seems to be a cross-process cookie-singleton artifact, the other is a hard use-after-free which you seem to be able to trip over easily on Debian - revisiting the refcounts on the session in _transferFinishedWithCode:.

DTW-Thalion:
@rfm at your convenience, sir....

rfm:
To be clear, after your previous set of changes, while I continued to get (with lower frequency) erratic test results, I no longer had reports of use after free from ASAN.  Specifically for  I seem to get either 117 passes or 120 passes and one failure pretty equally with other counts infrequently.

So, while ASAN could be missing the issue (eg we deallocate an object, re-use the memory for a new object, and then use a pointer which was set up for the first object and now happens to point to the second one), it seems possible that the use after free was cured, and the erratic test results have a different cause.

When using the latest commit with the condition lock, behaviour seems unchanged. Still mostly either 117 passes or 120 passes and 1 failure (found both cookies), but with occasional variations like 116 passes.

So my feeling is that the latest commit is no improvement, and I'm not sure that I follow the analysis there.  Also I think the implementation, while safe enough in gnustep-base itself, is a problematic pattern best avoided (certainly not something to recommend to your team).  When you and I first learned ObjC ressurecting an object in -dealloc was basically fine (I used it in several places), but nowadays it's illegal because the runtime now has responsibility for destroying (freeing the memory) and certainly when using ARC it's going to do that.  The reason I use RETAIN,RELEASE,DEALLOC etc macros in base itself is in anticipation of using ARC rather than manual retain counting internally some day (though it may never happen), and I prefer to avoid patterns that won't work with ARC.  

In short, I think this latest change would probably be better reverted.

I suspect that my Debian system is unusual in the thread scheduling and number of the cores I the Ryzen CPU ... Perhaps every thread in the test is getting its own core and generating some particularly frequent timing issues because there's much more real parallesation than in a VM. Of course there are many other variables so that intuition could be completely wrong.

I really need to take a better look at the test code itself: I  don't know how/why it's reporting different numbers of test cases being executed, so I need to understand that.  I had thought it might be due to the issue you mentioned of different threads conflicting in writing tests results, but if so that should have been fixed.

Incidentally, perhaps I should change the test framework to use a lock to serialise test reult output, rather than depending on test cases themselves to do it.

rfm:
That being said, this (last commit excluded) already seems a big enough improvement to be worth merging. So if you would rather do that and address the remaining test issues separately, I'm ok with it.

DTW-Thalion:
This last one was to see if it actually made a difference and sounds like it doesn't, so yes, I agree to revert. I am definitely thinking the test suite may need to be looked at, I can't isolate this myself with your Debian setup - once I got through MinGW issues I am working blind.

DTW-Thalion:
Reverted -- head's back at 480546894, so the last commit's gone. Good to merge whenever suits you. I'll leave the erratic counts to you since I can't trip them on my libcurl; the :708 one is the stale shared Cookies.plist singleton, separate from the URL work.

rfm:
Great improvement, thanks.

rfm:
The erratic counts are (at least mostly) due to the cookie problem.  The reason is simply that the cookie test code only executes some testcases when it receives the cookie, so if the cookie is missing we get a different number of passed tests even though we are only seeing one failure.  The much rarer cases where the counts change without the cookie test failure presumably have a different root cause, but may also be because the test code simply isn't executing some of the testcases.

I couldn't see what is causing the cookie problem though.  I thought it might be because NSHTTPCookieStorage was not thread-safe (so I added some locking for that), however I now don't think that's it.  You mentioned a stale plist, but I would have thought that the on-disk storage in a plist could only be relevant where we have multiple processes working with the same store.  Maybe I missed something, but it looked like a single process with an in-memory store.  Perhaps there's some bug where in-memory settings are being replace by out of date info from disk?

rfm:
Never mind, I found/fixed the cookie problem.

DTW-Thalion:
Good - that would have driven me crazy knowing it was there and I couldn't find it. :)

rfm:
I reported the change of header folding behaviour in curl, and the maintainers gave corrected the curl documentation to reflect the new behaviour since 8.18.0, so we could drop our code for unfolding headers  when we are using that version or later.

PR #700, opened 2026-07-25
NSGeometry: complete NSIntegralRectWithOptions

DTW-Thalion:
NSIntegralRectWithOptions did not implement its full contract:

- The NSAlignWidth*/NSAlignHeight* options were ignored; it always aligned both the minimum and maximum edge of each axis.
- It never validated the options, so a set that does not select exactly two of an axis' minimum edge, maximum edge and size silently returned a rectangle instead of raising.
- NSAlignRectFlipped was not handled.
- An empty or zero-size rectangle was collapsed to {0,0,0,0} rather than aligned in place.
- A nearest tie used round(), which rounds a negative half away from zero, instead of towards positive infinity.

Each axis is now aligned from exactly two of its minimum edge, maximum edge and size (raising NSInvalidArgumentException otherwise); an inward option rounds an edge toward the interior of the rectangle, outward away from it, nearest to the closest unit with a tie towards positive infinity (negative infinity for a Y edge under NSAlignRectFlipped); and an empty or zero-size rectangle keeps its aligned origin.

Every option combination, the flipped tie, the empty/zero-size handling and the exception were checked against Apple Foundation on a macOS runner. The behaviour is exercised by a new NSIntegralRectWithOptions set in Tests/base/Functions/NSGeometry2.m (7 of its assertions fail against the previous implementation).

### DIFF
--- a/Source/NSGeometry.m
+++ b/Source/NSGeometry.m
@@ -36,6 +36,7 @@
 #import "common.h"
 #include <math.h>
 #import "Foundation/NSGeometry.h"
+#import "Foundation/NSException.h"
 #import "Foundation/NSScanner.h"
 #import "Foundation/NSNotification.h"
 #import "GSPrivate.h"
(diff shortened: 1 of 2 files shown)


fredkiefer:
With that change in place you could rewrite your gui change to use this function.

DTW-Thalion:
Thanks. Once this is in, I will rewrite -[NSView backingAlignedRect:options:] to call NSIntegralRectWithOptions instead of the private alignment helper it uses now.

rfm (approved):
Looks good

PR #702, opened 2026-07-27
NSUnarchiver: decode a char into a wider integer destination

DTW-Thalion:
Fixes the Windows gorm loading failure reported in gnustep/libs-gui#318 and gnustep/libs-gui#405.

BOOL is a char on most platforms, but the libobjc2 runtime defines it as an int on Windows unless built with STRICT_APPLE_COMPATIBILITY. A gorm encodes BOOL as unsigned char, so on Windows the decode reads those values into int slots and raises "expected int and got unsigned char" (the backtrace on libs-gui#318 shows this decoding NSMenuItem._enable). Every gorm then fails to load, including menus, panels and the print panel from libs-gui#405.

NSUnarchiver already widens a float into a double and converts integer sizes across platforms, but a char could only decode into a char. This widens a char archived value into a short, int, long or long long destination, preserving the sign of the archived char. A matching char destination is unchanged.

Tests/base/NSArchiver/charWidening.m covers the widening: unsigned and signed char into int/short/long/long long, edge values, and that a matching char destination still decodes unchanged. base/NSArchiver and base/coding pass, including the cross-architecture archive fixtures in base/coding.

@gcasa mentioned looking into this on libs-gui#318; this is the NSUnarchiver side of it.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,14 @@
+2026-07-27 Todd White <todd.white@thalion.global>
+
+	* Source/NSUnarchiver.m: Decode a char archived value into a wider
+	integer destination (short/int/long/long long), preserving the sign
+	of the archived char.  BOOL is a char on most platforms but an int
+	with the libobjc2 runtime on Windows, so a gorm (which encodes BOOL
+	as unsigned char) failed to load there with "expected int and got
+	unsigned char".  See gnustep/libs-gui#318 and gnustep/libs-gui#405.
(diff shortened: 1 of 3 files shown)


rfm (approved):
Looks good, and I see no reason not to loosen the type checking restriction a little more at this point.

gcasa:
@fredkiefer I just looked at this one earlier today.   I think this is fine... it looks good.  I have run into this previously on Windows, and I wasn't sure what the issue was.

PR #706, opened 2026-07-28
NSUnarchiver: decode an integer into a char destination

DTW-Thalion:
Follow up to #702, which let a char in an archive decode into a wider integer destination so that a gorm written where BOOL is a char loads on Windows, where the libobjc2 runtime makes BOOL an int.

This is the other direction. A gorm saved on Windows archives its BOOL values as ints, and every other platform decodes them into chars, which raised "expected unsigned char and got int".

The code that handles a value whose natural size is not the same as on the machine the archive was created on already reads the archived width and narrows it to the destination, and its size switch already covers one byte destinations, so the only thing needed was to let the type check fall through to it for a char destination. That also covers decodeArrayOfObjCType:count:at:. The unsigned branch of the array conversion read its size bits from the local type info rather than from the archive tag, which raised "invalid size in archive" once a char destination could reach it, so it reads the archive tag now.

Checked against a real gorm. Loading GSPrintPanel.gorm, re-archiving it through an NSArchiver subclass that writes every char as an int, which is what a Windows build produces, then reading it back raises "expected unsigned char and got int" without this change and returns the GSNibContainer with it.

Tests/base/NSArchiver/charNarrowing.m covers int, short and long long into char, signed and unsigned, plus the two array cases. 7 of its 10 assertions fail without the change and all 10 pass with it. NSArchiver, coding, NSKeyedArchiver and NSData all pass.

Relates to gnustep/libs-gui#318.

### DIFF
--- a/Source/NSUnarchiver.m
+++ b/Source/NSUnarchiver.m
@@ -250,6 +250,18 @@ typeCheck(char t1, char t2)
       if ((c == _C_FLT || c == _C_DBL) && (t1 == _C_FLT || t1 == _C_DBL))
 	return NO;
 
+/* Also allow an integer in the archive to be decoded into a char, as BOOL is
+ * a char on most platforms but an int with the libobjc2 runtime on Windows
+ * (unless that is built with STRICT_APPLE_COMPATIBILITY), so an archive (such
+ * as a gorm) written there has to decode its BOOL values into char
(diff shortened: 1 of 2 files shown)


rfm (approved):
Thanks, I was expecting this one.

Issue #707, opened 2026-07-29
KVO of a key path through an array observes nothing


DTW-Thalion:
Observing a key path whose intermediate value is an array notifies nothing. The observation is registered on the array object itself instead of on its elements, so changing a property of an element is never reported.

```objc
#import <Foundation/Foundation.h>

@interface Element : NSObject
{
  NSString *name;
}
- (NSString *) name;
- (void) setName: (NSString *)aName;
@end

@implementation Element
- (NSString *) name { return name; }
- (void) setName: (NSString *)aName { ASSIGN(name, aName); }
@end

@interface Obs : NSObject
{
@public
  int count;
}
@end

@implementation Obs
- (void) observeValueForKeyPath: (NSString *)keyPath
                       ofObject: (id)object
                         change: (NSDictionary *)change
                        context: (void *)context
{
  count++;
}
@end

int
main(int argc, char **argv)
{
  CREATE_AUTORELEASE_POOL(arp);
  Element *element = AUTORELEASE([[Element alloc] init]);
  NSMutableDictionary *dict = [NSMutableDictionary dictionary];
  Obs *obs = AUTORELEASE([[Obs alloc] init]);

  [element setName: @"a"];
  [dict setObject: [NSArray arrayWithObject: element] forKey: @"list"];

  [dict addObserver: obs
         forKeyPath: @"list.name"
            options: NSKeyValueObservingOptionNew
            context: NULL];

  [element setName: @"b"];
  printf("notifications: %d\n", obs->count);

  [dict removeObserver: obs forKeyPath: @"list.name"];
  DESTROY(arp);
  return 0;
}
```

Prints `notifications: 0`.

In `Source/NSKVOSupport.m`, `_addNestedObserversAndOptionallyDependents` registers the rest of a key path with `_addKeypathObserver([object valueForKey: key], ...)`, and that adds the observer to the observation info of the value. When the value is an array, nothing on the array ever posts a change for the element's key, and `-addObserver:forKeyPath:options:context:` of the array is not called either, so the exception it raises for an unsupported registration does not appear.

The same holds for an object that implements `-addObserver:forKeyPath:options:context:` itself in order to pass the registration on to the objects it holds: the implementation is never called. That is how the gui library's array controller exposes its arranged objects, and it is why a table column bound to `arrangedObjects.<key>` is never told that one of the objects changed, gnustep/libs-gui#99.

[...]

PR #708, opened 2026-07-29
Fix: use an intermediate object's own observer registration in a key path

DTW-Thalion:
An object in the middle of a key path may register observations on the objects it holds, through its own `-addObserver:forKeyPath:options:context:`. That implementation is never called. The observation is written into the observation info of the object itself, so nothing it holds is observed and a change to one of those objects is never reported. The gui library exposes an array controller's arranged objects that way, which is why a table column bound to `arrangedObjects.<key>` is never told that one of the objects changed, gnustep/libs-gui#99.

The rest of the key path is now registered through that implementation when the object has one, and a relay reports the change for the whole key path, on the object being observed.

A plain array or set refuses such a registration and is left as it was, so key paths that read through an array, such as the aggregate operators, keep working, and a key path to a property of the objects in a plain array still reports nothing. A proxy is left alone too.

Tests/base/NSKVOSupport/keypathRegistration.m. Six assertions fail before the change and pass after.

Refers to #707.

### DIFF
--- a/Source/NSKVOInternal.h
+++ b/Source/NSKVOInternal.h
@@ -68,11 +68,13 @@
   } while (false)
 
 @class _NSKVOKeypathObserver;
+@class _NSKVOForwardingRelay;
 
 @interface _NSKVOKeyObserver : NSObject
 {
   _NSKVOKeypathObserver	*_keypathObserver;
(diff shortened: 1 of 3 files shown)


rfm (approved):
The implementation makes sense, so if it reproduces the behavior of OSX I think it's fine.

PR #710, opened 2026-07-29
Fix pr 605

rfm:
fixup for nullability

### DIFF
--- a/Headers/Foundation/NSArray.h
+++ b/Headers/Foundation/NSArray.h
@@ -56,19 +56,22 @@ enum
 typedef NSUInteger NSBinarySearchingOptions;
 #endif
 
+NS_ASSUME_NONNULL_BEGIN
+
 GS_EXPORT_CLASS
 @interface GS_GENERIC_CLASS(NSArray, __covariant ElementT) : NSObject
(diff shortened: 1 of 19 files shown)


triplef:
With this change we’re now getting these errors in our code base, because NSObject was wrapped in `NS_ASSUME_NONNULL_BEGIN/END`, but `initWithCoder:` is defined as nullable on Apple platforms:

```
MyClass.m:67:13: error: nullability specifier 'nullable' conflicts with existing specifier 'nonnull'
   67 | - (nullable instancetype)initWithCoder:(NSCoder *)coder
      |             ^
```

Would be great if that could be fixed. I’m also a bit suspicious of using `NS_ASSUME_NONNULL_BEGIN/END` in the first place, as it usually means APIs were not really audited for whether an implementation may return null. Not sure how to best handle that. It might be a bit of a bigger project to do that.

rfm:
Thanks.  In testing/fixing this on gcc I completely forgot that the contribution might be broken on clang.  I fixed the missing markup on the method you mentioned, and disabled the assumption generally unless we define a value to turn it on.

triplef:
Thanks Richard!

PR #711, opened 2026-07-29
Fix: compare expressions and predicates on what they describe

DTW-Thalion:
Expressions and predicates compared by identity only. Two expressions for the same key path, or two predicates of the same shape, were never equal, so neither could be found in a collection with `-containsObject:` or `-indexOfObject:`, and neither could be used as a dictionary key. On OS X they compare on what they describe, and their hashes agree.

`-isEqual:` and `-hash` are implemented for:

- the key path, constant, variable, function, aggregate and binary expressions, 
- for the expression kinds that hold nothing further,
-  for comparison and compound predicates 
- and the predicates that are always true or always false. 

A comparison predicate is equal to another of the same modifier, operator, options and operands, and to nothing else; a compound predicate to another of the same type holding equal subpredicates.

This is what a predicate editor needs to tell which of its row templates a predicate belongs to, since that is decided by looking for the predicate's expressions among the ones a template holds.

Tests/base/NSPredicate/equality.m. Thirteen assertions fail before the change and pass after.

### DIFF
--- a/Source/NSPredicate.m
+++ b/Source/NSPredicate.m
@@ -433,6 +433,16 @@ extern void     GSPropertyListMake(id,NSDictionary*,BOOL,BOOL,unsigned,id*);
 
 @implementation GSTruePredicate
 
+- (BOOL) isEqual: (id)other
+{
+  return (self == other) || [other isKindOfClass: [GSTruePredicate class]];
+}
+
+- (NSUInteger) hash
+{
+  return 1;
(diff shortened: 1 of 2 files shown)


DTW-Thalion:
The Windows MinGW Clang job failed in `base/NSURLConnection/test00.m:193`, the keepalive set, which this change does not touch. Every other platform is green.

PR #712, opened 2026-07-29
Fix: CONTAINS asks whether the left side holds the right

DTW-Thalion:
CONTAINS was parsed as IN with the two sides exchanged, so the predicate stopped describing what had been written. `-predicateOperatorType` read back as IN rather than contains, `-leftExpression` gave the constant and `-rightExpression` the key path, and `name CONTAINS 'y'` read back as `y IN name`. A CONTAINS predicate built rather than parsed, which is how a predicate editor builds one, was never true: the evaluation had no case for that operator and fell through to false.

CONTAINS is now parsed as a contains predicate, keeping the side that was written first on the left, and evaluated as the mirror of IN: a string on the left is searched for the substring on the right, a collection on the left for the object on the right, honouring the case and diacritic options. IN is untouched, and the flag that exchanged the two sides is gone with its only user.

Predicates written with CONTAINS evaluate as before. Ones read back, or written out again, now read as they were written.

Tests/base/NSPredicate/contains.m. Ten assertions fail before the change and pass after.

### DIFF
--- a/Source/NSPredicate.m
+++ b/Source/NSPredicate.m
@@ -1115,6 +1115,34 @@ GSICUStringMatchesRegex(NSString *string, NSString *regex, NSStringCompareOption
 			      options: compareOptions
 				range: range] == NSOrderedSame ? YES : NO);
 	}
+      case NSContainsPredicateOperatorType:
+	/* The mirror of IN: the left hand side is the collection, or the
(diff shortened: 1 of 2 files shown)


rfm (approved):
Looks good to me.

PR #713, opened 2026-07-29
Fix: make -[NSCompoundPredicate initWithType:subpredicates:] return a working predicate

DTW-Thalion:
`-[NSCompoundPredicate initWithType:subpredicates:]` returned an instance of NSCompoundPredicate itself. The work is done by a subclass for each type, so what came back implemented none of it: `-predicateFormat`, `-evaluateWithObject:` and even `-description` raised NSInvalidArgumentException, having been left to the abstract class. The type and the subpredicates were kept, so the object looked right until it was asked to do anything. Only `+andPredicateWithSubpredicates:`, `+orPredicateWithSubpredicates:` and `+notPredicateWithSubpredicate:`, which allocate a subclass themselves, gave a predicate that worked.

The initialiser now allocates the subclass belonging to the type it is given, and raises for a type that names none. Called on a subclass, as the factory methods do, it behaves as before.

Tests/base/NSPredicate/compoundInit.m. 12 assertions fail before the change and pass after.

### DIFF
--- a/Source/NSPredicate.m
+++ b/Source/NSPredicate.m
@@ -544,6 +544,35 @@ extern void     GSPropertyListMake(id,NSDictionary*,BOOL,BOOL,unsigned,id*);
 - (id) initWithType: (NSCompoundPredicateType)type
       subpredicates: (NSArray *)list
 {
+  /* The work is done by a subclass for each type, so hand back one of those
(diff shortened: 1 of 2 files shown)


rfm (approved):
Looks good.

PR #714, opened 2026-07-29
Fix: keep the archiver whole when an object refuses to encode

DTW-Thalion:
An object whose `-encodeWithCoder:` raises corrupts the archiver's heap. While an object encodes itself the archiver points `_enc`, its current scope, at a dictionary held by its array of objects, and puts the previous scope back afterwards. That only happened on the way out through the bottom of the method, so an object that raised left `_enc` pointing at the borrowed dictionary. The archiver's `-dealloc` then released it, though the array of objects still held it, and released the array in turn, which released the same dictionary again. The top level scope leaked along with it.

What that looks like is not a clean exception but a damaged heap: the exception is delivered, and the next allocation of any size aborts in malloc, often far from the archive that caused it.

The scope is now put back if the object raises, and the exception carries on.

Tests/base/NSKeyedArchiver/encodeException.m. Its four assertions cannot run before the change, since the process aborts partway through the file; they pass after.

### DIFF
--- a/Source/NSKeyedArchiver.m
+++ b/Source/NSKeyedArchiver.m
@@ -336,11 +336,23 @@ static NSDictionary *makeReference(unsigned ref)
 
       /*
        * At last, get the object to encode itself.  Save and restore the
-       * current object scope of course.
+       * current object scope of course.  The scope is borrowed from the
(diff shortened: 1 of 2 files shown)


rfm (approved):
Nice fix, thanks.

PR #720, opened 2026-07-30
Fix mac build

iamleeg:
Three problems that blocked the build on macOS 26.5.2 Tahoe:

1. The version macros in `GSVersionMacros.h` shadow constants that Foundation defines.
2. `GSObjCRuntime.h` relies on clang ABI for weak pointers that it only declared on non-macOS platforms, but they aren't defined in the runtime headers there either.
3. `GSStandaloneXMLParser` uses `IMP` caching; the function pointers need to be cast to the correct type on macOS before use (and it's safe to do so on other platforms anyway).

Incidentally, a recent change between the last time I pulled and the current `master` fixes an `autogsdoc` crash I saw where `autogsdoc` cast an `NSXMLParser` to `GSStandaloneXMLParser` which didn't work on macOS. My original goal with this branch was to fix that, but after addressing the compiler issues I don't see that crash any more anyway.

### DIFF
--- a/Headers/GNUstepBase/GSObjCRuntime.h
+++ b/Headers/GNUstepBase/GSObjCRuntime.h
@@ -69,13 +69,15 @@
 
 #if defined(OBJC_CAP_ARC)
 #  include <objc/objc-arc.h>
-#elif !defined(__APPLE__)
+#else
 
 GS_EXPORT void objc_copyWeak(id *dest, id *src);
 GS_EXPORT void objc_destroyWeak(id *obj);
 GS_EXPORT id objc_initWeak(id *addr, id obj);
 GS_EXPORT id objc_loadWeakRetained(id *addr);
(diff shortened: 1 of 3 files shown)


rfm (approved):
Wow, that was fast, thanks.
I would have got round to testing changes on OSX, but perhaps not for a few days :-)

iamleeg:
Haha only because I happened to pull changes today :)On 30 Jul 2026, at 16:48, rfm ***@***.***> wrote:﻿
@rfm approved this pull request.

Wow, that was fast, thanks.
I would have got round to testing changes on OSX, but perhaps not for a few days :-)

triplef:
Maybe it would make sense to add macOS builds to the CI here? If our tests run on macOS we could run them there as well (same thing we did for libs-corebase in https://github.com/gnustep/libs-corebase/pull/117).

PR #727, opened 2026-08-07
Serialize key value observing changes across threads

### LINKED ISSUE #613: KVO thread safety
I think we should discuss, what kind of guarantees GNUstep's thread safety model for concurrent mutations of KVO-Observed values should be. 

I did not find a clear answer to this in Apple's documentation. Generally, Apple considers unsynchronized concurrent modification of shared state a programming error. Declaring a property atomic, provides synchronization at the accessor level. With KVO, however, even the modification of different values can cause updates in the shared data structures of the runtime. My main concern here is the "observer graph", which is the data structure that keeps track of observer dependencies (Key paths affecting observers). Updates to values who share the same observer graph, can cause notifications to be lost and bring the application into an inconsistent state.

The current KVO implementation generally synchronizes access to shared state within the observer graph, in case observed values are mutated concurrently. However, this does not at all guarantee any form of thread safety. I found instances, where shared state access is not synchronized at all. More importantly, two concurrent traversals / reconstructions of the observer graph can still bring the internal data structures into an inconsistent state. Resolving this would require the entire code path to be serialized by locking in order to make any guarantees at all. So the question to answer is whether there should be any such guarantees in the first place.

I have set up a few experiments to find out how Apple's implementation behaves, when key paths are mutated. 

For example, considering the following setup:
```objc
@interface Value : NSObject
@property (nonatomic, copy) NSString *token;
@end

@interface Holder : NSObject
@property (nonatomic, retain) Value *left;
@property (nonatomic, retain) Value *right;
@property (nonatomic, readonly) NSString *derived; // depends on: left.token, right.token
@end
```

Suppose we observe `derived` in an instance of `Holder`. I ran an experiment that mutated some of the observed values concurrently from different threads.

1. Concurrent mutation of left.token -> notification count matched write count.
2. Concurrent replacement of left -> notification count matched write count.
3. Concurrent mutation of left.token (thread A) and right.token (thread B) -> behaved correctly.
4. Concurrent replacement of left and right -> `NSInternalInconsistencyException`.

Of course, none of the tests where I observed correct behavior are proof that this is "allowed" and not just undefined behavior coincidentally behaving in a nice way. It's possible, that only the fourth case caused an update in the runtime-internal data structures, which led to inconsistency.

I see these options:
1) Guard everything with mutexes and serialize updates.
- The application state is preserved.
- Users may experience slowdowns due to lock contention when concurrently updating values.
2) Guard everything with mutexes and fail fast
- When failing to acquire a lock, throw an inconcistency excception.
- Concurrent updates are lost, internal structures are as if the update had never happened, application state may be inconsistent but at least the user knows about it.
3) No guarantees, best-effort
-  Keep locking where practical, detect concurrent graph changes as best as possible and throw in such cases
- Probably what Apple is doing

DTW-Thalion:
Setting an observed property from more than one thread loses notifications,
hangs, or crashes. willChangeValueForKey: and didChangeValueForKey: reach the
change depth and pending change dictionary of one key path observer with
nothing held between them. A concurrent change reads as nested and its
notification is dropped; concurrent writes to the dictionary corrupt the map,
which is the hang inside GSIMapCleanMap.

A recursive lock is now held from willChange to the matching didChange, on
both the observation info and the key path observer, because two key paths
through different objects share one key path observer.
observeValueForKeyPath:ofObject:change:context: runs after both locks are
dropped, so a callback setting a second observed object does not deadlock
against a thread taking the two in the other order.

The later commits pay for it. An observed set costs 266ns against 345ns,
and at 8 threads throughput rises from 2.86 to 6.62 million sets a second,
where unchanged it peaks at 2 threads and then falls below its 1 thread rate.

An observation registered with NSKeyValueObservingOptionPrior is still called
with the locks held: a prior notification precedes the setter.

Tests/base/NSKVOSupport/concurrent.m covers the four cases from the issue.

Closes #613

### DIFF
--- a/Source/NSKVOInternal.h
+++ b/Source/NSKVOInternal.h
@@ -111,6 +111,7 @@
   void                      	*_context;
   NSMutableDictionary       	*_pendingChange;
   int                        	_changeDepth;
+  int                        	_deliveryCount;
   gs_mutex_t                 	_changeLock;
 }
 - (instancetype) initWithObject: (id)object
@@ -132,6 +133,14 @@
  */
 - (void) lockChange;
 - (void) unlockChange;
+
+/* Bracket the call to the observer, which happens with no lock held.  The
+ * change dictionary is read for the length of that call, so it may not be
+ * cleared and refilled by a willChange until every such call has returned.
+ */
+- (void) beginDelivery;
+- (void) endDelivery;
+- (BOOL) isDelivering;
 @end
 
 /* Observes on behalf of a key path whose intermediate object registers
--- a/Source/NSKVOInternal.h
+++ b/Source/NSKVOInternal.h
@@ -125,7 +125,10 @@
 @property (nonatomic, assign) NSKeyValueObservingOptions options;
 @property (nonatomic, assign) void                      *context;
 
-@property (retain) NSMutableDictionary *pendingChange;
+/* Read and written only with -lockChange held, so the accessors do not need
+ * to synchronize on their own.
+ */
+@property (nonatomic, retain) NSMutableDictionary *pendingChange;
 
 /* Held from a willChange to the matching didChange.  Two key paths reached
  * through different objects share one key path observer, so the observed
--- a/Source/NSKVOInternal.h
+++ b/Source/NSKVOInternal.h
@@ -111,6 +111,7 @@
   void                      	*_context;
   NSMutableDictionary       	*_pendingChange;
   int                        	_changeDepth;
+  gs_mutex_t                 	_changeLock;
 }
 - (instancetype) initWithObject: (id)object
                        observer: (id)observer
@@ -124,6 +125,13 @@
 @property (nonatomic, assign) void                      *context;
 
 @property (retain) NSMutableDictionary *pendingChange;
+
+/* Held from a willChange to the matching didChange.  Two key paths reached
+ * through different objects share one key path observer, so the observed
+ * object's own lock does not cover this.
+ */
+- (void) lockChange;
+- (void) unlockChange;
 @end
 
 /* Observes on behalf of a key path whose intermediate object registers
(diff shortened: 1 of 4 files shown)


rfm:
This looks good to me, but I'm working on a phone rather than at my computer, so I think I need to take another look over the weekend.

rfm (approved):
OK, looks good.

triplef:
Thank you for this @DTW-Thalion! I had Claude explain this to me because I wasn’t clear on the performance impact of this based on the description and the fact that it adds locks — but it sounds like we can expect some nice performance improvements from the fast-path optimizations you did in the later commits, which is fantastic! (Our app makes heavy use of KVO and KVC.)

However while chatting with Claude it seemingly found an issue with the change. I didn’t dig into it further yet — would you mind taking a look to see if this is valid?

> ### KVO: `_dispatchWillChange` and `_dispatchDidChange` can lock and unlock different sets of `_changeLock`s
> 
> PR #727 introduced a lock (`_NSKVOKeypathObserver._changeLock`) that is acquired in
> `_dispatchWillChange` and released in `_dispatchDidChange`. The two functions each
> independently re-derive the set of locks to operate on, by fetching
> `[observationInfo observersForKey: key]` and filtering on `keyObserver.isRemoved`.
> 
> Nothing keeps those two derivations in agreement. `-[_NSKVOObservationInfo addObserver:]`
> and `-removeObserver:` take `_lock`, not `_changeLock`, so observer registration can change
> in the window between the will and the did — including from the same thread. When it does,
> the lock set and the unlock set diverge, and a `_changeLock` is either never released or
> released by a thread that does not own it.

Full issue description: [kvo-changelock-issue.md](https://github.com/user-attachments/files/30900102/kvo-changelock-issue.md)

DTW-Thalion:
It shouldn't do that - but if it does, that is a problem for sure - let me go back and check this - I may have missed something.

DTW-Thalion:
So like many AI generated analyses, it is half the picture . :)

What Claude picked up on in your analysis is something which predates the changes in #727, but an edge case I definitely missed when working on making the KVO thread safe.

Even under load and high thread count the chance you would see this is unlikely based on the way most implementations deal with trying to release a lock you don't own in an imbalance (didn't occurred in the load tests).

That said, in sorting this out, I noted a missed notification case, which is now fixed, and I ensure the locks are balancing properly.

PR #743 fixes both items.

triplef:
Thank you, I appreciate you looking into this and the quick fix!

DTW-Thalion:
Just a note, the fix costs a small 2% decrease in performance over #727 , but under thread scaling that is noise. It is still quite a bit faster than the previous single thread version.

PR #728, opened 2026-08-07
Raise for a key path that continues through a collection

### LINKED ISSUE #707: KVO of a key path through an array observes nothing

DTW-Thalion:
A key path that continues through an array or a set registers an observation
that can never fire, because the collection reports no change for a property of
the objects it holds. Observing list.name, where list holds one object, reports
nothing when that object's name changes.

_attachRestOfKeypath now hands the rest of the key path to the collection's own
addObserver:forKeyPath:options:context:, which raises, as it does when the same
key path is registered on the collection directly. An operator reads through the
collection rather than naming a property of the objects it holds, so it is still
registered.

macOS 26.5.2 raises for every form of this: through a dictionary, through a
plain array property, and through an ordered to-many property with mutation
helpers. It accepts an operator and notifies for it.

Two tests asserted that such a key path was accepted, one in
keypathRegistration.m and one in kvoToMany.m, and now assert the exception. The
second covered an ordered to-many property, which macOS refuses as well.

This changes Source/NSKVOSupport.m, as #727 does, so whichever lands second
needs a rebase.

Closes #707

### DIFF
--- a/Source/NSKVOSupport.m
+++ b/Source/NSKVOSupport.m
@@ -489,11 +489,52 @@ _registersObservationsItself(id object)
       != class_getMethodImplementation([NSSet class], selector);
 }
 
+/* Whether the value refuses observer registration, which is what a plain
+ * collection does.  A key path cannot continue through one: the collection
+ * reports no change for a property of the objects it holds.
+ */
+static BOOL
+_refusesObservation(id object)
+{
+  SEL   selector = @selector(addObserver:forKeyPath:options:context:);
+  IMP   implementation;
+
+  if (nil == object)
+    {
+      return NO;
+    }
+  implementation
+    = class_getMethodImplementation(object_getClass(object), selector);
+
+  return implementation
+      == class_getMethodImplementation([NSArray class], selector)
+    || implementation
+      == class_getMethodImplementation([NSSet class], selector);
+}
+
 /* Observe the rest of a key path on the value of its first key. */
 static void
 _attachRestOfKeypath(_NSKVOKeyObserver *keyObserver)
 {
   id value = [keyObserver.object valueForKey: keyObserver.key];
+
+  /* An operator reads through the collection and reports the result for the
+   * whole key path, so it is registered as any other key is.  Any other key
+   * names a property of the objects the collection holds, and the collection
+   * raises for it, as it does when the same key path is registered on it
+   * directly.
+   */
+  if (_refusesObservation(value)
+    && NO == [keyObserver.restOfKeypath hasPrefix: @"@"])
+    {
+      _NSKVOKeypathObserver *keypathObserver = keyObserver.keypathObserver;
+
+      [value addObserver: keypathObserver.observer
+              forKeyPath: keyObserver.restOfKeypath
+                 options: keypathObserver.options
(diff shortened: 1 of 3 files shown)


rfm (commented):
While this looks/sounds reasonable, I have never really used KVO and am not familiar with how it should work, so while I have no real problem accepting this PR  I'd rather leave it unmatched for a few days to let others comment.

DTW-Thalion:
> While this looks/sounds reasonable, I have never really used KVO and am not familiar with how it should work, so while I have no real problem accepting this PR I'd rather leave it unmatched for a few days to let others comment.

Agree - maybe @triplef or @hmelder have thoughts?

triplef:
I think that makes sense and sounds like it matches macOS behavior. Thank you.

rfm:
OK, I think 5 days is long enough, and we can always revert/fix later.

PR #729, opened 2026-08-07
Free node strings the old dictionary does not own

### LINKED ISSUE #721: NSXMLDocument leaks a few libxml2 strings on every parse
Parsing data into an NSXMLDocument and releasing it leaves a few libxml2 strings behind. It is small, 53 bytes in 3 allocations for the document below, but it happens on every parse, and it is what makes several libs-gui test files report leaks under the leak sanitizer.

Reproducer, base only, no AppKit:

```objc
#import <Foundation/NSXMLDocument.h>

int main(int argc, char **argv)
{
  NSAutoreleasePool *pool = [NSAutoreleasePool new];
  NSString *xml = @"<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
                   "<document type=\"com.apple.InterfaceBuilder3.Cocoa.XIB\">"
                   "<objects><customObject id=\"a\" customClass=\"NSObject\"/>"
                   "<buttonCell id=\"b\" title=\"Hello\"/></objects></document>";
  NSData *data = [xml dataUsingEncoding: NSUTF8StringEncoding];
  NSXMLDocument *doc;

  doc = [[NSXMLDocument alloc] initWithData: data options: 0 error: NULL];
  if (doc != nil)
    {
      [doc nodesForXPath: @"/document" error: NULL];
    }
  [doc release];

  [pool release];
  return 0;
}
```

Built against a base compiled with GNUSTEP_WITH_ASAN=1 and run with detect_leaks=1, that reports:

```
SUMMARY: AddressSanitizer: 53 byte(s) leaked in 3 allocation(s).

Direct leak of 38 byte(s) in 1 object(s) allocated from:
    #1 xmlStrndup                              libxml2.so.2
    #4 xmlSAX2StartElementNs                   libxml2.so.2
    #8 xmlParseElement                         libxml2.so.2
    #9 xmlParseDocument                        libxml2.so.2
    #10 xmlReadMemory                          libxml2.so.2
    #11 -[NSXMLDocument initWithData:options:error:]   NSXMLDocument.m:138
```

Only a handful of strings are left, not the whole tree, so xmlFreeDoc is running. It looks like a few nodes are getting dropped in -[NSXMLNode dealloc], where the subnodes are detached and theNode->_private is cleared before the node is freed, and where there is already a FIXME about not knowing when a namespace node can safely be freed. I have not tried to change any of that, since getting the ownership wrong there would be a good deal worse than the leak.

Where it shows up in practice: +[GSXibKeyedUnarchiver checkXib5:] builds an NSXMLDocument to decide whether some data is an Xcode 5 xib, and releases it correctly, but the leak is left behind on every xib load. With the libs-gui suite run under the leak sanitizer, GSXib5KeyedUnarchiver/buttonCell.m, GSXib5KeyedUnarchiver/menu.m and NSNibLoading/basic.m all report leaks that come back to this.

DTW-Thalion:
Detaching a node moves it to a document of its own, and setTreeDoc replaces the
strings of its subtree with entries in the dictionary of that document. A parser
interns short strings and allocates the rest, so a replaced string is owned
either by the old dictionary or by the node. Replacing an allocated one without
freeing it loses it.

setTreeDoc now asks xmlDictOwns which it is: a string the old dictionary owns is
left to it, one it does not own is freed once replaced. That covers the text
content, the element name, the attribute name, and the namespace href and
prefix.

libxml2 2.9.14 parsing and freeing the same document with no GNUstep in the
process leaks nothing.

Parsing alone does not leak either. A node has to be wrapped first, so
-rootElement is enough. The document in the report leaks 53 bytes in 3
allocations, its three attribute values too long to intern.

No test in the suite can fail on a leak.
Tests/base/NSXMLNode/detachedStrings.m covers the hazard the fix introduces
instead: a detached node's strings stay valid once the document they were parsed
into is released.

Closes #721

### DIFF
--- a/Source/NSXMLNode.m
+++ b/Source/NSXMLNode.m
@@ -88,6 +88,44 @@ cleanup_namespaces(xmlNodePtr node, xmlNsPtr ns)
     }
 }
 
+/* Move one string of a node to the document the node is joining.
+ * A parser interns short strings in the dictionary of the document it builds
+ * and allocates the rest, so a string reached from a node is owned either by
+ * that dictionary or by the node itself.  Replacing an allocated one without
+ * freeing it loses it, and freeing one the dictionary owns would free it
+ * twice, so which it is has to be asked.
+ */
+static const xmlChar *
+adoptString(const xmlChar *str, xmlDocPtr oldDoc, xmlDocPtr doc, int mode)
+{
+  int	owned;
+
+  if (str == NULL)
+    {
+      return NULL;
+    }
+  owned = (oldDoc != NULL && oldDoc->dict != NULL)
+    ? xmlDictOwns(oldDoc->dict, str) : 0;
+
+  if (mode == 1)
+    {
+      const xmlChar *adopted = xmlDictLookup(doc->dict, str, -1);
+
+      if (0 == owned && adopted != str)
+        {
+          xmlFree((xmlChar *)str);
+        }
+      return adopted;
+    }
+  if (mode == 2 && owned)
+    {
+      /* The dictionary of the old document is about to go, and the new
+       * document has none to put the string in. */
+      return xmlStrdup(str);
+    }
+  return str;
+}
+
 /* Recursively set document pointer for a node tree.
  * This is needed when we can't use xmlDOMWrapAdoptNode due to bugs.
  * Also handles string adoption from old dictionary to new dictionary.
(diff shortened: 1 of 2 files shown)


rfm (approved):
I never knew about the dictionary ownership. Good find.

PR #730, opened 2026-08-07
Let the run loop block when a socket stream event is unhandled

### LINKED ISSUE #375: Deadlock and Increasing Memory Leak in NSURLConnection/test02 on Windows on ARM64
image

This requires some further debugging in lldb.

DTW-Thalion:
Closes #375. Two defects behind it.

-[GSSocketOutputStream runLoopShouldBlock:] answered NO while an event
had been delivered and not consumed, and set the trigger, so the run
loop polled with a zero timeout and dispatched the stream on every pass.
After a request is sent on a kept alive connection the delegate has
nothing to write, so nothing clears NSStreamEventHasSpaceAvailable and
the poll never ends: 2.97 seconds of processor time in a window of 3
seconds. Blocking is safe, because a delegate with more to write is
driven by the loop in -_dispatch, measured at 8MB in 163 events.
Clearing the trigger instead loses the read, since the socket's event is
registered against this stream and its -_dispatch hands FD_READ to the
sibling.

The tests bound the fixed ports 1229 to 1238. A port in TIME_WAIT cannot
be rebound, so on Windows the bind fails, and
-[SimpleWebServer _startListening] answered YES anyway, leaving requests
to time out with nothing reported. Each server now asks for port 0 and
reports the port it was given.

test02 run 6 times back to back now passes every time; master only passed 1
in 5.

### DIFF
--- a/Source/GSSocketStream.m
+++ b/Source/GSSocketStream.m
@@ -2993,15 +2993,13 @@ setNonBlocking(SOCKET fd)
 #if	defined(_WIN32)
 - (BOOL) runLoopShouldBlock: (BOOL*)trigger
 {
+  /* The socket's event is registered against this stream and carries the
+   * events of the sibling input stream as well, so the run loop is left to
+   * block on it. A delegate with more to write is driven by the loop in
+   * -_dispatch, which sends NSStreamEventHasSpaceAvailable for as long as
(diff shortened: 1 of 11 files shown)


rfm (approved):
This seems entirely sensible

Issue #701, opened 2026-07-27
support system configuration for user dir paths


rmottola:
I think it would be nice to support dynamic path configuration in:
` /etc/xdg/user-dirs.defaults`

As a reference: [https://www.freedesktop.org/wiki/Software/xdg-user-dirs/](url) 

For example:

```
DESKTOP=Desktop
DOWNLOAD=Downloads
TEMPLATES=Templates
PUBLICSHARE=Public
DOCUMENTS=Documents
MUSIC=Music
PICTURES=Pictures
VIDEOS=Videos
PROJECTS=Projects

```

rmottola:
Topics:

-  just a configure time path detection? For packaged gnustep it would not allow later changes by a system, it is preferred to reparse later
- reparse it at every App launch?
- handle fallback
- translations: only on display (like we have now and which is Mac and Win philosophy) or does linux sometimes
- system default to disable look-up logic?

rfm:
I have added a provisional implementatlon,  parsing $HOME/.config/user-dirs.dir and using the values found in preference to the builtin defaults.  I also added code to let similar values in the GNUstep config file override those provided by XDG (just to be consistent in allowing our own config to have the final say).

DTW-Thalion:
PR #731 adds on to the items you added @rfm - still needs a call on translations.

rfm:
FYI Riccardo.  My understanding from reading around is that apps such as gnustep apps) are not supposed to read the stuff in /etc ... that is default information for the xdg-user-dir tool, not for apps.  That tool generates the the files (in the home directory of the user) that we are supposed to read, and nomally that's done the first time a user logs in.

I don't think it's the role of the base library to run the tool,  but arguably it might be the role of the GWorkspace app to do so.

rmottola:
No, i don't think it is the role of the library to run the tool at all.
However, given the fragmentation, I think it is good to read those files to know what the configuration could be.

Translations are delicate: in GWorkspace I implemented the approach where the original path is in english and then gets translated on Display. Both Apple and Windows do so. An idea would be to add an extra variable that says if they are translated or no and that would allow gui and GWorkspace to know if the name still needs to be translated or not.

[...]

PR #731, opened 2026-08-07
Read the XDG system defaults for the well known directories

DTW-Thalion:
Addresses #701 

The user's own $HOME/.config/user-dirs.dirs was read, but not the system
wide user-dirs.defaults, which is where a user with no file of their own
gets these values. It is looked for in each directory named by
XDG_CONFIG_DIRS, /etc/xdg where that is unset, and it names a
subdirectory of the home directory without the XDG_ prefix and the _DIR
suffix. An earlier directory in XDG_CONFIG_DIRS takes precedence over a
later one, and the user's own file takes precedence over both.

The system file is parsed with no owner to check against, since it
belongs to root rather than to the user; ParseConfigurationFile() still
refuses a file which anyone other than its owner may write.

Setting GNUSTEP_XDG_USER_DIRS to NO in the GNUstep configuration file
turns the lookup off. An absent key leaves it on.

NSDesktopDirectory answered a hardcoded "Desktop", so the desktop value
from XDG or from GNUSTEP_DESKTOP_DIR was assigned and never read. It now
reads gnustepUserDesktop, as the other user directories do.

Both files are read at each launch, so a change takes effect without a
rebuild. base/NSPathUtilities, NSFileManager and NSBundle are 369 passed
and 0 failed, before and after.

### DIFF
--- a/Source/NSPathUtilities.m
+++ b/Source/NSPathUtilities.m
@@ -606,8 +606,24 @@ ExtractValuesFromConfig(NSDictionary *config)
 
   /* Check for user subdirectories.  Ones from the GNUstep file are added
    * second, so they override the XDG ones.
+   *
+   * Setting GNUSTEP_XDG_USER_DIRS to NO in the GNUstep configuration file
+   * turns the XDG lookup off, leaving the builtin names and the GNUSTEP_
+   * keys below.  An absent key leaves the lookup on.
    */
-  xdg = UserDirsParseXDG();
+  {
+    NSString	*useXDG = [c objectForKey: @"GNUSTEP_XDG_USER_DIRS"];
+
+    if (nil == useXDG || YES == [useXDG boolValue])
+      {
+	xdg = UserDirsParseXDG();
+      }
+    else
+      {
+	xdg = nil;
+      }
+  }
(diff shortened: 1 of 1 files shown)


rfm:
I may be wrong  but from memory of implementing this, I think The proposed PR is completely wrong, perhaps because it tries to do literally what Riccardo asked for, and that's not right:  the existing code is, apart from any bugs, basically complete.  The xdg system defaults are not for an app (like gnustep apps) to read, they are something for the xdg-user-dirs tool to read.  That tool then generates the localised file that apps read (normally using a library, in our case there's no need for an external library because we already have parsing code).
Normally the system runs xdg-user-dirs to create the file (if necessary) when the user logs in.
So, barring bug fixing, I think there's nothing to do here, unless we want to run the xdg-user-dirs tool if we think some issue has prevented it from running.

DTW-Thalion:
OK, that's fair - and you're right I was focused on what Riccardo was suggesting, but missed the wider context of XDG. Will close this and if there is a reason to resurrect it, I can just pull the branch.

rmottola:
I still am convinced we should try to read the system default as a backup - just reading, not creating, because in piratical experience the defaults in  each home directory don't always exist.

Issue #448, opened 2024-09-10
Speed-Up CI


hmelder:
There are a couple of tests which take longer than 30s and time out. I've increased the timeout to 300s, but we should really fix these tests.

```
base/NSStream/socket.m:
Failed file:  socket.m did time out!
base/NSURL/test00.m:
Failed file:  test00.m did time out!
base/NSURL/test02.m:
Failed file:  test02.m did time out!
```

DTW-Thalion:
PR #730 should reduce the need for these long timeouts as it was a result of a problem with #375 
cc @hmelder

DTW-Thalion:
Measured the three files named here, on Linux with the network working.

base/NSURL/test00.m takes 50.9s of the 55s the whole NSURL set takes, and
the time is deliberate. The last section runs the keepalive drop cases
with the respond helper delaying 0, 4, 8 and 12 seconds, and each delay
runs 2 times, which is 48 seconds of waiting. It does not time out here,
so the raised timeout is accommodating a test that sleeps rather than one
that hangs. Running those 4 cases concurrently, or shortening the delays
where the keepalive thresholds allow, is where the time is.

base/NSURL/test02.m takes 0.87s here, so what fails on Windows is not
slowness.

base/NSStream/socket.m connects to www.google.com on port 80, giving the
run loop a 30 second budget, and then again on 443. That is the source of
anything over 30 seconds when the runner has no egress, and a test which
needs an outside host will stay slow or flaky in CI. Pointing it at a
local server, or skipping when the name does not resolve, would take it
out of the picture.

Since #730 the NSURLConnection tests bind port 0 rather than 1234, which
is the port the NSURL helpers are launched with, so those 2 sets no
longer contend for it within a run.

PR #733, opened 2026-08-07
tests: skip the external host tests where there is no route out

DTW-Thalion:
One of the 3 slow tests in #448. socket.m connects to www.google.com on
port 80 and again on 443, giving the run loop 30 seconds each time.

Measured with the name changed to one that does not resolve: master takes
61 seconds and reports 2 failures. -hostWithName: answers a host whatever
the name is, so the check here is whether -address is nil. When it is,
the 2 sets skip and the file takes 1 second. With the network present the
file passes 38 tests, as before.

The wait is 10 seconds rather than 30. The per event NSLog in the stream
delegate is now NSDebugMLLog.

base/NSURL/test00.m and base/NSURL/test02.m are unchanged. test00 waits
on purpose, and test02 runs in under a second here.

### DIFF
--- a/Tests/base/NSStream/socket.m
+++ b/Tests/base/NSStream/socket.m
@@ -11,6 +11,12 @@
 #define	SSL_SUPPORTED	1 /* Assume Apple supports it */
 #endif
 
+/* How long to wait for a host outside this machine. A connection which is
+ * going to work does not need 30 seconds, and a runner with no route out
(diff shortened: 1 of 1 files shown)


rfm:
I'm away from my computer, but I'm not sure what the point of the external site tests actually was.  Maybe they were just easy to add testcases from before we implemented a local server?  It might make sense to just remove them.

PR #737, opened 2026-08-08
Fix: NULL pw_gecos, the home directory fallback and TMPDIR

DTW-Thalion:
NSFullUserName dereferences pw.pw_gecos without testing it, and ignores the
result pointer from getpwnam_r. bionic sets pw_gecos to NULL, so the call
segfaults. The getpwnam branch below already tests the pointer, and the same
file uses the fuller test at around line 1991. Tests/base/Functions goes from a
crash after 1 test to 229 passed.

NSHomeDirectoryForUser used #elif for the Android branch, which left the POSIX
getpwnam_r fallback out of the build. androidFilesDir is nil in any process
without a JNI context, so a tool, a daemon or a test binary got nil back.
Android now tries its own answer first and falls through to the fallback,
which answers / on bionic.

NSTemporaryDirectory consults TEMP and TMP but not TMPDIR, which is the POSIX
name and the one Android sets. It fell back to /tmp, which does not exist
there, and the root is mounted read-only. The nesting is flattened rather than
deepened.

The first two sit on the general POSIX paths rather than in Android specific
code. Measured on Android x86_64: Tests/base/NSPathUtilities 164 passed and 0
failed, Tests/base/Functions 229 passed and 0 failed.

### DIFF
--- a/Source/NSPathUtilities.m
+++ b/Source/NSPathUtilities.m
@@ -1980,7 +1980,8 @@ NSHomeDirectoryForUser(NSString *loginName)
     {
       s = [[NSProcessInfo processInfo] androidFilesDir];
     }
-#elif !defined(_WIN32)
+#endif
+#if !defined(_WIN32)
 #if     defined(HAVE_GETPWNAM_R)
   if (nil == s)
     {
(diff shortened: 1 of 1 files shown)


DTW-Thalion:
@rfm #737 - #739 are needed before Android CI (#740) will pass CI and will be mergable.

rfm (approved):
Looks good

PR #740, opened 2026-08-09
CI: build and run the test suite on Android

### LINKED ISSUE #538: Android CI Runner
I already added an Android CI in libobjc2, but the GNUstep base test suite is more complex due to some test cases loading bundles relative to the test executable.

DTW-Thalion:
Do not merge without #739 first.

Adds a workflow that cross-builds gnustep-base for x86_64-linux-android API 31
and runs all of Tests/base on a headless emulator. Closes #538 .

Nothing is disabled. ICU, libxml2, libxslt, GnuTLS, libcurl, libffi,
libdispatch and iconv are cross-built and cached, so the run covers the same
code a Linux run does.

The deployment rules in tests.sh are the substance. Resources are found
relative to the executable, tests launch helpers and load bundles by path,
several read their own source back, and gnustep-tests leaves a GNUmakefile in
the working directory that two tests read. The emulator runs as the shell
user, which SELinux does not allow to create a socket under /data/local/tmp,
so the deployment lives under /data/local/tests, labelled
shell_test_data_file. Without that, gdnc cannot bind its NSMessagePort and
every distributed notification test fails.

#737 and #738 have merged and this branch is rebased on master, so the TMPDIR
support NSTemporaryDirectory needs is in place.

Measured on this branch: 116 directories, 13457 passed, 4 failed, 42 dashed
hopes. The four are the environment assumptions listed in
expected-failures.txt, and #739 turns them into hopes so the list can go.

### DIFF
--- a/.github/scripts/android/android.cross.config
+++ b/.github/scripts/android/android.cross.config
@@ -1,0 +1,25 @@
+# libs-base cross answers for Android x86_64-linux-android, API 31.
+# Produced by running each configure test on an emulator.
+# NDK 29.0.14206865, libobjc2 -fobjc-runtime=gnustep-2.2.
+cross_CMDLINE_TERMINATED=1
+cross_NEED_WORD_ALIGNMENT=0
+cross_VASPRINTF_RETURNS_LENGTH=1
+cross_VSPRINTF_RETURNS_LENGTH=1
+cross_reuseaddr_ok=1
+cross_have_kvm_env=0
+cross_working_register_printf=0
+cross_wide_register_printf=0
+cross_have_poll=yes
+cross_ffi_ok=yes
+cross_gs_cv_objc_works=yes
+cross_gs_cv_objc_load_method_worked=yes
+cross_gs_cv_objc_compiler_supports_constant_string_class=yes
+cross_non_fragile=yes
+cross_have_unexpected=yes
+cross_safe_initialize=yes
+cross_objc2_runtime=1
+cross_utf8literal_ok=yes
+cross_gs_cv_program_invocation_name_worked=no
+cross_found_iconv_libc=no
+cross_found_iconv_liconv=yes
+cross_found_iconv_lgiconv=no
--- a/.github/scripts/android/base.sh
+++ b/.github/scripts/android/base.sh
@@ -1,0 +1,165 @@
+#!/bin/bash
+#
+# Cross-install tools-make, then build libs-base, its tools, the per-directory
+# test helpers and the test bundles, all for Android.
+#
+set -eu
+
+: "${ANDROID_NDK_HOME:?}"
+: "${GS_PREFIX:?}"
+: "${GS_BASE:?}"                     # the libs-base checkout
+: "${GS_API:=31}"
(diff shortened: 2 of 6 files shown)


rfm:
Sounds really good, but one point for discussion;  I really don't like using lots of power to repeatedly build software unnecessarily.  We already have several workflows running on every commit, for very little benefit, but (though it rarely happens) we might have long periods with no commits where we don't notice failures caused by changes in libraries we depend on.
I'd therefore like to restructure to a small set of workflows run on all commits, and a larger set run periodically at quite a long interval.
Ideally I would like to have some workflows run daily if a commit has occurred but weekly otherwise. I don't know a way to do that though.

For me obvious high usage setups we should test on every commit would be:
Modern Linux with clang 
Modern windows with GCC
Old Linux with GCC (oldest version we support).

The first is the most heavily used setup.  The second most heavily used would be a modern Linux with GCC, but if we test the most common OS with clang we can probably afford to test modern GCC on Windows)

Anyway, exactly how we should schedule testing other workflows is non-obvious to me.  We can always run them manually when we want to, so I would err on the side of having longer delays.

DTW-Thalion:
I was thinking something similar and had it on my list to try to trim CI - I will do some research and come back with some options. CI for me is ~15 mins a run and that definitely gets expensive.

DTW-Thalion:
On CI efficiency: one push runs 11 checks and 172 runner minutes, 147 of them the CI workflow. The 4 Windows jobs are 127 of the 172. Over the last 90 days master took 192 push runs of the CI workflow on 48 days.

Ubuntu x64 Clang, Windows x64 MinGW GCC and GCC-4.9.4 together cost 40.9 minutes, 24% of the 172. Over those 90 days, that set per push with everything else once on a day master moved is 14100 runner minutes against 33000 today. Everything else weekly is 9600.

GitHub has no direct expression for daily if a commit occurred and weekly otherwise. Two schedule entries and a guard job cover it: the guard sets an output when the weekly cron fired or when git log --since='24 hours ago' origin/master is not empty, and every other job carries needs: guard and an if on that output.

Two changes need no scheduling decision, filed as #745 and #746. The Windows library build ran plain make where the other platforms pass -j, and Build source went from 7.6 minutes to 4.2. No workflow had a concurrency group, and 80 of the 192 master runs were superseded by a newer push before finishing, spending 7266 of 28224 CI runner minutes on commits already replaced.

On Windows MinGW GCC, 74% of the test phase is the per-file build of the test tools and 9% is the tests. NSURL is 17.8 minutes per push, which is the sleeping in #448, and headers is 9.9 minutes to check that 94 headers compile. Windows arm64 costs 2.2 times Windows x64 MSVC Clang for the same work, so it is the first job to take off the per-push set.

The smaller set covers Linux with clang, Windows with GCC and GCC 4.9 on every push, and a break on Rocky, arm64, MSVC, MinGW clang, NetBSD or Android shows within a day, or a week on the weekly variant.

findutils is missing from the rockylinux:8 image, so gnustep-tests cannot list its directories and that job has been reporting success without running a test. #744 adds the package.

PR #741, opened 2026-08-10
Fix: initialise a process with the environment it was started with

DTW-Thalion:
A process initialised through GSInitializeProcess() or
GSInitializeProcessAndroidWithArgs() with no environment array gets an empty
environment dictionary, so [[NSProcessInfo processInfo] environment] is empty
and GNUSTEP_CONFIG_FILE, and everything else read from the environment, has no
effect. GSInitializeProcessAndroidWithArgs() also takes an envp argument and
then calls _gnu_process_args() with NULL, discarding it. An Android activity
has no main(), so that is how every GNUstep application on Android starts, and
such an application cannot be given a configuration file at all.

_gnu_process_args() now falls back to the process environment when it is given
none, and GSInitializeProcessAndroidWithArgs() passes on the envp it was called
with. On Windows the existing wide-character environment still takes
precedence.

Tests/base/NSProcessInfo/initialise.m

1 failed before, 0 after.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2026-08-10 Todd White <todd.white@thalion.global>
+
+	* Source/NSProcessInfo.m:
+	* Tests/base/NSProcessInfo/initialise.m:
+	Use the process environment when a process is initialised without
+	one, and pass the environment given to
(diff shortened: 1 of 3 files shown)


rfm (approved):
Yes, the environment should never be empty.

PR #747, opened 2026-08-10
Fix: read a whole request in the test HTTP server

DTW-Thalion:
The HTTP server the NSURLSession tests run against treats one read as one request. handleClientSocket: passes the bytes of a single 4096 byte recv() to handleConnectionData:forSocket:, which takes everything after the blank line as the body without consulting Content-Length. Whether the body is complete therefore depends on how the sender happens to segment the request, and a body larger than the buffer can never be complete. Resources/largeBody.txt is 3642 bytes, so the large upload case fitted in one read on most platforms and failed on Rocky Linux 8, where uploadTaskTests.m:68 reported that HTTPBody did not match what was sent.

handleClientSocket: now accumulates what it reads and hands a request to the routes only once the headers and Content-Length bytes of body have arrived, consuming exactly that much of the buffer so a reused connection still works. largeBody.txt grows to 10926 bytes, which is larger than one read, so the case is exercised everywhere rather than by accident.

Tests/base/NSURLSession/uploadTaskTests.m:68, reached twice through /largeUploadOK.

With the larger fixture and the old server, 2 assertions failed. With the fix, 0 failed and the set went from 115 to 117 passed.

### DIFF
--- a/Tests/base/NSURLSession/Helpers/HTTPServer.m
+++ b/Tests/base/NSURLSession/Helpers/HTTPServer.m
@@ -60,6 +60,53 @@ NSString (ServerAdditions)
 }
 @end
 
+/* The length of the complete request at the front of data, or 0 when the whole
+ * request has not arrived yet.  A request is complete once the blank line
+ * ending the headers has been read along with as many body bytes as
+ * Content-Length asks for.  A request without Content-Length has no body.
+ */
+static NSUInteger
+requestLength(NSData *data)
+{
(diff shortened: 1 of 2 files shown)


rfm (approved):
I've never been happy with adding this http server for testing NSURLSession: having different server implementations in different test directories makes no sense (we should have a single http testing server-side framework we use from all http testing).
However, we are where we are, and the bugfix is good.

PR #748, opened 2026-08-11
Android: start a packaged tool from the directory it is installed in

DTW-Thalion:
An application cannot start gdnc, gpbs or any other tool on Android.
+[NSTask launchPathForTool:] searches the tool directories a filesystem layout
gives, which are inside the writable data directory, and Android refuses execve
there for any application targeting API 29 or higher:
https://developer.android.com/about/versions/10/behavior-changes-10

A package carries an executable file only as lib/<abi>/lib<name>.so, installed
into one directory whose name is settled when the package is installed:
https://developer.android.com/ndk/guides/abis
A layout cannot name that directory, but the library is installed into it too,
so GSPrivateSymbolPath() names it, which is the mechanism a config file
location beginning './' already uses.

That directory is added ahead of the layout's, which are kept so a tool built
and installed outside a package is still found. executablePath() tries
lib<name>.so as it tries the extensions Windows adds. A tool started this way
is given the directory on LD_LIBRARY_PATH, which execve does not provide, and
the process temporary directory, within which a port is named.

Tests/base/NSPathUtilities/toolsdirectory and Tests/base/NSTask/toolpath cover
a process that has no package.

### DIFF
--- a/Source/Additions/NSTask+GNUstepBase.m
+++ b/Source/Additions/NSTask+GNUstepBase.m
@@ -181,6 +181,24 @@ executablePath(NSFileManager *mgr, NSString *path)
     {
       return path;
     }
+#if	defined(__ANDROID__)
+  /* A package may carry an executable file only as lib/<abi>/lib<name>.so, so
+   * a tool installed from one is named lib<name>.so rather than <name>.  A
+   * tool built and installed outside a package keeps its own name and is found
+   * above.
+   */
+  {
+    NSString	*dir = [path stringByDeletingLastPathComponent];
+    NSString	*name = [path lastPathComponent];
+
+    path = [dir stringByAppendingPathComponent:
+      [NSString stringWithFormat: @"lib%@.so", name]];
+    if ([mgr isExecutableFileAtPath: path])
+      {
+	return path;
+      }
+  }
+#endif
 #endif
   return nil;
 }
--- a/Source/GSPrivate.h
+++ b/Source/GSPrivate.h
@@ -568,6 +568,16 @@ GSPrivateStrExternalize(GSStr s) GS_ATTRIB_PRIVATE;
 NSString *
 GSPrivateSymbolPath(Class theClass) GS_ATTRIB_PRIVATE;
 
+#if	defined(__ANDROID__)
+/* GSPrivateAndroidToolsDirectory() returns the directory this library was
+ * loaded from, which on Android is the directory a package's executable files
+ * are installed into and the only directory a tool may be started from.
+ * Returns nil where the library's own location cannot be determined.
+ */
+NSString *
+GSPrivateAndroidToolsDirectory(void) GS_ATTRIB_PRIVATE;
+#endif
+
 /* Combining class for composite unichars
  */
 unsigned char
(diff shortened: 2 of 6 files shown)


rfm (commented):
I don't really understand this one.  I see that we can't run an executable from certain locations on Android, but it's not clear to my why this is an issue for base (or how the changes fix it).
Clearly it makes no sense for the standard tools directory to point to a location that's unusable, but the location of the tools directory comes from the config file, and is initially set up by gnustep-make from the selected filesystem layout.
So I would expect us to have an android file added to the FilesystemLayouts directory and for gnustep-make to be configured using that, so that correct locations are built in to gnustep-base.
But maybe the real locations for executables (if indeed modern android allows this at all) have to be determined dynamically at runtime.  There *is* an existing mechanism for paths in the filesystem layout to be determined relative to the config file location (and that in turn from an environment variable or relative to the base library), but maybe that's not workable on android an we need some extension?
The bits about setting up the environment are also unclear to me :-(

rfm (commented):
I don't see a point in a new method for android native libraries: the existing API provides NSStandardLibraryPaths() (which calls NSSearchPathForDirectoriesInDomains()) to return the paths to directories which may contain libraries.

DTW-Thalion:
I'll send you a document where I traced all of this - Android has some particular places you are allowed to access files (or where the NDK goes looking), so some of the decisions I made were based on making it work with the NDK - certainly open to using GNUstep mechanisms to achieve the same things if Android will allow it and we can refine the approach.

DTW-Thalion:
@rfm, I think you'll prefer this:

The new method is removed, and the branch now uses the mechanism you suggested.

dladdr, called inside a packaged application, names the file the library was
loaded from, and the parent of that file is the directory
ApplicationInfo.nativeLibraryDir reports. Both readings come from one run:

```
dli_fname  = /data/app/~~T-Embn…==/org.gnustep.gsapp-whwyp…==/lib/x86_64/libgnustep-base.so
via loader = /data/app/~~T-Embn…==/org.gnustep.gsapp-whwyp…==/lib/x86_64
via JNI    = /data/app/~~T-Embn…==/org.gnustep.gsapp-whwyp…==/lib/x86_64
```

So GSPrivateSymbolPath([NSProcessInfo class]) finds the directory with no JNI
context, which is what NSPathUtilities.m already does for a config file
location beginning './'. NSSearchPathForDirectoriesInDomains() then answers a
directory a tool can be started from, and +launchPathForTool: needs no Android
branch of its own.

A filesystem layout cannot carry the answer. The path holds two randomised
strings, and both change on every install and update; four reinstalls in one
session produced four different paths. A GNUstep.conf cannot be anchored there
either, because only files matching lib*.so under lib/<abi>/ are installed into
that directory. The location has to be worked out at run time.

The directory is added ahead of the layout's directories rather than replacing
them, and both halves of that were settled by testing rather than by reading.
Replacing them broke a test binary run from a shell, which loads the library
from one directory and finds its tools in another. Appending broke the packaged
case, where a tool of the same name in the writable directory matched first and
could not be started.

One thing no path mechanism covers is the file name. A package carries gdnc as
libgdnc.so, so executablePath() tries that form as it tries the extensions
Windows adds; the running process is named libgdnc.so.

On the environment: a tool started with execve is given the platform's linker
search path, which does not include the package's own directory, so
LD_LIBRARY_PATH is set to it. TMPDIR is set because a port is named within the
temporary directory, and a tool working out its own would name a different one,
so the two would never meet.

rfm (approved):
OK, I guess I'm convinced.  The deliberate restrictions in Android do make life complicated.

PR #750, opened 2026-08-11
Let a thread reuse the key value slots it has already looked up

DTW-Thalion:
valueForKeyWithCaching takes a static mutex on every call, to read a table that is only ever added to, so every valueForKey: in a process serialises on one lock whatever object it is sent to.

A slot is added to that table once and never removed, so a pointer to one stays valid and a thread can keep the last few it used. A thread that finds the slot it wants takes no lock. The flag saying the per thread key exists is read outside the lock that creates it, so it is read and written atomically.

ns per -[NSObject valueForKey:] on distinct objects, one per thread, 32 cores, median of 3 interleaved runs: 1 thread 33.2 to 25.0, 8 threads 829.1 to 34.5, 24 threads 1860.3 to 55.0.

KVC, NSKVOSupport, NSObject, NSPredicate and NSSortDescriptor give 932 passed before and after. A 16 thread probe over 3 classes and 3 keys checks every value returned and finds no mismatch, and ThreadSanitizer reports nothing here.

Two things to note: it rests on slots never being removed, which nothing enforces, and the update at the end of the function still writes over a slot another thread may be reading, which is there already.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,13 @@
 2026-08-11 Todd White <todd.white@thalion.global>
 
+	* Source/NSKeyValueCoding+Caching.m: Let each thread keep the
+	slots it has most recently used, so that valueForKey: reads the
+	shared table, and takes its lock, only for a class and key that
+	thread has not seen.  A slot is added to the shared table once
+	and never removed, so a pointer to one stays valid.
+
+2026-08-11 Todd White <todd.white@thalion.global>
+
 	* Source/NSCalendarDate.m: Read the second that GSPrivateTimeNow
 	records before writing it, and write it only when it has
 	changed, so that every thread asking for the time does not
--- a/Source/NSKeyValueCoding+Caching.m
+++ b/Source/NSKeyValueCoding+Caching.m
@@ -87,6 +87,31 @@ static void inline _KVCCacheSlotRelease(const void *ptr)
 #import "GNUstepBase/GSIMap.h"
 #import "GSPThread.h"
 
+/* A slot is added to the shared table once and never removed from it, so a
+ * pointer to one stays valid for the life of the process and each thread can
+ * keep a few of them to itself.  A thread that finds the slot it wants here
+ * reads nothing the other threads write.
+ */
+#define KVC_THREAD_CACHE_SIZE 8
+
+struct _KVCThreadCache
+{
+  struct _KVCCacheSlot *slots[KVC_THREAD_CACHE_SIZE];
+};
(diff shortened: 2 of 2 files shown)


DTW-Thalion:
@triplef this one is up your alley as it is KVC and KVO heavy code, so the numbers may be worth a look.

Ordinary application work rather than the micro benchmark above: each thread sorts its own array of 500 model objects with two NSSortDescriptors and then filters it with an NSPredicate, both of which resolve their keys through valueForKey:. Milliseconds per round, before and after: 

- 1 thread 0.699 to 0.661, 
- 4 threads 1.515 to 0.668, 
- 8 threads 6.655 to 0.711, 
- 16 threads 19.038 to 1.072, 
- 24 threads 22.880 to 1.246.

triplef:
Wow super interesting! I’ll have to try it out with our app. Do you know if Apple’s KVO does something similar, or any idea how numbers compare to Apple platforms?

DTW-Thalion:
I don't know how this would compare on apple platforms (or how Apple does this) - this was a pure valgrind audit of the libs-base code to find hotspots, and I happened across this. In my case, we're doing genomic data processing so I am highly focused on mutli-threaded and GPU operatons - so that is the lense I am using when assessing these.

DTW-Thalion:
@triplef  did some more digging: Apple does not serialise these lookups either, so this brings GNUstep to the same shape .

The same benchmark on a macOS runner, which reports 3 cores, so only the 1 and 2 thread rows are free of oversubscription: 1 thread 1.480 ms per round, 2 threads 1.684, a rise of 14 percent.

GNUstep before the change, on a 32 core box where 8 threads still leave cores idle: 0.688, 0.900, 1.412 and 4.042 ms at 1, 2, 4 and 8 threads, so 5.9 times the single thread cost by 8 threads. After the change: 0.652, 0.655, 0.657 and 0.695, a rise of 7 percent at 8 threads.

The absolute figures come from two different machines and are not worth comparing. The shape is, and the shape says the lock was the difference. On what Apple's implementation actually does I can only report what the scaling shows, not the mechanism.

triplef:
Thanks for digging in further. I tried out the branch with our app, but I can’t see a measurable performance difference for our use case. We’re using KVO and KVC almost exclusively on the main thread, for which your numbers predict only a minimal improvement, so it sounds like this would be expected.

rfm (approved):
This all looks reasonable, and since Frederik has found no issues, I'm happy to merge.

PR #752, opened 2026-08-12
Look up a localized string without the lock every bundle shares

DTW-Thalion:
_localizationsLock is one static mutex for the whole process, and -localizedStringForKey:value:table: took it twice on every call, with the lookup of the key inside the second, so a localized string in one bundle waited on a localized string in another. -infoDictionary took no lock at all, and -cleanPathCache destroyed both it and the localization tables outside the lock.

A thread now keeps the table it was last given for a bundle and table name, holding a retain on it, and a counter incremented under the lock by the load of a table and by a clean makes an earlier entry unusable. A lookup which finds its table takes no lock. The information dictionary is read under the same lock, which -cleanPathCache now takes, so it can discard whatever it likes.

ns per lookup with the returned string checked on every call, 32 cores, medians of 10 interleaved runs, every thread in one bundle: 1 thread 431.8 to 397.9, 8 threads 12833.4 to 415.5, 24 threads 81486.4 to 867.1. A bundle per thread gives the same.

The whole suite gives 0 failures and 44 dashed hopes before and after. No instance variable is added.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,14 @@
 2026-08-11 Todd White <todd.white@thalion.global>
 
+	* Headers/Foundation/NSBundle.h:
+	* Source/NSBundle.m: Publish the map of localization tables in a
+	bundle rather than changing it in place, so that a lookup by
+	-localizedStringForKey:value:table: reads the map without taking
+	the lock that every bundle in the process shares.  The lock is
+	taken only where a table is added.
+
+2026-08-11 Todd White <todd.white@thalion.global>
+
 	* Source/NSCalendarDate.m: Read the second that GSPrivateTimeNow
 	records before writing it, and write it only when it has
 	changed, so that every thread asking for the time does not
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,11 +1,12 @@
-2026-08-11 Todd White <todd.white@thalion.global>
+2026-08-12 Todd White <todd.white@thalion.global>
 
-	* Headers/Foundation/NSBundle.h:
-	* Source/NSBundle.m: Publish the map of localization tables in a
-	bundle rather than changing it in place, so that a lookup by
-	-localizedStringForKey:value:table: reads the map without taking
-	the lock that every bundle in the process shares.  The lock is
-	taken only where a table is added.
+	* Source/NSBundle.m: Let a thread reuse the localization table it
+	was last given for a bundle and table name, holding a retain on
+	that table, so that -localizedStringForKey:value:table: takes the
+	lock every bundle shares only where a table is loaded.  Protect
+	-infoDictionary with the same lock, and take it in
+	-cleanPathCache, which destroyed the information dictionary and
+	the localization tables without it.
 
 2026-08-11 Todd White <todd.white@thalion.global>
 
--- a/Headers/Foundation/NSBundle.h
+++ b/Headers/Foundation/NSBundle.h
@@ -117,7 +117,7 @@ GS_EXPORT_CLASS
   NSMutableArray	*_bundleClasses;
   Class			_principalClass;
   NSDictionary		*_infoDict;
-  NSDictionary		*_localizations;
+  NSMutableDictionary	*_localizations;
   unsigned		_bundleType;
   BOOL			_codeLoaded;
   unsigned		_version;
--- a/Headers/Foundation/NSBundle.h
+++ b/Headers/Foundation/NSBundle.h
@@ -117,7 +117,7 @@ GS_EXPORT_CLASS
   NSMutableArray	*_bundleClasses;
   Class			_principalClass;
   NSDictionary		*_infoDict;
-  NSMutableDictionary	*_localizations;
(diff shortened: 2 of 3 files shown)


rfm:
There's obviously no need to use a single mutex.  However, other stuff I'm not so sure about.  I think this has exposed a bug in the existing code:
We have a -cleanPathCache method which is clearly supposed to force information to be re-fetched from disk in case of a change, yet it does not currently grab the lock when it destroys _infoDict and _localizations.
I think all this data should be discarded safely in -cleanPathCache, which implies that it should all be appropriately lock protected (but with more fine grained locking).
In practice, we get away with it, because cleaning the cache is rare and the chance of another thread using it while it's being cleaned is therefore very low,  but I expect it causes an occasional crash.

DTW-Thalion:
Thinking about this a bit:

On finer grained locking, I built a quick harness rather than guess: a lock per bundle, chosen by the bundle instead of shared by all of them, measured against master and against the published map this PR currently has. 

Median of 3 interleaved runs on a 32 core machine, ns per lookup on the hit path, with the returned string checked on every iteration. 

Every thread reading one bundle:

threads	1	4	8	16	24
master	429.9	4769.1	12401.1	41793.2	81314.5
a lock per bundle	421.4	5365.8	12516.5	46491.8	77861.6
no lock on the read path	404.5	408.8	426.8	747.9	902.4

A bundle per thread:

threads	1	4	8	16	24
master	419.9	4068.4	14146.8	48788.9	90044.1
a lock per bundle	424.1	429.5	566.3	870.5	985.6
no lock on the read path	399.5	407.3	604.4	797.8	921.0

A lock per bundle answers the second table and leaves the first where it was. An application reads the main bundle, a few frameworks and the theme, so its threads are usually in the same bundle, and that is the case which does not move the numbers.

The published map does not meet what you asked for,  and I am dropping it. It would replace the map rather than freeing it, so a reader can still be holding an old one, which is the opposite of discarding the data in cleanPathCache.

What I would like to put in its place: each thread keeps the table it last used for a given bundle and table name, holding a retain on it, and a counter which cleanPathCache increments makes an entry from before a clean unusable. 

A lookup that finds its table takes no lock and does one dictionary lookup. A lookup that does not takes the existing lock and loads the table as now. All the shared state stays under that lock, cleanPathCache takes it and destroys whatever it likes, and a table another thread is still reading stays alive because that thread holds a retain on it, which also removes the freed string above. No instance variable is added, so there is no ABI question, and the single mutex is then only taken to load a table or to clean, which is rare enough that splitting it stops mattering.

Two questions. Does that count as appropriately lock protected for what you had in mind? And do you want infoDictionary in the same change, building _infoDict under the lock and returning it retained and autoreleased?

DTW-Thalion:
Two details I left out above. infoDictionary at NSBundle.m:3226 builds _infoDict with no lock at all, and cleanPathCache at :3531 destroys both _infoDict and _localizations outside any lock, so the discard is not the only unprotected part of it.

The string localizedStringForKey: answers is owned by the table it came from, so after a clean a caller can be left holding a freed string. The retain each thread would hold on its table removes that as well.

rfm:
I like that proposal, and yes I think the info dictionary should be covered too: my understanding is that the cleanup method is intended to reset the bundle so that all it's resources get updated from disk the next time anything wants them, and that clearly *should* be thread safe.

DTW-Thalion:
That is what the branch does now, and the body is rewritten to match.

A thread keeps the table it was last given for a bundle and table name, holding a retain on it. A counter, incremented under the lock by the load of a table and by cleanPathCache, makes an entry from before the change unusable. A lookup which finds its table takes no lock; one which does not takes the lock and loads as before. infoDictionary builds _infoDict outside the lock and installs it under it, answering it retained and autoreleased, and cleanPathCache takes the lock around both destroys. The ivar goes back to NSMutableDictionary, so the header is unchanged and the published map is gone.

ns per lookup with the returned string checked on every call, 32 cores, medians of 10 interleaved runs, every thread in one bundle: 

- 1 thread 431.8 to 397.9, 
- 4 threads 4098.6 to 402.0, 
- 8 threads 12833.4 to 415.5, 
- 16 threads 43478.6 to 700.7, 
- 24 threads 81486.4 to 867.1. 

A bundle per thread gives the same numbers.

There was an occasional crash that I was able to reproduce. A probe runs 8 threads looking up a string and taking the information dictionary while another thread rewrites the strings file and calls cleanPathCache 200 times. On master it dies within seconds: `GSFFIInvocation: Class 'nil'(class) does not respond to forwardInvocation: for 'isEqualToString:'`, then a segmentation fault. With 2 readers it survives but answers one string that was never written. With this change the same probe completes 61215 reads, every answer one of the values written, and the information dictionary never nil.

One case this does not change: while a thread loads a table, the empty table is in the cache under that name, so other threads are answered the fallback value until the load finishes. The same probe counts 424 of those in 28544 reads on master with a single reader.

Suite: 0 failures and 44 dashed hopes both ways.

rfm (approved):
Looks good.

PR #754, opened 2026-08-12
Answer a method signature without a lock and without retaining it

DTW-Thalion:
+signatureWithObjCTypes: took a static mutex on every call and answered RETAIN of the cached signature, so a caller queued for one lock and then wrote to the reference count of one shared object.

A set of types seen before is answered from a table of published entries, read with an acquire and written with a release. A cached signature is kept for the life of the process, so its retain, release and autorelease do nothing. The cache holds at most 4096 sets of types; past that a signature is built, answered autoreleased and not cached. Types decoded from a connection or an archive take that uncached path.

ns per fetch, medians of 3 interleaved runs: 1 thread 70.7 to 55.5, 8 threads 1778.1 to 59.8, 24 threads 6805.6 to 104.6. The ARC path at 24 threads is 7003.0 to 135.4.

A probe building 200000 sets of types past the limit leaves resident memory where it was; master grows by 40 MB.

Replacing an entry orphans 16 bytes, bounded by the limit at 48 KB. Over releasing a cached signature is not noticed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,12 @@
 2026-08-11 Todd White <todd.white@thalion.global>
 
+	* Source/NSMethodSignature.m: Answer a set of types that has been
+	seen before from a table read without the cache lock, and keep
+	every signature for the life of the process rather than counting
+	references to it, which is what its cache already does.
+
+2026-08-11 Todd White <todd.white@thalion.global>
+
 	* Source/NSCalendarDate.m: Read the second that GSPrivateTimeNow
 	records before writing it, and write it only when it has
 	changed, so that every thread asking for the time does not
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,12 +5,19 @@
 	something in it.  The array is local to the call, so its count
 	may be read without the lock.
 
-2026-08-11 Todd White <todd.white@thalion.global>
+2026-08-12 Todd White <todd.white@thalion.global>
 
-	* Source/NSMethodSignature.m: Answer a set of types that has been
-	seen before from a table read without the cache lock, and keep
-	every signature for the life of the process rather than counting
-	references to it, which is what its cache already does.
+	* Source/NSMethodSignature.m:
+	* Source/GSInvocation.h:
+	* Source/NSDistantObject.m:
+	* Source/NSInvocation.m: Answer a set of types that has been seen
+	before from a table read without the cache lock, and keep a
+	cached signature for the life of the process rather than counting
+	references to it, which is what the cache already does.  Limit how
+	many sets of types may be cached, answering an uncached signature
+	past the limit rather than failing, and use an uncached signature
+	for types which came from a connection or an archive, so that a
+	peer cannot fill the cache.
 
 2026-08-11 Todd White <todd.white@thalion.global>
 
--- a/Source/GSInvocation.h
+++ b/Source/GSInvocation.h
@@ -67,6 +67,11 @@ typedef struct	{
 @end
 
 @interface NSMethodSignature (GNUstep)
+/* Answers a signature which is not put in the cache and is deallocated like
+ * any other object.  For types which came from outside the process, so that
+ * a peer or an archive cannot fill the cache.
+ */
++ (NSMethodSignature*) _uncachedSignatureWithObjCTypes: (const char*)t;
 - (const char*) methodType;
 - (NSArgumentInfo*) methodInfo;
 @end
--- a/Source/NSDistantObject.m
+++ b/Source/NSDistantObject.m
@@ -764,9 +764,10 @@ GS_ROOT_CLASS @interface	GSDistantObjectPlaceHolder
 	      const char	*types;
 
 	      types = [m methodType];
-	      /* Create a local method signature.
+	      /* Create a local method signature.  The types came from the
+	       * other end of the connection, so it is not cached.
 	       */
-	      m = [NSMethodSignature signatureWithObjCTypes: types];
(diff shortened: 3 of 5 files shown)


rfm:
I think this code was considered poor and waiting for a fix, pending on deciding how best to flush old items from the cache so that it would not grow forever if an attacker (or some strange/stupid code) kept requesting different signatures.
If we consider it worthwhile for performance to never shrink the cache, then we probably need some other way to protect it from excessive growth.  The only thing I can think of that seems workable is a fixed cache size limit, but then what would we do when the limit is reached?  Possibly log an error message and return nil?

DTW-Thalion:
If I have latitude to change the design a bit, then I can probably come up with something that is fairly workable. I don't think you can make it a fixed cache size without running into risks of corrupting processes  (I would be worried about it in our workflows with genomes),  but there may need to be a better structure for reclaiming the entries.

rfm:
By fixed size, I meant a limit on how many signatures could be added: on the basis that sane programs don't keep on using new signatures that haven't been seen before.
  

  

  
  
  
  
  
>

DTW-Thalion:
The growth is worth separating from this change. Master already keeps every signature for the life of the process: `_initWithObjCTypes:` is called from one place, inside the cache, and the cache never removes an entry, so what this patch would put in front of that map is a fixed table of 1024 slots. 

I put an instrumented build counting insertions over the whole base test suite: 82 of about a thousand test programs cache any signature at all, the busiest single process caches 18, and the total across the suite is 173. A distinct 12 argument signature costs about 236 bytes, so 100000 of them cost 23 MB resident.

Two paths let a type string from outside the process into the cache, both of them on master today. `-[NSConnection typeForSelector:remoteTarget:]` decodes a `char*` off the wire at NSConnection.m:2197 and NSDistantObject.m:769 hands a remote proxy's `methodType` to `+signatureWithObjCTypes:`. `-[NSInvocation initWithCoder:]` at NSInvocation.m:759 decodes a type string from any archive and does the same.

On what to do at a limit: nil is hard for callers to survive, since the result feeds forwarding and NSInvocation, and a DO server would answer every request with an exception instead of growing. Answering an ordinary autoreleased signature which is not entered in the cache is master's behaviour minus the cache, so nothing observable changes and memory stays bounded. A cap in the low thousands leaves a lot of headroom over the 18 the busiest test program uses.

On whether the cache needs immortal signatures at all, ns per `-methodSignatureForSelector:` fetched on every call, median of 3 interleaved runs on 32 cores:

| threads | 1 | 4 | 8 | 24 |
|---|---|---|---|---|
| master, caller managing its own references | 72.0 | 630.9 | 2098.5 | 5719.9 |
| this patch | 54.2 | 55.7 | 58.4 | 111.4 |
| this patch without immortality | 55.1 | 56.0 | 58.7 | 110.4 |
| master, ARC caller | 55.0 | 470.0 | 1500.4 | 6344.9 |
| this patch, ARC caller | 66.2 | 68.9 | 74.2 | 142.4 |
| this patch without immortality, ARC caller | 64.7 | 220.6 | 520.1 | 1908.0 |

Immortality gains nothing for a caller that manages its own references, because the cache holds the only reference that matters and the fetch can simply not retain. For an ARC caller, which retains and releases the returned object itself, the difference is 13 times at 24 threads.

So the shape I would like to build, if it seems the right one: cap the number of cached signatures, answer an uncached autoreleased signature past the cap rather than nil, and give the connection and coder paths that uncached constructor so a remote peer or an archive never consumes the cap at all. Immortality then belongs to a cached instance rather than to the class, which the uncached ones can get by being a private subclass, so no instance variable is added and nothing changes for a program that stays inside the cap.

rfm:
That sounds very good. In particular using the uncached route for distributed objects and archives makes sense because the related I/O means that the slower speed of the uncached code is relatively unimportant; a point I hadn't noticed before.

DTW-Thalion:
Built, and the body is rewritten to match.

The cache now holds at most 4096 sets of types. Past that, +signatureWithObjCTypes: builds a signature, answers it autoreleased and does not cache it. An uncached signature is a private subclass which counts references as any other object does, so the retain, release and autorelease which do nothing belong to a signature the cache keeps rather than to the class. No instance variable is added. 4096 is a guess at the right order and easy to change: the busiest program in the base test suite caches 18.

+_uncachedSignatureWithObjCTypes: is declared in GSInvocation.h and used at NSDistantObject.m:769 for the types a proxy answers and at NSInvocation.m:760 for the types an archive holds.

A probe builds 200000 distinct sets of types past the limit, checking each answers a working signature. Resident memory: 21380 to 21380 KB on this branch, 21900 to 62420 KB on master, which also answers the same object twice for types it has never been asked for before and caches the signature an archive names. Below the limit the same set of types still answers one object.

ns per fetch, medians of 3 interleaved runs, 1/4/8/24 threads: 

- master 70.7, 637.0, 1778.1, 6805.6; 
- this branch 55.5, 56.1, 59.8, 104.6. 
- Through the ARC path master 77.1, 691.7, 1569.2, 7003.0 
- and this branch **68.3, 68.6, 72.3, 135.4.** 

Unchanged from the version without the limit, which is the point.

One cost to name: publishing an entry where the table already has one for that index orphans 16 bytes, since a reader may hold the old entry. That is bounded by the limit, 48 KB for the life of the process, and memcheck over a run past the limit reports those and nothing else.

Suite from this tree: 13467 passed, 44 dashed hopes, 0 failures. The unpatched run of the same tree aborted NSStream/socket.m; the reworked run did not.

rfm (approved):
Great.  A substantial improvement over the older code :-)

PR #759, opened 2026-08-13
Fix app bundle resource lookup on Android

triplef:
`+[NSBundle initialize]` looks the Locale maps up using the gnustep-base library bundle's path, which is nil on Android because the library resources are not installed in a library domain there.  Since #735 that lookup is a `GSPrivateResourcePath()` call, which searches the main bundle when it is given a nil root path (the earlier `-pathForResource:` was sent to a nil bundle and did nothing).

The main bundle's resources are Android assets and are unreadable until `+setJavaAssetManager:withJNIEnv:` has run, so the lookup failed and `bundle_directory_readable()` cached the failure.  That cache is only undone by `-cleanPathCache`, which is sent to `_mainBundle` and did nothing because no main bundle had been created yet, leaving the process unable to find any resource at all: Info.plist, localizations and the CA certificate bundle were all missing for the rest of the run.

Move the map loading into `loadLanguageMaps()` and call it on Android from `+setJavaAssetManager:withJNIEnv:`, once the assets are readable, instead of from `+initialize`.  Searching the main bundle when there is no library bundle is the intended fallback there, so the maps are also found where the library resources are packaged as application assets.

Discard the whole path cache, rather than the main bundle's entries, once the asset manager is available, and create the main bundle at that point, since assets are located relative to its resource path.

### DIFF
--- a/Source/NSBundle.m
+++ b/Source/NSBundle.m
@@ -1698,6 +1698,63 @@ GSPrivateInfoDictionary(NSString *rootPath)
   return info;
 }
 
+/* Load the language alias and canonical maps used by altLang().
+ *
+ * The maps are resources of the gnustep-base library bundle, but that bundle
+ * is not always present (on Android the library resources are packaged as
+ * application assets rather than being installed in a library domain), so we
+ * fall back to searching the main bundle for them.
+ *
+ * NB. This must not be called before the resources of the bundle being
+ * searched are readable: on Android that is only the case once the asset
+ * manager has been set, which is why +initialize does not call this there.
+ */
+static void
+loadLanguageMaps(void)
+{
+  NSString	*rootPath;
+  NSString	*file;
+
+  /* A nil rootPath makes GSPrivateResourcePath() search the main bundle,
+   * which is what we want when there is no library bundle.
+   */
+  rootPath = [_gnustep_bundle bundlePath];
+
+  /* The Locale aliases map converts canonical names to old-style names
+   */
+  file = GSPrivateResourcePath(@"Locale", @"aliases",
+    rootPath, @"Languages", NOT_LOCALIZED);
+  if (file != nil)
+    {
+      NSDictionary  *d;
+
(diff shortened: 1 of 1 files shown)


triplef:
@DTW-Thalion could you check if this makes sense to you please?

DTW-Thalion:
Had a read through this, looks pretty good, though I have a few thoughts:

GSPrivateResourcePath() falls back to the main bundle on a nil root path, at NSBundle.m:1692, so #735 turned a message to a nil bundle into a real search. bundle_directory_readable() caches a failure deliberately, storing [NSNull null] and translating it back on read, so the failure is sticky. cleanPathCache only removes entries with hasPrefix: _path, and sent to a nil _mainBundle it does nothing at all. The chain holds.

In the fix, the language maps look worth another pass. langAliases and langCanonical are plain statics read by altLang() with no lock, and loadLanguageMaps() ASSIGNs both, which releases the previous dictionary. 

Under +initialize that write is single threaded and safe by construction - all good. From the asset manager call it happens at an arbitrary later point, so a concurrent altLang() can be left messaging a released object. Setting the asset manager probably happens before any other thread runs, but that becomes an assumption the code rests on without stating it, and the same applies if the method can be called more than once. Not sure how much concurrency is an issue on Android?

A few smaller things:

- removeAllObjects clears the entries of every bundle, while only the main bundle's _infoDict and _localizations are dropped with it. The generation bump covers the per thread localization memo, so what is left is those two ivars on any other instance. Is another bundle reachable that early on Android?

- The maps now load only from the asset manager call. Is there an Android entry point that never reaches it, a packaged command line tool for instance? Not a regression, since the load failed there before as well, but the hardcoded defaults from +initialize would become the permanent answer.

- [[self mainBundle] cleanPathCache] runs just after the cache was emptied, so its path removal half discards whatever creating the main bundle had populated. Harmless, and the comment already covers why the call is there.

- Ordering loadLanguageMaps() after the main bundle exists looks right, since the nil root path fallback needs it, and pathCacheLock is released before mainBundle is asked for, so nothing nests.

Issue #758, opened 2026-08-12
NSNotificationCenter: one lock for the whole table


DTW-Thalion:
-_postAndRelease: holds the table's single lock across the whole collection phase, so posts of unrelated names serialise against each other. Measured on a 32 core machine, ns per post, medians of interleaved runs, every thread posting its own name with one observer: master 33200 at 24 threads, and 2372 with the lock removed altogether. With the lock chosen by [name hash] across 1024 stripes it is 2409, so splitting by name reaches that ceiling.

Where every thread posts the same name it buys nothing: 23910 stock against 19245 striped, where removing the lock gives 2614.

A correct split is more than striping. The NAMED map is itself mutated by a post, since a node whose observation list empties is removed and freed; the observation allocator and the map cache are shared; and addPost writes the retained receiver into each observation. So the add and remove paths need covering too, not just the post.

Before writing any of that: is a lock per name a direction worth taking, or should the table stay under one lock?

Two cheaper shapes were measured and are not worth proposing: pthread_spinlock gives 13121 and PTHREAD_MUTEX_ADAPTIVE_NP 16295, neither portable, and a bounded trylock spin through GS_MUTEX_TRYLOCK measured 61431, worse than master, since every waiter writes the lock line.

rfm:
>>>Before writing any of that: is a lock per name a direction worth taking, or should the table stay under one lock?

I suspect there's no good answer.  The issue here is lock contention between threads, so we aren't going to help predominantly single threaded processes.  Also, we aren't going to help situations where there's no significant amount of notifications in different threads, or where we have lots of notifications with the same name, so we are reduced to  cases where we have multithreaded applications making heavy use of notifications with different names, and my feeling is that there aren't many of those (maybe none).

That being said, this would make a huge difference to that one app that was designed for lots of different notifications in different threads (and I can see that would sometimes be a very appealing design model).
So I think the key considerations are:
can we do it with no/negligible penalty to single threaded code?
can we do it without degrading maintainability;  clear/simple and easy to understand, localised change so we don't impede structural design etc?
I think the answer to those is probably yes, so I don't see why we shouldn't be doing this.

DTW-Thalion:
Built as #760. On 24 threads, each posting its own name, a post costs 1203ns against 41108ns for master in the same rounds. A single post costs 2393 instructions against 2514, so the single threaded case improves rather than paying for it.

[...]

PR #760, opened 2026-08-13
Post under a lock chosen by the notification name

DTW-Thalion:
A post holds the one lock the whole table shares while it collects, so posts of unrelated names serialise. On 24 threads, each posting its own name, a post costs 41108ns against 1203ns.

The lock becomes 32 recursive stripes. lockNCTable takes all of them in index order, so it means what the single lock meant and no other caller changes. A post takes the stripe its name falls in. The outer map of names is only read by a post and only written under every stripe, so posts on different stripes only read it.

Observations registered with no name are consulted whatever the posted name, so they keep a lock of their own, taken only where one exists. Then a post costs 3019 instructions against 2931; otherwise 2393 against 2514. The shared free list is guarded where it is pushed.

Single thread figures are callgrind, since this machine is bimodal below 4 threads.

The suite passes 13467 with 44 dashed hopes and no failures. A 16 thread run with named, wildcard and nameless observers loses no post, and catches an unsafe variant that does. #758 asks about this change.


rfm (approved):
Thanks for the optimisation.

PR #761, opened 2026-08-13
Post NSUserDefaultsDidChangeNotification when a lookup changes

DTW-Thalion:
Nothing is posted when these methods alter what objectForKey: can see: registerDefaults:, setVolatileDomain:forName:, removeVolatileDomainForName:, addSuiteNamed:, removeSuiteNamed: and setSearchList:. Everything else that changes the mapping posts already, through _changePersistentDomain: or its own call.

Each now posts, in the shape _changePersistentDomain: already uses, a flag set under the lock and the post made after unlocking.

setSearchList: keeps its existing isEqual: guard, so an equal list still posts nothing. The rest post unconditionally, matching the note in setObject:forKey: about notifying even where the value has not changed. The registration domain that resetStandardUserDefaults reinstalls on the new shared instance is not covered here.

Tests/base/NSUserDefaults/notifications.m exercises all of them.

On master 6 of its 8 assertions fail, one for each method, and all 8 pass with the change. The suite passes 13488 with 44 dashed hopes and no failures. #756 discusses this.


rfm (approved):
Thanks.

PR #762, opened 2026-08-13
Initial NSURLSessionWebSocketTask skeleton

HendrikHuebner:
I have been working on a (mostly complete) implementation of `NSURLSessionWebSocketTask` using libcurl's websocket API.

It's a lot of code (~2500 Additions + tests) and I dare say more complex than the original NSURLSession because of the lifecycle management of the websocket session, so I'm breaking it up into several smaller patches.

One major caveat is that I uncovered bugs in libcurl while working on this (See https://github.com/curl/curl/issues/22413), so it will only be usable with the next curl release. Otherwise users might see internal assertion failures due to curl serving incorrectly chunked WebSocket frames.


rfm:
While this looks very welcome (I'd definitely like to have it), the NSURLSession code is currently in flux, being changed to make it portable (not depend on clang), so I think we will want to wait a little before attempting to resolve differences and merge.  Perhaps look again in a week or two?

HendrikHuebner:
@rfm I'll only be able to work on this until the end of August. We should try to get it upstreamed until then, otherwise someone else will have to work on finalizing it.

rfm:
I'm not against finalising the portability work at the same time as incorporating websocket support, as long as you are ok with reading your work and resolving conflicts.
As far as the curl bug is concerned, I suppose we would need to gate the websocket methods in the header on the curl version detected at configure time.  The other option would be a fallback to use the more mature libwebsocket with older versions of curl, but that would be a lot more work.

triplef:
@HendrikHuebner there’s a couple errors building this with GCC on Windows CI. I think we need to use macros like `GS_GENERIC_CLASS` for generics and `_Nullable` instead of `nullable`.

HendrikHuebner:
@rfm I'm submitting the remaining commits for the NSURLSessionWebsocketTask later.

rfm:
Thanks for letting me know.  Is the new PR expected to build on this one or replace it?
I have some slight concern about this one: the workflow to test it does not seem to be operating when requested (so we don't know if this code compiles and doesn't break any testcases on the various platforms), and I don't know why that is :-(

triplef:
I _think_ this PR can be closed in favor of #775, but I’ll let @hmelder make that call as he’s taking over the project from @HendrikHuebner.

hmelder:
> I _think_ this PR can be closed in favor of #775, but I’ll let @hmelder make that call as he’s taking over the project from @HendrikHuebner.

Yep this can be closed.

PR #763, opened 2026-08-14
Build the NSURLSession implementation without blocks

### LINKED ISSUE #482: NSURLSession implementation needs to be portable between compilers

DTW-Thalion:
Closes #482.  Three things stopped NSURLSession building with gcc.

The implementation created 25 block literals. 11 scheduled work on the session
work thread and 11 a delegate message on the delegate queue; those go through a
selector or an NSInvocation now. The other 3 built a completion handler for a
delegate method, so the delegate replies by messaging the task instead, with
resumeWithRedirectRequest:, resumeWithResponseDisposition: and
resumeWithBodyStream:.

The instance variables inside the @implementation blocks need the non-fragile
ABI, so both classes hold them through GSInternal.h now.

gcc has no _Atomic in Objective-C, so the counters and flags go through
GSAtomic.h. Its native branch uses clang builtins and MinGW gcc was reaching
it, so that branch now tests for clang as well.

allTasks and tasksOfKind: replace the two getTasks methods. Those, and the
three delegate methods taking a handler, carry GS_DEPRECATED. The convenience
methods creating a task with a handler have no replacement of their own, so
they keep working as they are.

The tests no longer use blocks outside the 4 covering the handler API. They
pass 133 assertions with clang and 102 with gcc.

### DIFF
--- a/Headers/Foundation/NSURLSession.h
+++ b/Headers/Foundation/NSURLSession.h
@@ -225,14 +225,14 @@ GS_NSURLSession_IVARS;
  * handler is called before this method returns.
  */
 - (void) getTasksWithCompletionHandler:
-  (GSNSURLSessionTasksCompletionHandler)completionHandler;
+  (GSNSURLSessionTasksCompletionHandler)completionHandler GS_DEPRECATED;
 
 /**
  * Deprecated: use -allTasks instead.  The completion handler is called
  * before this method returns.
  */
 - (void) getAllTasksWithCompletionHandler:
-  (GSNSURLSessionAllTasksCompletionHandler)completionHandler;
+  (GSNSURLSessionAllTasksCompletionHandler)completionHandler GS_DEPRECATED;
 
 /**
  * This serial NSOperationQueue queue is used for dispatching delegate messages
@@ -660,7 +660,8 @@ GS_EXPORT_CLASS
   task: (NSURLSessionTask *)task
   willPerformHTTPRedirection: (NSHTTPURLResponse *)response
   newRequest: (NSURLRequest *)request
-  completionHandler: (GSNSURLSessionRedirectHandler)completionHandler;
+  completionHandler: (GSNSURLSessionRedirectHandler)completionHandler
+  GS_DEPRECATED;
 
 /* An HTTP request is attempting to perform a redirection to a different URL.
  * Reply by sending -resumeWithRedirectRequest: to the task, with the request
@@ -679,7 +680,8 @@ GS_EXPORT_CLASS
  */
 - (void) URLSession: (NSURLSession *)session
   task: (NSURLSessionTask *)task
-  needNewBodyStream: (GSNSURLSessionBodyStreamHandler)completionHandler;
+  needNewBodyStream: (GSNSURLSessionBodyStreamHandler)completionHandler
+  GS_DEPRECATED;
 
 /* A new body stream is needed to continue the upload.  Reply by sending
  * -resumeWithBodyStream: to the task.  The reply may be sent at any time and
@@ -705,7 +707,8 @@ GS_EXPORT_CLASS
 - (void) URLSession: (NSURLSession *)session
   dataTask: (NSURLSessionDataTask *)dataTask
   didReceiveResponse: (NSURLResponse *)response
-  completionHandler: (GSNSURLSessionResponseDispositionHandler)completionHandler;
+  completionHandler: (GSNSURLSessionResponseDispositionHandler)completionHandler
+  GS_DEPRECATED;
 
 /** Informs the delegate of a response.  This message is sent when all the
  * response headers have arrived, before the body of the response arrives.
--- a/Headers/Foundation/NSURLSession.h
+++ b/Headers/Foundation/NSURLSession.h
@@ -37,6 +37,7 @@
 #import <Foundation/NSOperation.h>
 #import <Foundation/NSProgress.h>
 #import <Foundation/NSDate.h>
+#import <GNUstepBase/GSBlocks.h>
 
 #if GS_HAVE_NSURLSESSION
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_9, GS_API_LATEST)
@@ -46,6 +47,7 @@
 
 @class NSError;
 @class NSHTTPURLResponse;
+@class NSInputStream;
 @class NSOperationQueue;
 @class NSURL;
 @class NSURLAuthenticationChallenge;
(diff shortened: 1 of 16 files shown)


rfm (approved):
Thanks a lot for doing this.  I attempted it once myself, and threw away my changes after messing things up, so it's really great to see it working.

There is some tidying up to do, but I want to merge this and work on details later:
In headers, wrap new methods in macros to show they are GNUstep additions.
Use GS_NON_PORTABLE rather than GS_DEPRECATED,
Maybe add a category in the additions library to provide portable methods on OSX for anyone wanting to write more portable code there.

DTW-Thalion:
Let me know what you want for clean up, and I'll pick away at it over the next week or so.

rfm:
I think I have finally managed to fix all the leaks associated with this (mostly more a function of the testcases than the implementation).  The basic problem is that Apple designed these APIs to retain their delegates (contrary to the OpenStep convention that delegates are not retained), and also so that sessions retain tasks and tasks retain sessions.  There's a lot of potential for retain loops there.
I also need to figure out how the sharedSession method is to be used.

Headers/Foundation/NSURLSession.h
```diff
@@ -599,15 +652,43 @@ typedef NS_ENUM(NSInteger, NSURLSessionResponseDisposition) {
...
+  task: (NSURLSessionTask *)task
+  needNewBodyStream: (GSNSURLSessionBodyStreamHandler)completionHandler
+  GS_DEPRECATED;
+
+/* A new body stream is needed to continue the upload.  Reply by sending
+ * -resumeWithBodyStream: to the task.  The reply may be sent at any time and
+ * from any thread.
+ */
+- (void) URLSession: (NSURLSession *)session
+  taskNeedsNewBodyStream: (NSURLSessionTask *)task;
```
hmelder, line 691: This and -resumeWithBodyStream is a GNUstep extension and not part of Apple's NSURLSession implementation. I think we should mark them as such.

Headers/Foundation/NSURLSession.h
```diff
@@ -619,14 +700,24 @@ typedef NS_ENUM(NSInteger, NSURLSessionResponseDisposition) {
...
 
+/** Deprecated: implement URLSession:dataTask:didReceiveResponse: instead.
+ * This method requires a compiler with blocks and is not looked for by a
+ * library built without them.
+ */
+- (void) URLSession: (NSURLSession *)session
+  dataTask: (NSURLSessionDataTask *)dataTask
+  didReceiveResponse: (NSURLResponse *)response
+  completionHandler: (GSNSURLSessionResponseDispositionHandler)completionHandler
+  GS_DEPRECATED;
```
hmelder, line 711: Why do we deprecate APIs for compatibility reasons? We support an outdated compiler and runtime, but that does not mean we should actively discourage users to use newer features. For me, deprecating something means that Apple deprecated the method as well. This is not a deprecation reason IMO.

Source/NSURLSession.m
```diff
@@ -27,6 +27,99 @@
...
+   * own event-handling system integrated with the work thread run loop. \
+   * \
+   * Event creation and deletion is driven by the various callbacks \
+   * registered during initialisation of the multi handle. \
+   */ \
+  CURLM * _multiHandle; \
+  /* Drives a dedicated thread running an NSRunLoop.  All libcurl multi \
+   * handle activity (adding handles, socket events and the timer) happens on \
+   * that thread, which serialises access in place of a dispatch queue and \
+   * keeps GNUstep free of a libdispatch dependency.  The helper holds the \
```
hmelder, line 72: "which serialises access in place of a dispatch queue and  keeps GNUstep free of a libdispatch dependency"

The reader of this comment will likely be confused with this information here. Why introduce the older implementation here. You can just explain why we need a work queue, and its invariants.

Source/NSURLSessionTask.m
```diff
@@ -1733,12 +1967,29 @@ - (GSNSURLSessionDataCompletionHandler) _completionHandler
...
+  CALL_BLOCK(handler, data, response, error);
 }
 
 - (void) dealloc
 {
-  _Block_release(_completionHandler);
+  if (NULL != _completionHandler)
+    {
+      Block_release(_completionHandler);
+    }
```
hmelder, line 1992: the null check is already in `_Block_release` in libobjc2 (or libdispatch/libclosure depending on the platform).

Source/NSURLSessionTask.m
```diff
@@ -1756,7 +2007,21 @@ - (GSNSURLSessionDownloadCompletionHandler) _completionHandler
 
 - (void) _setCompletionHandler: (GSNSURLSessionDownloadCompletionHandler)handler
 {
-  _completionHandler = _Block_copy(handler);
+  if (NULL != handler)
+    {
+      _completionHandler = Block_copy(handler);
+    }
+}
```
hmelder, line 2014: Same with `Block_copy`.

Source/NSURLSessionTask.m
```diff
@@ -1771,7 +2036,10 @@ - (void) _updateCountOfBytesWritten: (int64_t)count
 
 - (void) dealloc
 {
-  _Block_release(_completionHandler);
+  if (NULL != _completionHandler)
+    {
+      Block_release(_completionHandler);
+    }
```
hmelder, line 2042: and here

Source/NSURLSessionTask.m
```diff
@@ -1189,177 +1240,336 @@ - (instancetype) initWithSession: (NSURLSession *)session
...
+  /* Continue with the transfer */
+  curl_easy_pause(internal->_easyHandle, CURLPAUSE_CONT);
+}
+
+/* The three bridges below send the deprecated form of a delegate method,
+ * the one which answers through a completion handler, and forward the answer
+ * to the reply method above.  A compiler without blocks cannot build a
+ * handler to pass, so the deprecated selectors are not looked for there and
+ * these are never reached (see +initialize).
+ */
```
hmelder, line 1500: Again, why deprecate a more streamlined API only because of a legacy runtime?

Source/NSURLSessionTask.m
```diff
@@ -1189,177 +1240,336 @@ - (instancetype) initWithSession: (NSURLSession *)session
...
+  [self _setAwaitingResponseDisposition: NO];
+  /* Unpause to flush the buffered body, then deliver any completion that was
+   * held while awaiting the delegate. */
+  curl_easy_pause(internal->_easyHandle, CURLPAUSE_CONT);
+  [internal->_session _deliverHeldCompletionForTask: self];
+}
+
+/* The delegate replies to URLSession:taskNeedsNewBodyStream: here. */
+- (void) resumeWithBodyStream: (NSInputStream *)bodyStream
+{
```
hmelder, line 1486: This is a GNUstep extension and the behaviour should be better documented, or factored out into GSURLSession.

hmelder (commented):
I'm not sure what to think of the change from a consolidated process-wide thread pool, to an implementation spawning a worker thread per `NSURLSession` instance. I interacted with a big application that uses GNUstep and Qt and has around 100 threads, because of NSOperationQueue's, manually created NSThread's, Qt Threads, the Qt thread pool, and libdispatch. So there is a lot of pressure on the scheduler.

libdispatch has an API that does not require blocks and accepts pointers to callback functions instead. So throwing out libdispatch in favour of this hacky approach was not necessary in the first place. Having a centralized thread pool, or,  in the case of libdispatch, a heavily optimized concurrency runtime, is really beneficial for Foundation.

First, libdispatch only has a handful of worker threads that service differently prioritized root dispatch queues. By default, all other dispatch queues are scheduled into one of those root queues.  We therefore reduce scheduling pressure, and effectively schedule tasks based on their priority. On big-little architectures, background tasks are scheduled on the efficiency cluster instead.

libdispatch offers platform-independent I/O, and barriers to synchronize concurrent operations. We now introduced winsock2 specific stuff here, which is not great for future OS portability.

Secondly, GNUstep is a small and informal project that cannot afford re-inventing the world. IMO, we should not do networking and concurrency and instead resort to battle-tested libraries like libcurl and libdispatch.

rfm:
I think there's no issue with having libdispatch as an option, but currently it's not in a state where it's portable/reliable enough for us to depend on it and its maintainers do not seem to have that much interest in portability.   For instance, you opened an issue (#833) in July 2024 and they haven't dealt with it yet!

Basically, that means we must implement things such that they can work properly without libdispatch, but can also have the option to use it internally in cases and on platforms where it does work.

The case of blocks is somewhat different as they are simply inherently non-portable, so we just can't use them internally even though we can largely implement the Apple blocks APIs for compatibility.  Mostly blocks jsut provide an additional callback mechanism, but where there's a real problem is if Apple design APIs which depend on blocks to function: in those cases we need to ensure we add alternative methods to make those classes fully functional.

I agree with your point that a few methods (additions for that purpose) should be marked as GNUstep additions.  Also, those methods should, if possible, have a second implementation in categories in the baseAdditions library, so that they can be used on Apple systems by people wanting to write portable code.

Also, your point about deprecation: hopefully those cases have all been update to use the new macro I added, intended to make clear the difference between warning  of deprecation because we intend to remove a method in future, and warning because a method is not portable.

Finally, you may be right that the test for a null block pointer is unnecessary (maybe all implementations already check for and ignore a null value), but it's safe/defensive coding and inexpensive, so as there's no clear spec that everyone works to, I wouldn't remove the check unless we are absolutely certain that all implementations on all platforms behave that way.

PR #767, opened 2026-08-19
Fix getwd argument mismatch and NSThread define guard

HendrikHuebner:
The spinlock define guard is required for emscripten portability, since emscripten already defines this type.

### DIFF
--- a/Source/NSFileManager.m
+++ b/Source/NSFileManager.m
@@ -1330,7 +1330,7 @@ static gs_mutex_t       classLock = GS_MUTEX_INIT_STATIC;
   if (getcwd(path, PATH_MAX-1) == 0)
     return nil;
 #else
-  if (getwd(path) == 0)
+  if (getcwd(path, PATH_MAX) == 0)
     return nil;
 #endif /* HAVE_GETCWD */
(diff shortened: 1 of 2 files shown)


rfm (changes requested):
I know very little about emscripten, but these changes look like an attempt to address a configure-time issue, rather than an issue in the code:

The change to NSFileManage.m changes getwd() to getcwd() in the code path where we know that getcwd() is *not* available.  The only way such code could work is if the package is not properly configured (so the preprocessor definititon to indicate the availablity of the function has not been set correctly).

Similarly the change to NSThread.m seems to suggest that the HAVE_PTHREAD_SPIN_LOCK preprocessor definition has been set incorrectly.

I think these changes are unnecessary/undesirable, and that what is needed is some change to ensure that configure is run and correctly sets up those values.

HendrikHuebner:
@rfm 

Yeah this fix is bogus. It was LLM generated and I misunderstood what it was doing (I thought this was a change in the number of arguments for the getwd call). 

The spinlock definition guard prevents a clash with emscripten, but I'll revise the approach here and will open a separate PR if necessary.

I apologize for wasting your time with this.

PR #768, opened 2026-08-19
Fix incompatible function pointer calls

HendrikHuebner:
1. We should be casting all IMPs to the correct function pointer signature since anything else would be UB
2. autorelease_imp can be a void function pointer

This has worked fine on most platforms, however, WebAssembly indirect calls require an exact signature match, so I'll need this fixed.

### DIFF
--- a/Source/Additions/GSObjCRuntime.m
+++ b/Source/Additions/GSObjCRuntime.m
@@ -1982,7 +1982,7 @@ GSAutoreleasedBuffer(unsigned size)
   static Class	buffer_class = 0;
   static Class	autorelease_class;
   static SEL	autorelease_sel;
-  static id	(*autorelease_imp)(Class, SEL, id);
+  static void	(*autorelease_imp)(Class, SEL, id);
   static int	instance_size;
   static int	offset;
   NSObject	*o;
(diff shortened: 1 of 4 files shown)


rfm:
Point 1, that we should transition to always using correct/full function signatures, is entirely correct.

Point 2 that autorelease can return void is  just illegal.  Anywhere that tries to have a function returning void implement the autorelease method needs to be fixed.

rfm:
Looking through the diffs, it seems like the changes are all to set the wrong function signature for the autorelease method.  That suggests that either a compiler bug is incorrectly telling you that the correct signatures are wrong or, probably more likely, there's a faulty declaration in some header file that the compiler is reading which is causing misleading error/warning messages.  Can you find where it is getting the idea that the autorelease method returns void?

HendrikHuebner:
> Looking through the diffs, it seems like the changes are all to set the wrong function signature for the autorelease method. That suggests that either a compiler bug is incorrectly telling you that the correct signatures are wrong or, probably more likely, there's a faulty declaration in some header file that the compiler is reading which is causing misleading error/warning messages. Can you find where it is getting the idea that the autorelease method returns void?

As I understand `autorelease_imp` points to `+[NSAutoreleasePool addObject:]`  (See https://github.com/gnustep/libs-base/blob/c4f97074534475f3c495509737e1b5b3d7bd9412/Source/NSObject.m#L1130)

But this is a void function: https://github.com/gnustep/libs-base/blob/c4f97074534475f3c495509737e1b5b3d7bd9412/Headers/Foundation/NSAutoreleasePool.h#L213.

Am I missing something? Maybe `autorelease_imp` should be renamed accordingly.

rfm:
Sorry, my mistake, I was thinking this was a cached imp of the -autorelease method.  I will review again when I'm home in front of a computer screen (rather than working on my phone).

rfm (approved):

PR #769, opened 2026-08-19
Avoid unsupported thread priority APIs on Emscripten

HendrikHuebner:
I'd like to just disable thread priority setting on emscripten for now, since it's currently unsupported.

https://emscripten.org/docs/porting/pthreads.html

### DIFF
--- a/Source/NSThread.m
+++ b/Source/NSThread.m
@@ -1049,6 +1049,10 @@ unregisterActiveThread(NSThread *thread)
     return NO;
   }
   return YES;
+#elif defined(__EMSCRIPTEN__)
+  /* Emscripten advertises POSIX priority scheduling but does not implement
+   * the scheduler entry points. */
+  return NO;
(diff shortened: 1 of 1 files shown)


rfm (commented):
This looks like another configure time issue, since the default/fallback behavior is already to disable setting thread priority (the method does nothing and returns NO).
So, if some other codepath is being taken when you compile, it must be because configure was somehow broken to make it look like the target platform supports thread priorities.

This seems likely to be an issue with emscripten failure to run autoconf/configure correctly. so that gnustep-base is being configured to run on your build machine rather than webassembly.

HendrikHuebner:
Hmm the reasoning behind this change is that emscripten defines `_POSIX_THREAD_PRIORITY_SCHEDULING` and shims the priority scheduling API. Yet, we should not claim that a thread priority was configured when nothing actually happened. Maybe for the WASM build we can just override `_POSIX_THREAD_PRIORITY_SCHEDULING` instead? 

Either way, its not really a blocker for my WebAssembly bringup.

rfm (approved):
Ah  in that case your fix makes perfect sense.  I agree that it's better to return NO than YES in that situation.

PR #774, opened 2026-08-27
Various emscripten compatibility fixes

HendrikHuebner:
This PR contains a few compatibility fixes:

- Fix the Emscripten signal() handler type in NSObject.m. Emscripten declares handlers using int, while NSInteger causes an incompatible function-pointer type.

- Avoid redefining pthread_spinlock_t in NSThread.m when Emscripten’s pthread headers already provide it.

- Emscripten defines POSIX scheduling macros, but leaves the implementation stubbed. We should just return zero instead of following the POSIX branch

- Emscripten declares `inet_ntop` with `socklen_t` instead of `size_t`. Given this code is 30 years old and the manpages say it should be `socklen_t`, the current declaration might be wrong I think? For now this change only affects emscripten.

### DIFF
--- a/Source/NSObject.m
+++ b/Source/NSObject.m
@@ -1063,7 +1063,11 @@ static id gs_weak_load(id obj)
       }
 #else /* HAVE_SIGACTION */
       {
+#if defined(__EMSCRIPTEN__)
+	void	(*handler)(int);
+#else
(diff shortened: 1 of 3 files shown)


rfm (approved):
That all looks fine.  Thanks.

PR #775, opened 2026-08-27
NSURLSessionWebSocket implementation

HendrikHuebner:
This is the full implementation for NSURLSessionWebSocket based on libcurl's callback based websocket API.

It does not handle all possible scenarios (For example HTTP redirects, cookies), but the basic functionality and Websocket connection protocol is implemented.

I ran a concurrent stress test with this setup, which spawns a dozen thread that hammer multiple WebSocket tasks with requests concurrently and verify some basic invariants (Such as the received / send count, pingpongs and the message sizes / content). The results matched the Apple implementation and the throughput is comparable.

I don't think it makes sense to upstream this setup as a unit test however. Unit testing this implementation is a bit difficult, since you need a client/server connection to do it (Even if just localhost) and testing nontrivial cases is hard.


hmelder:
@HendrikHuebner can you send me the stress test you used, and other unit tests?

Source/NSURLSessionTaskPrivate.h
```diff
@@ -36,10 +36,32 @@
...
+                                                                 URL:
+  (NSURL *)url;
+- (void)_installRequestHeaders:
+  (NSDictionary *)requestHeaders;
 
 -(CURL *)_easyHandle;
+-(void)_setEasyHandle: (CURL *)handle;
+-(char *)_errorBuffer;
+-(struct curl_slist *)_headerList;
+-(void)_setHeaderList: (struct curl_slist *)headerList;
```
hmelder, line 64: Where is all this implemented?

Source/NSURLSessionTaskPrivate.h
```diff
@@ -36,10 +36,32 @@
...
+- (void)_configureTransferCallbacks;
+- (void)_configureProtocolOptionsForRequest: (NSURLRequest *)request
+                              configuration:
+  (NSURLSessionConfiguration *)configuration;
+- (NSMutableDictionary *)_mergedRequestHeadersForRequest:
+  (NSURLRequest *)request
+                                                        configuration:
+  (NSURLSessionConfiguration *)configuration
+                                                                 URL:
+  (NSURL *)url;
```
hmelder, line 56: Half of this does not exist

Source/NSURLSessionWebSocketTask.m
```diff
@@ -0,0 +1,1785 @@
...
+                  taskIdentifier: identifier];
+  if (self != nil)
+    {
+      curl_easy_cleanup([self _easyHandle]);
+      [self _setEasyHandle: NULL];
+      [self _initializeEasyhandleForRequest: request];
+      [self _configureTransferCallbacks];
+      [self _configureProtocolOptionsForRequest: request
+                                    configuration: [session configuration]];
+      GS_CREATE_INTERNAL(NSURLSessionWebSocketTask);
```
hmelder, line 1359: This is extremely hacky. We cleanup the easy handle we allocated in the initializer of the superclass, only to create a new easy handle. The initialization in the superclass was based on the assumption that a NSURLSessionTask does HTTP transfers. We should move all this into the respective subclasses instead.

Source/NSURLSessionWebSocketTask.m
```diff
@@ -0,0 +1,1785 @@
...
+      internal->receive.controlBuffer = [[NSMutableData alloc] init];
+      internal->receive.maximumMessageSize = 1024 * 1024;
+      GSURLSessionWebSocketClearActiveSendEntryLocked(self);
+      internal->lifecycle.phase = GSURLSessionWebSocketLifecycleStateOpen;
+      internal->receive.phase = GSURLSessionWebSocketReceiveStateIdle;
+      internal->send.nextPingIdentifier = 1;
+      internal->receive.frameOffset = 0;
+      internal->send.frameStartRetryPending = NO;
+      internal->lifecycle.closeFrameSent = NO;
+      internal->lifecycle.closeFrameReceived = NO;
```
hmelder, line 1374: Perhaps we should create constructor functions for this.

Source/NSURLSessionWebSocketTask.m
```diff
@@ -0,0 +1,1785 @@
...
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 31 Milk Street #960789 Boston, MA 02196 USA.
+*/
+
```
hmelder, line 23: Would be nice to have a design overview here for future readers.

hmelder:
The NSURLSessionTask integration is also hacked. All the progress metering is not implemented for instance.

hmelder:
I've tested Apple's implementation to understand the behaviour and side-effects: https://gist.github.com/hmelder/855caa29d117bbaa057de7f15f6cb911

Source/NSURLSessionWebSocketTask.m
HendrikHuebner: yeah this got vibe-refactored. I was running out of time to work on this in my last week at algo. There is a lot to clean up.

Source/NSURLSessionTaskPrivate.h
HendrikHuebner: Yes some of these got refactored away. They shouldn't be called anywhere so just remove the declarations

Source/NSURLSessionWebSocketTask.m
HendrikHuebner: I didn't create a design doc but I can do a write up tomorrow maybe

PR #778, opened 2026-09-07
KVO: skip observers removed during change delivery

rupertdaniel:
## Problem

Since #743, `_dispatchDidChange` delivers notifications in two phases: pending notifications are collected under the change lock, then `-observeValueForKeyPath:ofObject:change:context:` is called after the locks are dropped. `keyObserver.isRemoved` is only checked during the collection phase, and `_NSKVOKeypathObserver` holds its `observer` unretained (`assign`, matching the reference platform).

If an observer's `-observeValueForKeyPath:...` removes **another** observer of the same change and that observer is then deallocated — a common pattern where the first notification tears down UI whose components each observe the same object and remove themselves in `dealloc` — the delivery loop still messages it. Depending on whether the freed memory has been reused, this is either a crash in `objc_msgSend` (with libobjc2, a null-pointer dereference at a small offset, since `object_dispose` clears the isa) or a notification delivered after `removeObserver:` returned. We hit the crash variant in a production Android application, 100% reproducibly, whenever a KVO change caused the observing UI to be torn down synchronously.

Before #743 this case was handled: the single-loop dispatch re-checked `isRemoved` immediately before each delivery.

## Reproduction

Two observers on the same key; the one notified first removes and releases the other mid-delivery:

```objc
@implementation Remover
- (void) observeValueForKeyPath: (NSString*)k ofObject: (id)o
                         change: (NSDictionary*)c context: (void*)x
{
  [subject removeObserver: victim forKeyPath: @"value"];
  DESTROY(victim);   // victim deallocated while the change is still being delivered
}
@end

[subject addObserver: victim  forKeyPath: @"value" options: ... context: NULL];
[subject addObserver: remover forKeyPath: @"value" options: ... context: NULL];
[subject setValue: 1];  // remover is notified first (reverse registration order),
                        // then the deallocated victim is messaged
```

## Fix

Hold the retained `_NSKVOKeyObserver` alongside each pending notification and re-check `isRemoved` at delivery time, skipping observers removed by an earlier delivery of the same change. This restores the guarantee that no notifications arrive after `removeObserver:` returns (which observer `dealloc` code relies on), while keeping delivery synchronous on the same stack as the change — the locking and deadlock-avoidance structure from #743 is unchanged.

`_dispatchWillChange` was audited for the same hazard: its `NSKeyValueObservingOptionPrior` delivery already sits behind a per-iteration `isRemoved` check and needs no change.

## Tests

Adds `Tests/base/NSKVOSupport/removalDuringDelivery.m`. Against current master (Android arm64, libobjc2):

```
Failed test: ... an observer removed and deallocated during delivery is not messaged
Passed test: ... the observer which removed it is notified once
Passed test: ... and keeps receiving changes
Failed test: ... while the deallocated observer stays unnotified
```

With this change all four pass, and the original application-level crash is confirmed fixed on device.

### DIFF
--- a/Source/NSKVOSupport.m
+++ b/Source/NSKVOSupport.m
@@ -1271,6 +1271,7 @@ _dispatchWillChange(_NSKVOObservationInfo *observationInfo,
 
 /* One observer call, held until every lock has been dropped. */
 typedef struct {
+  _NSKVOKeyObserver     *keyObserver;
   _NSKVOKeypathObserver *keypathObserver;
   NSMutableDictionary   *change;
 } GSKVOPendingNotification;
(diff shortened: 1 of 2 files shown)


rfm (approved):
Thanks, a nice clean fix with testcases :-)

PR #779, opened 2026-09-07
Fix strong-to-weak NSMapTable lookups after resizing

rupertdaniel:
## Summary

`NSMapTable` instances with strong keys and weak values (for example `+strongToWeakObjectsMapTable`) lose the ability to find their entries once the table resizes. The entries are still counted and enumerated, but `-objectForKey:` returns `nil` for them, `-setObject:forKey:` adds a duplicate node instead of replacing the existing one, and `-removeObjectForKey:` leaves the entry in place.

## Cause

`GSIMapRemangleBuckets()` in `Headers/GNUstepBase/GSIMap.h` relies on `GSIMapWeakIsEmpty()` to return the retained key of each surviving node so it can be rehashed into the new bucket array. That helper only loads the key when the keys are weak (`GSI_MAP_ZEROED(map) & 1`). When only the values are weak it leaves `k` as `nil`, so every node is rehashed with the hash of `nil` and lands in the wrong bucket.

## Fix

When keys are not weak, read the stored key with `GSI_MAP_READ_KEY()` instead of using the `nil` returned by the helper. The weak-key path is unchanged and still uses the retained key, and the existing `[k release]` stays balanced because `k` is `nil` in the strong-key case. Other callers of `GSIMapWeakIsEmpty()` only retain and release the key, so this is the only affected site.

## Test

`Tests/base/NSMapTable/weakGrowth.m` builds a table with capacity 0 for each combination of strong and weak keys and values, inserts 256 externally retained entries, and checks:

- `-count` is unchanged and every entry is found through a separately allocated equal key.
- Replacing every value keeps the count at 256, so no duplicate nodes were added.
- A released weak value, or a released weak key, is no longer found.
- After growing to 1024 entries, every surviving entry is still found.
- Removing every entry through equal keys empties the table.

Before the fix, the strong-key/weak-value case fails the four lookup, replacement and removal checks; the other three combinations pass. After the fix all 24 checks pass, and the existing `NSMapTable` tests (`basic`, `create`, `general`, `strongObjects`, `weakObjects`) are unaffected.

## Reproduction

A minimal standalone reproduction with the unpatched library:

```objc
NSMapTable *map = [NSMapTable strongToWeakObjectsMapTable];
NSMutableArray *owners = [NSMutableArray array];
for (int i = 0; i < 4096; i++) {
    NSObject *value = [NSObject new];
    [owners addObject: value];
    [map setObject: value forKey: [NSString stringWithFormat: @"key-%d", i]];
}
int missing = 0;
for (int i = 0; i < 4096; i++) {
    if ([map objectForKey: [NSString stringWithFormat: @"key-%d", i]] != owners[i]) missing++;
}
NSLog(@"count=%lu missing=%d", (unsigned long)map.count, missing);
```

Unpatched this reports `count=4096 missing=3446`, with the first entry becoming unreachable after the third insertion. A `strongToStrongObjectsMapTable`, or a strong-to-weak table preallocated with capacity 4096 so it never resizes, reports `missing=0`.

Verified on Android (aarch64, libobjc2 with the gnustep-2.2 ABI, NDK r27) against current `master`, running the six `Tests/base/NSMapTable` programs before and after the change: 87 passed / 4 failed before, 91 passed / 0 failed after.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2026-09-07 Rupert Daniel <rupert@cancelmonday.com>
+
+        * Headers/GNUstepBase/GSIMap.h: When rehashing a table with strong
+        keys and weak values, hash the stored key rather than the nil key
+        returned by GSIMapWeakIsEmpty(), so entries stay reachable after
+        the table resizes.
+        * Tests/base/NSMapTable/weakGrowth.m: New regression test for
+        lookup, replacement and removal after growth with each combination
+        of strong and weak keys and values.
+
 2026-09-02 Richard Frith-Macdonald <rfm@gnu.org>
 
         * configure.ac: Add test for eventfd presence on linux and use it
--- a/Headers/GNUstepBase/GSIMap.h
+++ b/Headers/GNUstepBase/GSIMap.h
@@ -657,7 +657,14 @@ GSIMapRemangleBuckets(GSIMapTable map,
 		  GSIMapKey	key;
 
 		  GSIMapRemoveNodeFromBucket(old_buckets, node);
-		  key.addr = (uintptr_t)k;
+		  if (GSI_MAP_ZEROED(map) & 1)
+		    {
+		      key.addr = (uintptr_t)k;
+		    }
+		  else
+		    {
+		      key = GSI_MAP_READ_KEY(map, &node->key);
+		    }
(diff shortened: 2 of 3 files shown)


rfm (approved):
Thanks, good spot.

PR #786, opened 2026-09-15
NSString: handle empty path extensions

hmelder:
GNUstep's Foundation appends a trailing dot, when NSString:stringByAppendingPathExtension: is called with an empty string as argument. This is not the case on Apple platforms.


rfm (approved):
Looks good.  I confirmed the behavior on my system.

PR #788, opened 2026-09-17
NSURLSessionTask: Defer Curl handle initialization

hmelder:
A new (private) designated initializer for NSURLSessionTask now accepts a Curl easy handle as an argument. Previously there was only one initializer, `NSURLSessionTask:initWithSession:request:taskIdentifier:`, that configures the task for HTTP transfers.

With `NSURLSessionWebSocketTask` and `NSURLSessionStreamTask`, different Curl configurations need to be supported as well. The initializer of such a subclass can now create a Curl easy handle and add it to the task object, transfering ownership. The semantics of the existing initializer is unchanged.


rfm (approved):
Looks good.
I see one workflow is failing, on NSURLSession tests, but I don't think that's caused by this change.

PR #790, opened 2026-09-19
NSMessagePort: Never block in accept() on the listening socket

probonopd:
## Problem

A process that serves an `NSMessagePort` (any vended `NSConnection` using the default message ports) can hang forever with its main thread blocked in `accept()`:

```
#3  __libc_accept (fd=6, ...)
#4  -[NSMessagePort receivedEvent:type:extra:forMode:]
#5  -[GSRunLoopCtxt pollUntil:within:]
#6  -[NSRunLoop acceptInputForMode:beforeDate:]
...
```

`ss -xlp` shows the port's listening socket with an empty backlog (Recv-Q 0), so there is nothing to accept. Every client that then talks to the process over DO also hangs waiting for its reply. We hit this with a desktop shell that vends a service: after it hung, every application that called the service (e.g. while quitting) hung too.

## Cause

The listening socket is created in blocking mode (unlike `NSSocketPort`, which sets `O_NONBLOCK` on its listener). The run loop can report the listener readable when the pending connection has already been accepted: an outer `-pollUntil:within:` collects the readable descriptors, one of the handlers runs a nested run loop (a modal loop, a synchronous DO call, an event tracking loop), the nested loop sees the same listener readable and accepts the connection, and when control returns the outer loop dispatches its stale result. `accept()` on the empty backlog then blocks. The comment in the `accept()` error path ("accept failed - handled in other thread?") shows the code already expects accept to fail rather than block in that case.

## Fix

Put the listener into non-blocking mode right after `listen()`/`getsockname()`, and fail to create the port if that is not possible (like the other setup errors there). `accept()` on an empty backlog then returns `EAGAIN` and the existing error path returns. Accepted sockets are unaffected: `GSMessageHandle` already makes them non-blocking.

## How to reproduce

The patch adds `Tests/base/NSMessagePort/listener.m`. It creates an `NSMessagePort`, checks that its listening descriptor is non-blocking, and hands the port a readable event for the listener when no client is connecting, exactly what the stale poll result does. An `alarm(5)` turns the hang into a failure.

```sh
cd Tests
gnustep-tests base/NSMessagePort
```

Without the fix in `Source/NSMessagePort.m` (only the test applied):

```
Passed test:  listener.m:44 ... a message port has a listening descriptor
Failed test:  listener.m:48 ... the listening descriptor does not block
Failed test: accept() on the listener blocked
Failed file:  listener.m aborted without running all tests!
```

With the fix:

```
Passed test:  listener.m:44 ... a message port has a listening descriptor
Passed test:  listener.m:48 ... the listening descriptor does not block
Passed test:  listener.m:60 ... a readable event without a pending connection returns
```

The `base/NSConnection`, `base/NSDistributedLock` and `base/NSURLConnection` tests give the same results with and without the fix (Linux).

cc @pkgdemon


rfm (approved):
This is excellent.  The failure to set non-blocking was clearly an oversight that needs fixing.

PR #791, opened 2026-09-20
Round the run loop wait up instead of truncating it

probonopd:
A wait of less than a millisecond truncates to a poll timeout of zero, so the run loop polls without waiting and runs again at once, spinning at full speed until the timer date is reached. Any program with a repeating timer pays a burst of busy looping before each fire; in a desktop session this showed up as a few percent of a processor core used while nothing happened.

```objc

@interface Ticker : NSObject
@end
@implementation Ticker
- (void) tick: (NSTimer *)t { (void)t; }
@end

int main(int argc, char **argv)
{
  @autoreleasepool {
    double hz = argc > 1 ? atof(argv[1]) : 30.0;
    Ticker *ticker = [[Ticker alloc] init];

    [NSTimer scheduledTimerWithTimeInterval: 1.0 / hz
                                     target: ticker
                                   selector: @selector(tick:)
                                   userInfo: nil
                                    repeats: YES];
    [[NSRunLoop currentRunLoop] addPort: [NSPort port]
                                forMode: NSDefaultRunLoopMode];
    [[NSRunLoop currentRunLoop] run];
  }
  return 0;
}
```

```
$ clang $(gnustep-config --objc-flags) spin.m -o spin $(gnustep-config --base-libs)
$ ./spin 30 & perf stat -e syscalls:sys_enter_poll -p $! -- sleep 3

before:  19322  syscalls:sys_enter_poll        0.8 % of a core
after:      90  syscalls:sys_enter_poll        0.0 % of a core
```

90 poll calls in 3 seconds is the 30 Hz of the timer itself. base/NSRunLoop passes with the change (65 tests), as do base/NSConnection and base/NSMessagePort.

The change rounds the wait up, so a wait shorter than a millisecond asks poll for one millisecond instead of none.

cc @pkgdemon


rfm (approved):
I think this is a good enough fix for now given the current code.  However, the issue also highlights the fact that this code is old and in need of an update: using millisecond granularity is probably wrong on modern hardware and we should be using APIs with microsecond granularity where possible.

